Issue #2, opened 2021-07-06
[EPIC] nucypher-ts MVP


piotr-roslaniec:
This is a rough draft of several `nucypher/nucypher-ts` modules in an attempt to tie together a boilerplate for the rest of the library.

The idea is to reimplement existing `nucypher/nucypher` flows using Porter as a replacement for all network interactions. I will be using acceptance and integration tests from `nucypher` as a reference point. The code borrows heavily from `nucypher/nucypher` codebase, especially abstractions such as policies, crypto powers, keyrings, etc.

The goal is to have at least one "user story" working. A lot of implementations will be stubbed-out initially, especially ones related to Porter or blockchain transactions. 

Further down the road, `nucypher/NuSnap` will be updated to implement  `nucypher/nucypher-ts`.

The design of the user-facing library is not final and will be rewritten heavily. As always, all comments are welcome!

### To Do
- [x] Implement a use story - "Alice grants access to Bob through a new blockchain policy"
   - [x] Alice / Enrico encrypts a message
   - [x] Alice creates a blockchain policy
   - [x] Alice publishes blockchain policy to blockchain
   - [x] Bob retrieves treasure map from ~~blockchain policy~~ sidechannel
   - [x] Bob retrieves capsule fragments 
   - [x] Bob decrypts the message
- [x] Characters
  - [x] Alice
  - [x] Bob
  - [x] Porter
  - [x] ~~Ursula~~ Currently replaced with Porter proxy
  - [x] Enrico
- [x] Ethereum
  - [x] Web3 client
  - [x] Contracts
- [x] Crypto - Review selection of js libs
  - [x] ~~HKDF with Blake2 (64)~~ Using keccak256 for now
  - [x] Keccak256
  - [x] ECDSA
  - [x] SHA256
- [x] ~~Should `nucypher-ts` be browser and Node.js compatible?~~ Focus on browser compatibility 
- [x] Utilities
  - [x] Compatibility with Serialization (Bytesplitter)
  - [x] Compatibility with NuCypher constants (`constant-sorrow`)
- [x] Design decisions (user API)
    - [x] Handling transactions, web3 provider
    - [x]  Remove `Buffer` usage
 - [x] Cryptopowers: 
     - [x] ~~Interface with NuSnap~~  Postponed
     - [x] Instantiate from seed (secret ket), `fromSecretKey` 
     - [x] ~~Instantiate from web3 provider, `fromWeb3`~~ Connect a web3 provider instead
 
[...]

PR #3, opened 2021-07-06
Alice grants

piotr-roslaniec:
Implements user story: "Alice grants access to Bob through blockchain policy"

Refers to: #2 

Summary of changes:
* Added 5 new `characters`:
  * `Alice` -  encrypts messages for the recipient, grants access to said messages using `BlockchainPolicy`
  * `Bob` - joins the policy, retrieves cFrags from `Ursula` using `TreasureMap` and `Porter`
  * `Enrico`  - a dashing rogue, encrypts messages for Alice on occasion
  * `Porter` - assists `Bob` and `Alice` in communicating with the NuCypher network; is a REST API client
  * `Ursula` - enables communication with specific nodes in the network when proposing arrangements or enacting policies
* Added wrappers for cryptographic material in a form of `keyring.ts`, `keys.ts` and `powers.ts`
* Implemented reencryption of the capsule, from generating kFrags, through `BlockchainPolicy` creation, to collecting cFrags from Ursulas using `TreasureMap` 
* Added initial selection of JS cryptographic libraries
* Communication with Porter, web3 and some other non-critical paths such as serialization have been stubbed.


src/characters/porter.ts
```diff
@@ -0,0 +1,153 @@
...
+    const asBytes = Buffer.from(resp.data.result.treasureMap, 'base64');
+    return PublishedTreasureMap.fromBytes(asBytes);
+  }
+
+  public static async executeWorkOrder(
+    workOrder: WorkOrder
+  ): Promise<WorkOrderResult> {
+    const data: PostExecuteWorkOrderRequest = {
+      ursula: workOrder.ursula.checksumAddress,
+      work_order: workOrder.toBytes().toString('base64'),
```
derekpierre, line 144: Is it toBytes() or payload()...? On the python side it has been payload() unless something changed...

piotr-roslaniec: It should be `payload`, I will correct that.

PR #6, opened 2021-08-05
[EPIC] Alice grants MVP

### LINKED ISSUE #2: [EPIC] nucypher-ts MVP

piotr-roslaniec:
Closes #2, #8, #9 

Notes for reviewers:
* This PR is accompanied by the release of [an NPM package](https://www.npmjs.com/package/nucypher-ts) and [documentation](https://nucypher.github.io/nucypher-ts/). 
* This early release is showcased in the [`nucypher-ts-demo`](https://github.com/nucypher/nucypher-ts-demo).
* It's compatible with the latest changes in treasure maps and key frags but doesn't take into account changes from https://github.com/nucypher/nucypher/pull/2802


src/characters/porter.ts
```diff
@@ -0,0 +1,148 @@
...
+          'Access-Control-Allow-Origin': this.porterUri,
+        },
+      }
+    );
+    return resp.data.result.retrieval_results
+      .map((result) => result.cfrags)
+      .map((cFrags) => {
+        const parsed = Object.keys(cFrags).map((address) => [
+          address,
+          VerifiedCapsuleFrag.fromVerifiedBytes(fromBase64(cFrags[address])),
```
derekpierre, line 143: This skips the verification process, correct? Does this mean that the cfrag bytes being returned by Porter are implicitly trusted as being correct? Should we be verifying the cfrag returned and not accepting it as correct without actual verification?

src/policies/collections.ts
```diff
@@ -0,0 +1,313 @@
...
+            ...toCanonicalAddress(ursulaAddress),
+            ...encodeVariableLengthMessage(encryptedKFrag.toBytes()),
+          ])
+      )
+      .reduce((previous, next) => new Uint8Array([...previous, ...next]));
+  }
+}
+
+export class AuthorizedKeyFrag {
+  public static readonly ENCRYPTED_SIZE = 477; // Hardcoded - depends on Umbral implementation
```
derekpierre, line 162: Perhaps we need to think about how to keep constants like these up-to-date with what is in `nucypher`?  Even if it is as simple as tests. Maybe the `nucypher-core` work could help with that... @fjarri ?

src/policies/policy.ts
```diff
@@ -0,0 +1,213 @@
...
+          `(${paymentPeriods} periods) without a remainder.`
+      );
+    }
+
+    // TODO: This check feels redundant
+    const ratePerPeriod = Math.floor(value / shares / paymentPeriods);
+    const recalculatedValue = paymentPeriods * ratePerPeriod * shares;
+    if (recalculatedValue != value) {
+      throw new Error(
+        `Invalid policy value calculation - ${value} cant be divided into ${shares} ` +
```
derekpierre, line 113:
```suggestion
        `Invalid policy value calculation - ${value} cannot be divided into ${shares} ` +
```

test/utils.ts
```diff
@@ -0,0 +1,219 @@
...
+    getNetwork: () => Promise.resolve({ name: 'mock', chainId: -1 }),
+  };
+};
+
+export const mockUrsulas = (): Ursula[] => {
+  return [
+    {
+      encryptingKey:
+        '025a335eca37edce8191d43c156e7bc6b451b21e5258759966bbfe0e6ce44543cb',
+      checksumAddress: '0x5cF1703A1c99A4b42Eb056535840e93118177232',
```
derekpierre, line 59: If these are supposed to match what Porter returns exactly then `encryptingKey` -> `encrypting_key`, and `checksumAddress` -> `checksum_address`.

.prettierignore
```diff
@@ -0,0 +1,2 @@
+# package.json is formatted by package managers, so we ignore it here
+package.json
```
derekpierre, line 2: missing newline.

src/characters/porter.ts
piotr-roslaniec: That's a good point, these should definitely be verified.

src/policies/collections.ts
piotr-roslaniec: I'm also interested in values defined in `nucypher` using `constant_sorrow`. I would prefer having a single source of truth for parameters shared across different projects (could be `nucypher-core`). One that defines these values "up-front", instead of (for example) calculating them at runtime.

CHANGELOG.md
```diff
@@ -0,0 +1,19 @@
...
+### [0.1.1-alpha](https://github.com/nucypher/nucypher-ts/compare/v0.1.0-alpha...v0.1.1-alpha) (2021-09-30)
+
+## 0.1.0-alpha (2021-09-30)
+
+
+### Features
+
+* Alice grants workflow ([7a7c3cd](https://github.com/nucypher/nucypher-ts/commit/7a7c3cd66909841eb2b10a3001f4df75ca6c006a))
+
+## 0.1.0-alpha (2021-09-30)
```
fjarri, line 14: Repeating lines?

src/agents/constants.ts
```diff
@@ -0,0 +1,35 @@
+interface Contracts {
+  NU: string;
```
fjarri, line 2: Is it worth having a `ChecksumAddress` type? Even if it's just an alias for a `string`?

fjarri: Ah, looking in the next file, and you already have a `ChecksumAddress` defined. Could it be used here?

src/characters/alice.ts
```diff
@@ -0,0 +1,196 @@
...
+    return this.signingPower.publicKey;
+  }
+
+  public get signer(): Signer {
+    return this.signingPower.signer;
+  }
+
+  public static fromSecretKey(
+    config: Configuration,
+    secretKey: Uint8Array
```
fjarri, line 55: Judging by how it is used in `NucypherKeyring`, it is rather secret material than a secret key (as in `umbral.SecretKey`), right?

src/characters/bob.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  public get verifyingKey(): PublicKey {
+    return this.signingPower.publicKey;
+  }
+
+  public get signer(): Signer {
+    return this.signingPower.signer;
+  }
+
+  public static fromPublicKeys(
```
fjarri, line 52: Perhaps we do not have to repeat the current design here and can separate local and remote characters instead? Now, for example, a remote Bob still has a Porter instantiated inside it.

src/crypto/api.ts
```diff
@@ -0,0 +1,20 @@
+import sha3 from 'js-sha3';
+import { PublicKey, Signature } from 'umbral-pre';
+
+import { fromHexString, toBytes } from '../utils';
+
+export const verifySignature = (
```
fjarri, line 6: Is it really worth it allowing union types here? I don't think `message` is ever a `string`, and the caller might as well handle the conversion (then probably this function is not needed at all).

src/crypto/constants.ts
```diff
@@ -0,0 +1,24 @@
+import { fromHexString } from '../utils';
+
+// Umbral constants
+export const SIGNATURE_LENGTH = 64;
```
fjarri, line 4: There are `serialized_size()` methods in Umbral objects, but they are currently only exposed in Python - we can do the same for WASM. Although it won't be necessary with the advent of `nucypher-core`

src/crypto/keys.ts
```diff
@@ -0,0 +1,30 @@
...
+  constructor(keyingMaterial: Uint8Array) {
+    if (keyingMaterial.length !== UMBRAL_KEYING_MATERIAL_BYTES_LENGTH) {
+      throw Error(
+        `Expected keyingMaterial to be ${UMBRAL_KEYING_MATERIAL_BYTES_LENGTH} bytes long, received ${keyingMaterial.length} bytes instead`
+      );
+    }
+    this.keyingMaterial = keyingMaterial;
+  }
+
+  public deriveSecretKeyFromLabel(label: string): SecretKey {
```
fjarri, line 20: Hm, why can't you use `SecretKeyFactory` here?

src/crypto/powers.ts
```diff
@@ -0,0 +1,189 @@
...
+      delegatingSecretKey,
+      receivingKey,
+      signer,
+      threshold,
+      shares,
+      false,
+      false
+    );
+    // Turn KFrags into VerifiedKFrags - we know they're good
+    const verifiedKFrags = kFrags.map((kFrag) =>
```
fjarri, line 87: Doesn't `generateKFrags()` already return `VerifiedKeyFrag`s?

src/policies/collections.ts
fjarri: In https://github.com/nucypher/nucypher/pull/2802 I'm just wrapping encrypted AKFrag into a `VariableByteString` to get rid of that constant - I'm tired of changing it all the time too.

As for CS values, they're also being removed in 2802.

fjarri: But in any case, if `nucypher-core` takes off, all this logic will be hidden in it.

fjarri:
As a general note, you'll need to add versioning support to all protocol objects, following https://github.com/nucypher/nucypher/pull/2767 and (when it's merged) https://github.com/nucypher/nucypher/pull/2805... or wait for `nucypher-core`

src/policies/collections.ts
vepkenez: @piotr-roslaniec @fjarri I had a go at defining those constants here… I got distracted by other things, but maybe it would be safe to assume they’d end up something like this https://github.com/nucypher/nucypher/pull/2556

fjarri: Yep, I saw that PR, and that's essentially what I did in 2802. Of course this is all subject to change depending on https://github.com/nucypher/nucypher/issues/2743

src/crypto/keys.ts
piotr-roslaniec: No reason, I just forgot about `SecretKeyFactory::makeKey`

piotr-roslaniec:
Has been addressed in addressed in #12

PR #13, opened 2021-10-13
Alice revokes

### LINKED ISSUE #10: Revoke blockchain policy
* Currently there is no way for Alice to revoke a blockchain policy once it was granted to Bob
* Implement revocation mechanism

piotr-roslaniec:
- Implements only on-chain part of the revocation flow
- Contains minor changes to how blockchain policy parameters are calculated and validated
- Fixes gas limit for policy creation being too low for multiple nodes
- Removes some unused code (arrangements, off-chain revocation)
- Closes #10


src/agents/policy-manager.ts
```diff
@@ -44,6 +45,33 @@ export class PolicyManagerAgent {
     return tx;
   }
 
+  public static async policyDisabled(
+    transactingPower: TransactingPower,
```
KPrasch, line 49: Considering that this is a read-only function, why is `transactingPower` needed here?

src/characters/alice.ts
```diff
@@ -185,4 +187,51 @@ export class Alice {
       policyEndTime
     ) as BlockchainPolicyParameters;
   }
+
+  public async revoke(
+    policy: EnactedPolicy
```
KPrasch, line 192: Considering that only `policyID` is used from `EnactedPolicy`, and alice may be revoking this policy without having any good way to instantiate an EnactedPolicy in a separate runtime form the granting operation, perhaps we just accept the policy ID and not an entire Policy here.

piotr-roslaniec: But don't we need other things such as revocation kits for off-chain revocation? This is not exactly a priority in this issue

piotr-roslaniec: How about:
```
 public async revoke(
    // provide atleast one:
    policyId?: Uint8Array, // to revoke on-chain 
    revocationKit?: RevocationKit // to revoke off-chain
  ) 
```

piotr-roslaniec:
@derekpierre That's true, this is a left-over code. Since we are only working with on-chain revocation we can also remove other related lines.

PR #20, opened 2021-12-07
Use proper string encoding when deriving policyId

ghardin1314:
When `Alice` grants a policy, she takes in a `label` directly from the user which is in utf-8 encoding. This `label` attribute undergoes no other transformations as it is passed through `Alice.createPolicy` or the constructor of `BlockchainPolicy` where it is passed to `HRAC.derive`. 

`HRAC.derive` assumes that the label is hex encoded which it is not. My proposed solution is for `HRAC` to assume utf-8 encoding but alternate solutions would be for `Alice` or `BlockchainPolicy` to change the encoding of `label` to hex.

### DIFF
--- a/src/policies/hrac.ts
+++ b/src/policies/hrac.ts
@@ -21,7 +21,7 @@ export class HRAC {
       new Uint8Array([
         ...publisherVerifyingKey,
         ...bobVerifyingKey,
-        ...fromUtf8(label),
+        ...toBytes(label),
       ])
     ).slice(0, HRAC.BYTE_LENGTH);
     return new HRAC(hrac);
--- a/src/policies/hrac.ts
+++ b/src/policies/hrac.ts
@@ -1,5 +1,5 @@
 import { keccakDigest } from '../crypto/api';
-import { fromHexString } from '../utils';
+import { fromUtf8 } from '../utils';
 
 export class HRAC {
   public static readonly BYTE_LENGTH: number = 16;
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -18,6 +18,11 @@ export const toBase64 = (bytes: Uint8Array): string =>
 export const fromBase64 = (str: string): Uint8Array =>
   Buffer.from(str, 'base64');
 
+export const toUtf8 = (bytes: Uint8Array): string =>
+  Buffer.from(bytes).toString('utf-8');
+
+export const fromUtf8 = (str: string): Uint8Array => Buffer.from(str, 'utf-8');
```
piotr-roslaniec, line 24: At lines 1 & 4 you can see similar logic implemented using [Web API's `TextEncoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder). We're primarily targeting browser clients so we try to use it when possible, instead of relying on node.js's `Buffer`.

src/policies/hrac.ts
```diff
@@ -1,5 +1,5 @@
 import { keccakDigest } from '../crypto/api';
-import { fromHexString } from '../utils';
+import { fromUtf8 } from '../utils';
```
piotr-roslaniec, line 2:
```suggestion
import { toBytes } from '../utils';
```

src/policies/hrac.ts
```diff
@@ -21,7 +21,7 @@ export class HRAC {
       new Uint8Array([
         ...publisherVerifyingKey,
         ...bobVerifyingKey,
-        ...fromHexString(label),
+        ...fromUtf8(label),
```
piotr-roslaniec, line 24:
```suggestion
        ...toBytes(label),
```

piotr-roslaniec (changes requested):
Hello @ghardin1314, thanks for catching this!

src/utils.ts
ghardin1314: Ah yeah totally skimmed past those first two functions. Implemented suggestions and removed my added functions.

PR #21, opened 2021-12-17
[WIP] Decoupling policy granting and generation

ghardin1314:
The rationale behind this is to decouple the generation of a `Policy` from the actual onchain granting of said `Policy`. This is to allow our smart contracts to be the owner and sponsor of our policies, not `Alice`. 

There may be a more practical way to implement this, but this seems to be the least invasive on the existing API. I see there are a lot of changes for the `0.5.0` candidate so let me know where you think this might best slot into.

### DIFF
--- a/src/characters/alice.ts
+++ b/src/characters/alice.ts
@@ -81,6 +81,21 @@ export class Alice {
     return await policy.enact(ursulas);
   }
 
+  public async generatePolicy(
+    policyParameters: BlockchainPolicyParameters,
+    includeUrsulas?: ChecksumAddress[],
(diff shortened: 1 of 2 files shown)


piotr-roslaniec:
Hi, @ghardin1314. The `0.5.0` release candidate was initially set to fix some breaking changes expected in the upcoming `nucypher/nucypher` release. We may as well release a version that contains updated `Policy` API before that.

I can take over this PR in order to facilitate these changes.

piotr-roslaniec:
Fixed by #22

PR #49, opened 2022-06-25
Add functionality for utilising Universal Bob and God Alice

theref:
Closes https://github.com/nucypher/tdec/issues/9


cygnusv:
I love the name "Universal Bob"

derekpierre:
Perhaps it would be best to have an EPIC branch for t-Dec work that PRs like this can go into. It allows the t-Dec code changes to be self-contained within an EPIC branch for now and helps with showing partners any specific progress in the t-Dec realm.

(cc @cygnusv , @arjunhassard )

cygnusv (changes requested):
See changes requested on nucypher#2951

theref:
> See changes requested on nucypher#2951

I've changed the scheme to one that will last 10 years `duration=10*86400*365` as mentioned in https://github.com/nucypher/nucypher/pull/2951

src/characters/tDec.ts
```diff
@@ -0,0 +1,61 @@
...
+  PublicKey,
+  SecretKey,
+} from '@nucypher/nucypher-core';
+
+import { fromHexString } from '../utils';
+
+import { Enrico } from './enrico';
+import { tDecDecrypter } from './universal_bob';
+
+const tDecConfig: { [key: string]: { [key: string]: any } } = {
```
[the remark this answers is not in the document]
theref: We definitely want to move to a more robust solution. We've had a couple of brief discussions (https://discord.com/channels/411401661714792449/469956126709121026/993575795815612576). The current thinking is that we'll have a separate repo that contains the different scheme configurations.

For now I think this ok, because it is only on `ibex`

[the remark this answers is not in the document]
derekpierre: Ah yes - makes sense 👍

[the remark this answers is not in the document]
theref: this is now done

src/characters/universal_bob.ts
```diff
@@ -0,0 +1,108 @@
...
+  private readonly verifyingKey: Keyring;
+
+  constructor(
+    porterUri: string,
+    policyEncryptingKey: PublicKey,
+    encryptedTreasureMap: EncryptedTreasureMap,
+    publisherVerifyingKey: PublicKey,
+    secretKey: SecretKey,
+    verifyingKey: SecretKey
+  ) {
```
piotr-roslaniec, line 31: Just a thought:
TypeScript has a shorthand for creating and assigning a class instance property from a constructor parameter.
```suggestion
  constructor(
    private readonly porter: Porter;
    private readonly keyring: Keyring;
    private readonly policyEncryptingKey: PublicKey;
    private readonly encryptedTreasureMap: EncryptedTreasureMap;
    private readonly publisherVerifyingKey: PublicKey;
    private readonly verifyingKey: Keyring;
  ) {
```

src/characters/tDec.ts
```diff
@@ -0,0 +1,44 @@
...
+} from '@nucypher/nucypher-core';
+import axios from 'axios';
+
+import { fromHexString } from '../utils';
+
+import { Enrico } from './enrico';
+import { tDecDecrypter } from './universal_bob';
+
+async function getTDecConfig(
+  configLabel: string
```
piotr-roslaniec, line 14: What do you think about providing users with fixed configurations via `configLabel`, vs implementing a Builder pattern to give them a bit more flexibility (while still constraining them to prevent misuse)?

Just my curiosity, I'm out of the loop.

theref: i can see both being useful. A static config that you can be accessed via a `label` is quite future safe - you know that all your code/servers etc dies the policy is still valid.

Having a builder puts more responsibility on the user to keep the artifcats safe. That's not a bad thing, it gives them far more flexability.

arjunhassard: From initial feedback, it appears that flexibility trumps simplicity. tDec adopters prefer to tune their own cohort configurations and offer those 'packages' to their adopters and end-users. See nucypher/tdec/issues/1

theref:
@KPrasch @arjunhassard @cygnusv Are we happy to merge this? It's only going into `tDec-EPIC` branch, not master.
I think it will make future reviews easier and keep things compartmentalised.

cygnusv (approved):

piotr-roslaniec (approved):
LGTM :tada:

derekpierre (approved):
🎸

PR #65, opened 2022-08-02
Add builder function for dynamic tDec configurations

theref:
Tests currently don't pass.
I'm not sure of the best way to test this. Ideally I'd like to have a full integration test (all the internal components already have unit tests) - but that requires a connection to a test network or a locally running network (with all the smart contracts deployed)

Otherwise we can try and mock some stuff out, but because this is so high level I've struggled. @piotr-roslaniec @KPrasch any help greatly appreciated

### DIFF
--- a/src/characters/tDec.ts
+++ b/src/characters/tDec.ts
@@ -5,12 +5,13 @@ import {
 } from '@nucypher/nucypher-core';
 import axios from 'axios';
 import { ethers } from 'ethers';
-import { EnactedPolicy } from '../policies/policy';
 
+import { EnactedPolicy } from '../policies/policy';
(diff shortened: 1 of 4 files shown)


piotr-roslaniec:
I usually implement integration tests where interactions with external entities (network, contracts, Porter) are mocked. I just assumed that since network and contracts would be tested elsewhere, I can assume they're correct.

So I would start with mocking and we can always run end-to-end tests using a local testnet, Porter, etc. using `docker-compose` if we think that's a good idea.

I will look into it.

piotr-roslaniec (approved):
LGTM :tada:

PR #72, opened 2022-08-15
Python compatible conditions authoring and integration

vepkenez:
* conditions serialized and deserialized
* messagekits readable by latest python code
* Tests pass
* ReencryptionRequest integration in progress

### DIFF
--- a/src/characters/enrico.ts
+++ b/src/characters/enrico.ts
@@ -5,26 +5,26 @@ import { toBytes } from '../utils';
 
 const accessControlConditions = [
   {
-    contractAddress: "",
-    standardContractType: "timestamp",
-    chain: "ethereum",
-    method: "eth_getBlockByNumber",
-    parameters: ["latest"],
+    contractAddress: '',
+    standardContractType: 'timestamp',
+    chain: 'ethereum',
+    method: 'eth_getBlockByNumber',
+    parameters: ['latest'],
     returnValueTest: {
-      comparator: ">=",
-      value: "1651276942"
+      comparator: '>=',
+      value: '1651276942',
     },
   },
-  {operator: "or"},
+  { operator: 'or' },
   {
-    contractAddress: "",
-    standardContractType: "timestamp",
-    chain: "ethereum",
-    method: "eth_getBlockByNumber",
-    parameters: ["latest"],
+    contractAddress: '',
+    standardContractType: 'timestamp',
+    chain: 'ethereum',
+    method: 'eth_getBlockByNumber',
+    parameters: ['latest'],
     returnValueTest: {
-      comparator: "<=",
-      value: "1651276942"
+      comparator: '<=',
+      value: '1651276942',
     },
   },
 ];
--- a/src/characters/enrico.ts
+++ b/src/characters/enrico.ts
@@ -1,6 +1,33 @@
-import { MessageKit, PublicKey, SecretKey } from '@nucypher/nucypher-core';
+import { PublicKey, SecretKey } from '@nucypher/nucypher-core';
 
+import { MessageKit } from '../core';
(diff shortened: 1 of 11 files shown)


src/characters/enrico.ts
```diff
@@ -1,20 +1,29 @@
...
     this.verifyingKey = verifyingKey ?? SecretKey.random().publicKey();
+    if (conditions) this.conditions = conditions;
   }
 
   public encryptMessage(plaintext: Uint8Array | string): MessageKit {
     return new MessageKit(
       this.policyEncryptingKey,
-      plaintext instanceof Uint8Array ? plaintext : toBytes(plaintext)
+      plaintext instanceof Uint8Array ? plaintext : toBytes(plaintext),
+      this.conditions
```
theref, line 26: I like that you can just set conditions for Enrico when he's created, and then forget about them for all of the encryptions.

could we also include the ability to pass in a `ConditionsSet` to `encryptMessage` to make it easier to have different conditions on each encryption?

theref (changes requested):

src/characters/enrico.ts
vepkenez: Yes I can't see a downside there, certainly does add some possibilities.

test/integration/enrico.test.ts
```diff
@@ -76,4 +78,38 @@ describe('enrico', () => {
...
+    const enrico = new Enrico(policyKey, undefined, conditions);
+    const encrypted = enrico.encryptMessage(toBytes(message));
+
+    const bytes = Buffer.from(encrypted.toBytes())
+    expect(bytes).toContain(188)
+
+    const conditionbytes = ConditionsIntegrator.parse(bytes).conditionsBytes
+
+    // now take the bytes and remake a ConditionSet
+    if (conditionbytes){
```
piotr-roslaniec, line 105: But we DO expect `conditionbytes` to be defined, right?

vepkenez: no, it will be undefined if the messagekit bytes don't contain the delimiter

vepkenez: @piotr-roslaniec https://github.com/nucypher/nucypher-ts/pull/72/files#diff-033ecd338eec0994fe53c0896219baee901e20068b8fbe6259a23d5759584b06R41

vepkenez: oh, sorry I see what you mean.
I was getting annoying linting issues because of the possibility of that being undefined... so it's an accommodation in the interest of removing red lines.

piotr-roslaniec: Ah, I see, there is no branching in logic there. Please feel free to disable Prettier and ESLint and I will withhold any comments about style etc. until the PR is ready for review.

src/characters/enrico.ts
derekpierre: From an API perspective, perhaps a little nuance here...if Enrico takes a ConditionSet on initialization, from an API perspective what does the ConditionSet passed into `encryptMessage` mean? Does it mean an override of the original ConditionSet i.e. use only the ConditionSet provided as the method parameter and not the one from initialization for this message? OR a ConditionSet that is added onto the ConditionSet used for initialization for encrypting this specific message i.e. aggregation of Conditions for encrypting this message...?

In the override case, the API would seem a little obscure to me - temporarily overriding an instance variable. For such a use case are you envisioning it being too bulky or unreasonable to just initialize a new temporary Enrico for encryption that uses a different condition?

vepkenez: Yes I agree @derekpierre,  creating Enricos per conditions seemed lightweight enough to me that it made sense to do it this way for now at least.

PR #78, opened 2022-08-23
Expose dictionary types in conditions

piotr-roslaniec:
- I still have some stuff to cover, but first I want to refactor how `ConditionSet` is built. So I'm putting this PR to review now before it grows too complex.
- Used in: https://github.com/nucypher/nucypher-ts-demo-tdec/pull/2

### DIFF
--- a/src/policies/conditions.ts
+++ b/src/policies/conditions.ts
@@ -82,6 +82,21 @@ export class ConditionSet {
 }
 
 export class Condition {
+  // TODO: Shared types, move them somewhere?
+  public readonly COMPARATOR_OPERATORS = ['==', '>', '<', '>=', '<=']; // TODO: Is "!=" supported?
+  public readonly SUPPORTED_CHAINS = [
+    'ethereum',
+    // 'polygon', 'mumbai'
+  ];
+
+  protected makeReturnValueTest = () =>
+    Joi.object({
+      comparator: Joi.string()
+        .valid(...this.COMPARATOR_OPERATORS)
+        .required(),
+      value: Joi.string().required(),
+    });
+
   defaults = {};
   state = {};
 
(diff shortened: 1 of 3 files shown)


src/policies/conditions.ts
```diff
@@ -82,6 +82,21 @@ export class ConditionSet {
 }
 
 export class Condition {
+  // TODO: Shared types, move them somewhere?
```
theref, line 85: Does the recent Rust work expose these?

theref (approved):

src/policies/conditions.ts
```diff
@@ -111,39 +126,83 @@ export class Condition {
...
+    contractAddress: Joi.string()
+      .pattern(new RegExp('^0x[a-fA-F0-9]{40}$'))
+      .required(),
+    chain: Joi.string()
+      .valid(...this.SUPPORTED_CHAINS)
+      .required(),
+    standardContractType: Joi.string()
+      .valid(...this.STANDARD_CONTRACT_TYPES)
+      .required(),
+    functionAbi: Joi.string(), // TODO: Should it be required? When?
```
KPrasch, line 202: Only required when the name of a standard contract is not supplied in the field `standardContractType`.

src/policies/conditions.ts
```diff
@@ -111,39 +126,83 @@ export class Condition {
...
+
+  public readonly schema = Joi.object({
+    contractAddress: Joi.string()
+      .pattern(new RegExp('^0x[a-fA-F0-9]{40}$'))
+      .required(),
+    chain: Joi.string()
+      .valid(...this.SUPPORTED_CHAINS)
+      .required(),
+    standardContractType: Joi.string()
+      .valid(...this.STANDARD_CONTRACT_TYPES)
```
KPrasch, line 200: We'll need to source these standard ABIs from somewhere either in the release, or online.  Since this is intended to run in the browser, etherscan might be a continent place to get contract ABIs from given the address and assuming it's been verified.

src/policies/conditions.ts
KPrasch: No *new* types are introduced in nucypher core 0.3.0.

src/policies/conditions.ts
```diff
@@ -152,7 +211,7 @@ class ERC721Ownership extends ContractCondition {
     chain: 'ethereum',
     method: 'ownerOf',
     parameters: [],
-    standardContractType: 'ERC721',
+    standardContractType: 'erc721',
```
KPrasch, line 214: I guess your linter lowercased this?

src/policies/conditions.ts
```diff
@@ -111,39 +126,83 @@ export class Condition {
...
+  public readonly METHODS_PER_CONTRACT_TYPE: Record<string, string[]> = {
+    erc20: ['balanceOf'],
+    erc721: ['balanceOf', 'ownerOf'],
+    erc1155: ['balanceOf'],
+  };
+  public readonly PARAMETERS_PER_METHOD: Record<string, string[]> = {
+    // TODO: Defined using context interface?
+    balanceOf: ['address'],
+    ownerOf: ['address'],
+  };
```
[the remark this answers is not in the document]
piotr-roslaniec: I was considering adding hints to the condition builder so that the user knows what parameters they would have to provide for a method call.

src/policies/conditions.ts
piotr-roslaniec: That was actually changed by hand. I will change it back since we need to maintain compatibility between Python too.

vepkenez (approved):

PR #81, opened 2022-08-24
Add ability to rebuild dynamic tDec entities from static json

theref:
When building tDec entities dynamically, the corresponding json configuration will be returned in addition to the constructed objects.
This json can be saved and used to rebuild those entities in the future.

Fixes https://github.com/nucypher/tdec/issues/41

### DIFF
--- a/src/characters/tDec.ts
+++ b/src/characters/tDec.ts
@@ -32,7 +32,7 @@ export async function generateTDecEntities(
   endDate: Date,
   porterUri: string,
   aliceSecretKey: SecretKey = SecretKey.random()
-): Promise<readonly [Enrico, tDecDecrypter, EnactedPolicy]> {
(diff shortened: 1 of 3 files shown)


src/characters/tDec.ts
```diff
@@ -32,7 +32,7 @@ export async function generateTDecEntities(
   endDate: Date,
   porterUri: string,
   aliceSecretKey: SecretKey = SecretKey.random()
-): Promise<readonly [Enrico, tDecDecrypter, EnactedPolicy]> {
+): Promise<readonly [Enrico, tDecDecrypter, EnactedPolicy, any]> {
```
piotr-roslaniec, line 35: We can avoid using `any` here by implementing `interface TDecConfig` as type for return value

piotr-roslaniec (approved):
LGTM :tada: 

Just one comment about avoiding `any` type.

PR #86, opened 2022-09-19
SDK as defined by Milestone 5

theref:


src/sdk/cohort.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  public static fromJson({
+    ursulaAddresses,
+    threshold,
+    porterUri,
+  }: CohortJSON) {
+    return new Cohort(ursulaAddresses, threshold, porterUri);
+  }
+
+  public toJson() {
```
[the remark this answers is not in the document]
theref: fixed

src/sdk/cohort.ts
```diff
@@ -0,0 +1,79 @@
...
+export class Cohort {
+  constructor(
+    public readonly ursulaAddresses: ChecksumAddress[],
+    public readonly threshold: number,
+    public readonly porterUri: string
+  ) {
+    this.ursulaAddresses = ursulaAddresses;
+    this.threshold = threshold;
+    this.porterUri = porterUri;
+  }
```
piotr-roslaniec, line 34: If you use "field declaration as constructor parameter" syntax, you don't need to manually assign values in the constructor body - They will be automatically assigned. 
```suggestion
  ) {}
```

src/sdk/cohort.ts
```diff
@@ -0,0 +1,79 @@
...
+    shares = 0,
+    include = [],
+    exclude = [],
+  }: CohortParameters) {
+    if (shares == 0 && include.length == 0) {
+      throw new TypeError('Shares is 0 and Include is an empty array');
+    }
+    if (shares == 0 && include.length > 0) {
+      shares = include.length;
+    }
```
piotr-roslaniec, line 48: Ah, interesting.

The way `include` works in `nucypher-ts` is that it forces Porter to always contact certain Ursulas, but not exclusively. Shall we reuse this protocol, or does it make sense to change how it works?

src/sdk/cohort.ts
```diff
@@ -0,0 +1,79 @@
...
+    return new Cohort(ursulaAddresses, Number(threshold), porterUri);
+  }
+
+  public toObj() {
+    const config = {
+      ursulaAddresses: this.ursulaAddresses,
+      threshold: this.threshold,
+      porterUri: this.porterUri,
+    };
+    return config;
```
piotr-roslaniec, line 77: Sorry for this pedantic nonsense, but I'd love to shave one line of code here!
```suggestion
    return {
      ursulaAddresses: this.ursulaAddresses,
      threshold: this.threshold,
      porterUri: this.porterUri,
    };
```

src/sdk/cohort.ts
```diff
@@ -0,0 +1,79 @@
...
+    public readonly ursulaAddresses: ChecksumAddress[],
+    public readonly threshold: number,
+    public readonly porterUri: string
+  ) {
+    this.ursulaAddresses = ursulaAddresses;
+    this.threshold = threshold;
+    this.porterUri = porterUri;
+  }
+
+  public static async create({
```
piotr-roslaniec, line 36: I think we should make constructors private for classes where we provide the `create` factory method. WDYT?

src/sdk/strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+  public decrypter?: tDecDecrypter;
+  constructor(
+    public cohort: Cohort,
+    public startDate: Date,
+    public endDate: Date,
+    private alice: Alice,
+    private bob: Bob,
+    private bobSecretKey: SecretKey,
+    public deployed: boolean
+  ) {}
```
piotr-roslaniec, line 24: If we agree on having private constructors, then these fields could be defined in the `constructor`,
```suggestion
  constructor(
    public cohort: Cohort,
    public startDate: Date,
    public endDate: Date,
    private alice: Alice,
    private bob: Bob,
    private bobSecretKey: SecretKey,
    public deployed: boolean,
    public policy?: EnactedPolicy,
    public encrypter?: Enrico,
    public decrypter?: tDecDecrypter,
  ) {}
```

src/sdk/strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+    startDate: Date,
+    endDate: Date,
+    porterUri: string,
+    provider: ethers.providers.Web3Provider,
+    aliceSecretKey?: SecretKey,
+    dkgAlice?: boolean
+  ) {
+    if (dkgAlice == true) {
+      throw new TypeError('DKG Alice is not yet implemented');
+    }
```
piotr-roslaniec, line 37: If so, maybe we could simplify the method signature here? And avoid some confusion as to the availability of this feature.
```suggestion
  ) {
```

src/sdk/strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+    if (typeof aliceSecretKey == 'undefined') {
```
piotr-roslaniec, line 38: It could also be `null` - let's not forget that this library may also be used by JavaScript users. We could simplify by checking for `falsy`:
```suggestion
    if (!aliceSecretKey) {
```

src/sdk/strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+  public encrypter?: Enrico;
+  public decrypter?: tDecDecrypter;
+  constructor(
+    public cohort: Cohort,
+    public startDate: Date,
+    public endDate: Date,
+    private alice: Alice,
+    private bob: Bob,
+    private bobSecretKey: SecretKey,
+    public deployed: boolean
```
piotr-roslaniec, line 23: I'd love to avoid having flags like this, and instead, return a new type `DeployedStrategy` that would expose relevant methods.

I've tried to lobby for this design in my SDK proposal version. WDYT? Is it overdone? Or maybe too complicated at this stage of work on SDK?

src/sdk/strategy.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  public async deploy(label: string) {
+    const policyParams = {
+      bob: this.bob,
+      label,
+      threshold: this.cohort.threshold,
+      shares: this.cohort.ursulaAddresses.length,
+      startDate: this.startDate,
+      endDate: this.endDate,
+    };
```
piotr-roslaniec, line 74: Using deconstructing pattern here could be neat:

```suggestion
    const { threshold, shares } = this.cohort; // shares can be defined on Cohort as a getter method
    const { bob, startDate, endDate } = this;
    const policyParams = {
      bob,
      label,
      threshold,
      shares,
      startDate,
      endDate,
    };
```

src/sdk/cohort.ts
```diff
@@ -0,0 +1,79 @@
...
+
+import { Porter } from '../characters/porter';
+import { ChecksumAddress } from '../types';
+
+export type Ursula = {
+  readonly checksumAddress: ChecksumAddress;
+  readonly uri: string;
+  readonly encryptingKey: PublicKey;
+};
+interface CohortParameters {
```
piotr-roslaniec, line 11: I like to use `type` instead of `interface` to indicate that a type is "just a container" for data, and not a "contract" that should be implemented by classes.

```suggestion
type CohortParameters {
```

src/sdk/cohort.ts
theref: i think that's the expected behaviour. If i set include to be the set of 5 nodes that I want to use (I don't want more than that) - then Porter will only contact them and they'll be the cohort right?

src/sdk/cohort.ts
theref: This was a hangover from copy/pasting - i love getting rid of redundant lines!!!

src/sdk/cohort.ts
theref: fully agree, i hand't even considered it.

We're giving the user plenty of option with `create`, `fromJSON` etc

src/sdk/strategy.ts
theref: I'm not sure what's best. I've pinged some people more wise than myself to see if they have strong opinions on this.

How would it relate to `revoking` a strategy? Is a revoked strategy no longer deployed? is there a more general case here?....not sure, thinking out loud tbh

src/sdk/cohort.ts
piotr-roslaniec: So the cohort defines Ursulas _exclusively_ - I think that will make sense for what `Cohort` is supposed to represent,

src/sdk/strategy.ts
piotr-roslaniec: I think a revoked strategy is still technically deployed but it's no longer in effect - strategy users may not perform decryption anymore.

The way I imagined structuring the `Strategy` verbs:
```
Strategy.deploy => DeployedStrategy
DeployedStrategy.revoke => Strategy
Strategy.update => Strategy
``

File: src/sdk/cohort.ts (hunk 3 above)
Comment (piotr-roslaniec) [reply src/sdk/cohort.ts:45]: What happens if `shares > ursulas.length` thought? Where would the "extra shares" go?

File: src/sdk/strategy.ts (hunk 9 above)
Comment (theref) [reply src/sdk/strategy.ts]: Given what you've said, `DeployedStrategy.revoke => DeployStrategy` no?

Can a revoked strategy be un-revoked?

Comment (piotr-roslaniec) [reply src/sdk/strategy.ts]: > Given what you've said, DeployedStrategy.revoke => DeployStrategy no?

I'd say it's the same as `Strategy` - it's not active, and it has to be deployed before it takes an effect.

> Can a revoked strategy be un-revoked? or is it basically dead

I think once a strategy is revoked, it should be treated as "dead". And if we'd like to "bring it back" we need to redeploy it etc.

It may be a good idea to avoid using `update` verb all-together, and instead rely on `revoke` and `deploy` to get the job done. Is there any practical reason to avoid that? I could see some merit in "atomic" updates (`update` over `revoke` & `deploy` combo) since it doesn't introduce any downtime.

Comment (theref) [reply src/sdk/strategy.ts]: @KPrasch @jMyles @derekpierre @vepkenez @fjarri Any thoughts here on design best practices?

We'd like to maximise for:
- future proofing
- ease of understanding (by adopters)
- ease of testing within the library
- ease of testing in adopter code

Comment (theref) [reply src/sdk/strategy.ts]: > It may be a good idea to avoid using the update verb altogether and instead rely on revoke and deploy to get the job done. Is there any practical reason to avoid that? I could see some merit in "atomic" updates (update over revoke & deploy combo) since it doesn't introduce any downtime.

Completely agree, I have actually already removed the `update` verb. `Strategy` is immutable, once it's been created it can only be deployed (maybe that's a good +1 in favour of not having `deployed` as an attribute...)

Comment (theref) [reply src/sdk/strategy.ts]: > I'd love to avoid having flags like this, and instead, return a new type DeployedStrategy that would expose relevant methods.

We could completely flip this....

We have a `Deployer` class which has a `strategies` attribute of immutable `Strategy` objects (dict mapping label to object).
The deployer class can handle all of the `deploy`, `revoke`, `delete` verbs

File: src/sdk/cohort.ts (hunk 3 above)
Comment (KPrasch) [reply src/sdk/cohort.ts:45]: > The way include works in nucypher-ts is that it forces Porter to always contact certain Ursulas, but not exclusively. 

This is sensible, not need to change it right now.  There is room for refinement, for example, what if the *include* nodes cannot be contacted (and therefor cannot be included) but these are best suited for an issue at the current development stage.

File: src/sdk/cohort.ts [hunk 12]
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    const porter = new Porter(porterUri);
+    const ursulas = await porter.getUrsulas(shares, exclude, include);
+    const ursulaAddresses = ursulas.map((ursula) => ursula.checksumAddress);
+    return new Cohort(ursulaAddresses, threshold, porterUri);
+  }
+
+  public toJSON() {
+    const numberToStringReplacer = (key: unknown, value: unknown) =>
```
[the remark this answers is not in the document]
theref: You're right, I've removed and updated the tests.

src/sdk/cohort.ts
```diff
@@ -0,0 +1,77 @@
...
+  public static fromJSON(json: string) {
+    const config = JSON.parse(json);
+    return Cohort.fromObj(config);
+  }
+
+  public static fromObj({ ursulaAddresses, threshold, porterUri }: CohortJSON) {
+    return new Cohort(ursulaAddresses, threshold, porterUri);
+  }
+
+  public toObj() {
```
derekpierre, line 70: Is this utilized anywhere other than in the `toJSON()` function? If not, should it just be private?

src/sdk/cohort.ts
```diff
@@ -0,0 +1,77 @@
...
+  public toJSON() {
+    return JSON.stringify(this.toObj());
+  }
+
+  public static fromJSON(json: string) {
+    const config = JSON.parse(json);
+    return Cohort.fromObj(config);
+  }
+
+  public static fromObj({ ursulaAddresses, threshold, porterUri }: CohortJSON) {
```
derekpierre, line 66: Is this only used by `fromJSON`? If so, should it just be private? We don't want it to circumvent the `create` function...no?

Unless this is some typescript magic...

src/sdk/cohort.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (shares == 0 && include.length == 0) {
+      throw new TypeError('Shares is 0 and Include is an empty array');
+    }
+    if (shares == 0 && include.length > 0) {
+      shares = include.length;
+    }
+
+    const porter = new Porter(porterUri);
+    const ursulas = await porter.getUrsulas(shares, exclude, include);
+    const ursulaAddresses = ursulas.map((ursula) => ursula.checksumAddress);
```
derekpierre, line 49: Here, Ursula sampling occurs when the Cohort is created - I think this is too early. The sampling should be done just before an actual `grant(...)`. The issue is the difference in timing between when the ursulas to use for the policy are determined, and when the `grant` is to occur. These should happen sequentially and hand-in-hand, because it is not guaranteed that the sampled Ursulas will be up and running at `grant` time. 

Cohort should perhaps have a function that does this, and not do this at Cohort creation time.

src/sdk/strategy.ts
derekpierre: After reading the discussion, my initial thoughts here are:

If feels like the philosophy of the `Strategy` class is analogous to an immutable configuration, and not necessarily an object that does any execution.

As a result, I like the separate `Deployer` idea, as the thing that performs an action based on the Strategy. It also feels like encrypter/decrypter should not be members variables stored within a Strategy object.

So perhaps something like:
```
StrategyDeployer.deploy(label, strategy) -> Policy
...
StrategyDeployer.revoke(policy, strategy)
```

For the encrypter/decrypter, Strategy can then be
```
Stategy. getEncrypter(policy) -> Encrypter   # this call can check whether the policy is valid before returning

Strategy.getDecrypter(policy) -> Decrypter  # this call can check whether the policy is valid before returning
```

In this way Strategy never stores `policy`, `encrypter`, `decrypter` and is much more reusable across different policies. For example, the same Strategy can be used with different labels. 

wdyt?

src/sdk/cohort.ts
theref: not sure about this, we wanted Cohort to represent a specific set of Ursulas, which means the sampling already needs to have happened, but I see what mean about them not being available at grant time.

The idea is that I can create a cohort with the Ursulas that I want, and then re-use that forever

src/sdk/cohort.ts
```diff
@@ -0,0 +1,77 @@
...
+export type Ursula = {
+  readonly checksumAddress: ChecksumAddress;
+  readonly uri: string;
+  readonly encryptingKey: PublicKey;
+};
+
+type CohortParameters = {
+  porterUri: string;
+  threshold: number;
+  shares?: number;
```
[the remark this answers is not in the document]
theref: yes, shares is optional, because a user can just supply a list of Ursula checksums instead - at which point shares is just the length of that array

src/sdk/strategy.ts
theref: Yes, i like the idea of removing `encrypter` and `decrypter` from `Strategy`. I think those functions would also be moved to the Deployer

```
Deployer.getEncrypter(policy) -> Encrypter 
Deployer.getDecrypter(policy) -> Decrypter 
```

src/sdk/cohort.ts
theref: fixed

src/sdk/cohort.ts
theref: fixed

src/sdk/deployer.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { Strategy } from './strategy';
+
+export class Cohort {
+  public strategies: { [key: string]: Strategy };
+  private constructor(public provider: ethers.providers.Web3Provider) {
+    this.strategies = {};
+  }
+
+  public add(strategyLabel: string, strategy: Strategy) {
```
derekpierre, line 11: Do we need to store the strategies in the deployer, or can the strategy object just be passed into the relevant function (`revoke`, `deploy`, `getEncrytpter` etc.) when needed? i.e. can we prevent the Deployer from having any state?

src/sdk/cohort.ts
derekpierre: Right. right.

In a localized sense I get it, but I am still a little concerned (may/may not be warranted) about centralizing forces on specific nodes if the same "static" Cohort object is just repeatedly reused and passed around i.e. if we provide some sort of "static/globalish" set of "good", "better", "best" option, then the same nodes can be used repeatedly, as opposed to resampling ursulas when needed based on configurations provided in the Cohort object.

Need to think about it some more.

src/sdk/deployer.ts
theref: Yeah I'm definitely +1 for removing state, the user can easily handle that themselves

src/sdk/cohort.ts
arjunhassard: We didn't really delve into the sampling/re-sampling sequence in our tdec discussions, so it's reasonable in the interest of fulfilling Milestone 5 to temporarily define Cohorts as a static list of pre-specified node addresses.

However, in the medium-term, something resembling Porter's `include` logic is more collusion-resistant, more redundant, economically fairer, and confers greater customizability to the adopter. 

The sampling mechanism as a whole may need a rethink in the context of tdec, but a core principle of the work token model is not to unnecessarily disadvantage new stakers (since staker population growth mitigates centralization and supports the token price). Even though fees are currently distributed/rebated purely by stake size, concentrating actual fragment-handling work amongst early stakers (via the problem with defaults mentioned here https://github.com/nucypher/tdec/issues/51#issuecomment-1249947113), still foments a imbalance in power (i.e. if you're not on the dominant cohort lists, you have less political power even with the same size stake) and could even hasten a DAO-driven change to said fee model (to address the discordance between work & remuneration).

src/sdk/cohort.ts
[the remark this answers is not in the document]
arjunhassard: For global configs (i.e. accessible to all adopters), it makes more sense not to equate `shares` and `include_ursulas` – indeed, to move the latter to a different configuration stage so it's not being reused unknowingly/inappropriately. 

The former should be specified in all cases, because it's compatible with static lists and with more complex downstream sampling mechanisms at the time of use (adding new recipients, encrypting new data, adjusting some other parameter etc.) AND can safely be reused by other users or applications who have the same views on relative collusion-resistance. 

The latter is a bit more nuanced – the primary use of the `include_ursulas` parameter will be on a per-app basis – i.e. only relevant and appropriate for a single use case and user base. Some unrelated hackathon project prioritizing the pinging of nodes run by Masterfile's media partners would be a strange pattern to facilitate. 

See this feedback in the context of mixing-and-matching known nodes and permissionless sampling:
_Yes, overall prefer the decentralized version, as long as the option exists to hand-pick later_ https://github.com/nucypher/tdec/issues/13#issuecomment-1211954228

src/sdk/cohort.ts
arjunhassard: If it's a similar amount of work to adjust Porter to fit the static node list design, as it is to adjust the SDK's Cohort definition as per my other comments, then we should choose the longer-term option – the superior design imo being resampling at data share event and separating out `include` so it's not part of global configs. This is a fairly important decision so deserves a voice chat/debate :)

src/sdk/cohort.ts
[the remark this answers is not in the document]
arjunhassard: For global configs (i.e. accessible to all adopters), it makes more sense not to equate `shares` and `include_ursulas` – indeed, to move the latter to a different configuration stage so it's not being reused unknowingly/inappropriately. 

The former should be specified in all cases, because it's compatible with static lists and with more complex downstream sampling mechanisms at the time of use (adding new recipients, encrypting new data, adjusting some other parameter etc.) AND can safely be reused by other users or applications who have the same views on relative collusion-resistance. 

The latter is a bit more nuanced – the primary use of the `include_ursulas` parameter will be on a per-app basis – i.e. only relevant and appropriate for a single use case and user base. Some unrelated hackathon project prioritizing the pinging of nodes run by Masterfile's media partners would be a strange pattern to facilitate. 

See this feedback in the context of mixing-and-matching known nodes and permissionless sampling:
_Yes, overall prefer the decentralized version, as long as the option exists to hand-pick later_ https://github.com/nucypher/tdec/issues/13#issuecomment-1211954228

src/sdk/cohort.ts
arjunhassard: If it's a similar amount of work to adjust Porter to fit the static node list design, as it is to adjust the SDK's Cohort definition as per my other comments, then we should choose the longer-term option.

src/sdk/cohort.ts
arjunhassard: We didn't really delve into the sampling/re-sampling sequence in our tdec discussions, so it's reasonable in the interest of fulfilling Milestone 5 to temporarily define Cohorts as a static list of pre-specified node addresses.

However, in the medium-term, something resembling Porter's `include` logic is more collusion-resistant, more redundant, economically fairer, and confers greater customizability to the adopter. 

The sampling mechanism as a whole may need a rethink in the context of tdec, but a core principle of the work token model is not to unnecessarily disadvantage new stakers (since staker population growth mitigates centralization and supports the token price). Even though fees are currently distributed/rebated purely by stake size, concentrating actual fragment-handling work amongst early stakers (via the problem with defaults mentioned here https://github.com/nucypher/tdec/issues/51#issuecomment-1249947113), still foments a imbalance in power (i.e. if you're not on the dominant cohort lists, you have less political power even with the same size stake) and could even hasten a DAO-driven change to said fee model (to address the discordance between work & remuneration).

arjunhassard (commented):
Excellent work! Post-devcon and for the next milestone, the superior (more collusion-resistant, more redundant, economically fairer/safer, more customizable) design involves resampling at data share events and separating out `include` so it's not part of global configs. This has implications now and is a fairly important decision so certainly deserves a voice chat/debate :)

src/sdk/cohort.ts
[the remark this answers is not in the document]
derekpierre: Agreed. At the moment it feels like there may need to be a `CohortConfiguration` which is just configuration values and a `Cohort` that gets generated from the `CohortConfiguration` at the appropriate time. Then that `Cohort` object would contain actual Ursulas to use.

The `CohortConfiguration` could be the defined "good", "better", "best" recommendations.

I agree that include/exclude is too nuanced for global configurations (all apps). That being said Masterfile could have a its own `CohortConfiguration` that provides include/exclude for their own ecosystem.

You could allow new `CohortConfiguration` object to based on other ones -  `CohortConfiguration.fromConfiguration(global_config, <values to add/modify>)` - for example `CohortConfiguration.fromConfiguration(<other config>,  include_ursulas=[0x..., 0x...])`.

I am pro on CohortConfiguration being immutable, and new objects needing to be created to change values.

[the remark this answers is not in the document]
theref: Yeah I've slowly been persuaded :D

website/docs/CBD.md
```diff
@@ -0,0 +1,106 @@
+---
+slug: cbd
+---
+
+# Condition Based Decryption
+
+Condition Based Decryption (CBD) is a programmable access control service, in which decryption rights are based on the verified fulfillment of predefined conditions.
+Access conditions can be EVM-based (e.g. does the designated recipient own this NFT), RPC-driven (e.g. did the designated recipient commence active mining before this block) or time-based (e.g. has a preset period of time elapsed, after which the recipient's requests will be ignored).
```
cygnusv, line 8:
```suggestion
Access conditions can be EVM-based (e.g. does the requester own this NFT?), RPC-driven (e.g. has the requester sent more than N transactions?) or time-based (e.g. has a preset period of time elapsed, after which the recipient's requests will be ignored).
```
The mining example is extremely confusing, because for starters, it'd be using the `eth_mining` RPC endpoint but evaluating it for each Ursula's ETH provider client, not for the requester. I propose using an account's transaction count, which is something that it's not accessible from the EVM but that can be queried with the RPC API. 

Also the notion of "designated recipient" it's a bit restrictive, and in a way is the same angle that PRE has. Instead, we should be talking about "requester", who can be anyone that fulfills the conditions.

website/docs/CBD.md
```diff
@@ -0,0 +1,106 @@
...
+These conditions are also composable and can be combined in any logical sequence or decision tree.  
+
+CBD involves splitting the secret (a decryption key) into multiples _shares_ and distributing those among authorized and collateralized node operators ([Stakers](https://threshold.network/earn/staker) in the Threshold network).
```
cygnusv, line 11:
```suggestion
CBD involves splitting a joint secret (a decryption key) into multiples _shares_ and distributing those among authorized and collateralized node operators ([Stakers](https://threshold.network/earn/staker) in the Threshold network).
```

Also, "joint" seems to be more in line with the new codename...

website/docs/CBD.md
```diff
@@ -0,0 +1,106 @@
...
+      conditionSet: conditions
+);
+```
+
+Finally, we can deploy this Strategy to the Threshold Network, and begin using Condition Based Decryption:
+
+:::note
+
+Deploying a Strategy requires writing to the blockchain.
+This means both a funded wallet is required and also a connection to the blockain via a `provider` (eg. MetaMask).
```
cygnusv, line 80:
```suggestion
This means both a funded wallet is required and also a connection to the blockchain via a `provider` (eg. MetaMask).
```

website/docs/CBD.md
```diff
@@ -0,0 +1,106 @@
...
+
+const plaintext = 'this is a secret';
+const encryptedMessageKit = encrypter.encryptMessage(plaintext);
+
+const decryptedMessage = await decrypter.retrieveAndDecrypt(
+    [encryptedMessageKit]
+);
+```
+
+At decryption time, the user will be asked to verify their address by signing a message in MetaMask.
```
cygnusv, line 105:
```suggestion
At decryption time, the requester will be asked to verify their address by signing a message in MetaMask.
```

website/docs/Threshold Decryption.md
```diff
@@ -0,0 +1,79 @@
+---
+slug: Threshold Decryption.md
+sidebar_position: 16
+---
```
cygnusv, line 4: Should this content move to CBD.md?

website/docs/CBD.md
```diff
@@ -0,0 +1,106 @@
...
+
+Condition Based Decryption obviously also needs some [Conditions](./conditions).
+In this tutorial we will check that the user owns a specific ERC721 NFT.
+
+```js
+import { Conditions } from '@nucypher/nucypher-ts'
+
+const NFTOwnership = new Conditions.ERC721Balance({
+    contractAddress: '0xBC4CA0EdA7647A8aB7C2061c2E118A18a936f13D',
+    chain: 'rinkeby',
```
piotr-roslaniec, line 46:
```suggestion
    chain: 'Rinkeby', // Ibex uses Rinkeby testnet
```
AFAIK we expect upper-cased network names

website/docs/CBD.md
theref: :joy:

website/docs/Threshold Decryption.md
theref: yes, I've kept it around for my own reference for now :)

website/docs/CBD.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Create a Condition
+
+Condition Based Decryption obviously also needs some [Conditions](./conditions).
+In this tutorial, we will check that the user owns a specific ERC721 NFT.
+
+```js
+import { Conditions } from '@nucypher/nucypher-ts'
+
+const NFTOwnership = new Conditions.ERC721Balance({
```
derekpierre, line 44: Is this supposed to be `Conditions.ERC721Ownership`, and not `ERC721Balance`? `ERC721.ownerOf` is specific NFT ownership, and `ERC721.balanceOf` is any NFT from the contract.

Additionally `ownerOf` takes the token id ("5954") whereas `balanceOf` takes an address - https://docs.openzeppelin.com/contracts/4.x/api/token/erc721#IERC721.

website/docs/ConditionSet.md
```diff
@@ -0,0 +1,64 @@
...
+The below example shows how to authenticate that a requester owns an NFT in one of two different collections.
+
+```js
+const genuineUndead = new Conditions.ERC721Balance({
+    contractAddress: '0x209e639a0EC166Ac7a1A4bA41968fa967dB30221',
+});
+const gnomePals = new Conditions.ERC721Balance({
+    contractAddress: '0x5dB11d7356aa4C0E85Aa5b255eC2B5F81De6d4dA',
+});
+const or = new Operator('or');
```
derekpierre, line 18: Can this be static constants? like `Operator.OR`, `Operator.AND`?

website/docs/Conditions.md
```diff
@@ -0,0 +1,231 @@
...
+});
+```
+
+or:
+
+```js
+const timelockConfig = {
+    contractAddress: "",
+    standardContractType: "",
+    chain:  "ethereum",
```
derekpierre, line 89: I don't think Timelock currently takes a chain - cc @KPrasch . Unless this more along the lines of what it should be.

website/docs/Conditions.md
```diff
@@ -0,0 +1,231 @@
...
+const rpc = Conditions.Condition(rpcConfig);
+```
+
+
+## `Conditions.Condition`
+`Conditions.Condition` provides full control over the configuration of a Condition.
+It  takes parameters:
+
+- `contractAddress` is the public address of the contract we'd like to query.
+- `standardContractType` can take values from `ERC20, ERC721` and `ERC1155`. Alternatively, an ABI can be passed through if a non standard contract is being used.
```
derekpierre, line 141: There is also `functionAbi` parameter (not defined in this list) which is what is used to specify the ABI.

website/docs/Conditions.md
```diff
@@ -0,0 +1,231 @@
...
+    parameters: [
+      ":userAddress"
+    ],
+    returnValueTest: {
+      comparator: ">",
+      value: "0"
+    }
+  }
+```
+### Ownership of at least one ERC1155 token from a batch of ids
```
derekpierre, line 165: Side note: we don't yet support ERC1155

website/docs/Conditions.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Function call of nonstandard Contract
+In this example, we will check that the user is staking `T` Threshold Token.
+The Threshold staking contract is located at [`0x01B67b1194C75264d06F808A921228a95C765dd7`](https://etherscan.io/address/0x01b67b1194c75264d06f808a921228a95c765dd7#readProxyContract).
+The function we wish to call is `stakes` which takes an `address` as its parameter.
+We need to provide the `contractAddress`, `functionName`, `functionParams`, and `functionAbi` when defining the condition.
+
+```js
+const customABICondition = {
+    contractAddress: "00x01B67b1194C75264d06F808A921228a95C765dd7",
```
derekpierre, line 192:
```suggestion
    contractAddress: "0x01B67b1194C75264d06F808A921228a95C765dd7",
```

src/sdk/cohort.ts
```diff
@@ -0,0 +1,73 @@
...
+  }: CohortJSON) {
+    const config = {
+      threshold: threshold,
+      shares: shares,
+      porterUri: porterUri,
+    };
+    return new Cohort(ursulaAddresses, config);
+  }
+
+  public toObj(): CohortJSON {
```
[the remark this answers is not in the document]
piotr-roslaniec: Yes, we don't have to specify returns objects; this type is used primarily internally. It's useful for type-checking when using common TypeScript idioms, such as deconstructing, etc.

I think we should expose typed objects instead of using opaque types (dicts, objects, JSON) when we can because it significantly improves the experience for TS developers.

arjunhassard (approved):
This is really shaping up!

src/sdk/strategy.ts
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  private static fromObj({
+    cohort,
+    startDate,
+    endDate,
+    aliceSecretKeyBytes,
+    bobSecretKeyBytes,
+    conditionSet,
+  }: StrategyJSON) {
```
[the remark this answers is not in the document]
piotr-roslaniec: This is not a return type, but an application of [object deconstructing pattern](https://basarat.gitbook.io/typescript/future-javascript/destructuring#object-destructuring).

src/sdk/cohort.ts
```diff
@@ -0,0 +1,73 @@
...
+    const config = JSON.parse(json);
+    return Cohort.fromObj(config);
+  }
+
+  public static fromObj({
+    ursulaAddresses,
+    threshold,
+    shares,
+    porterUri,
+  }: CohortJSON) {
```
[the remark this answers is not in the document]
piotr-roslaniec: This is not a return type, but an application of [object deconstructing pattern](https://basarat.gitbook.io/typescript/future-javascript/destructuring#object-destructuring).

src/sdk/strategy.ts
```diff
@@ -0,0 +1,296 @@
...
+
+    if (!bobSecretKey) {
+      bobSecretKey = SecretKey.random();
+    }
+    return new Strategy(
+      cohort,
+      startDate,
+      endDate,
+      aliceSecretKey,
+      bobSecretKey,
```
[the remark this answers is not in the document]
piotr-roslaniec: Not sure I follow, is `Strategy` supposed to perform PRE in addition to CBD? In vanilla `nucypher-ts` we use `RemoteBob` which stores the public keys of `Bob`.

src/sdk/strategy.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  private static fromObj(
+    provider: ethers.providers.Web3Provider,
+    {
+      policy,
+      cohortConfig,
+      aliceSecretKeyBytes,
+      bobSecretKeyBytes,
+      conditionSet,
+    }: DeployedStrategyJSON
```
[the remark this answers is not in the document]
piotr-roslaniec: This is not a return type, but an application of [object deconstructing pattern](https://basarat.gitbook.io/typescript/future-javascript/destructuring#object-destructuring).

src/sdk/cohort.ts
```diff
@@ -0,0 +1,73 @@
...
+  public toJSON() {
+    return JSON.stringify(this.toObj());
+  }
+
+  public static fromJSON(json: string) {
+    const config = JSON.parse(json);
+    return Cohort.fromObj(config);
+  }
+
+  public static fromObj({
```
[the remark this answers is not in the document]
ShravanSunder: Consider using zod schema validation https://github.com/colinhacks/zod#instanceof

src/sdk/strategy.ts
```diff
@@ -0,0 +1,279 @@
...
+  u8ToBase64Replacer,
+} from '../utils';
+import { Web3Provider } from '../web3';
+
+import { Cohort, CohortJSON } from './cohort';
+
+type StrategyJSON = {
+  cohort: CohortJSON;
+  startDate: Date;
+  endDate: Date;
```
KPrasch, line 23: This is an artifact from a "Vanilla PRE" Alice paying and publishing policy metadata to the `SubscriptionManager` contract.   Since the start and end dates here are not expressed as conditions, and nodes do not check the `SubscriptionManager` contract for payment of policy validity during CBD evaluation, these lines have no effect.

If these strategies are intended for use with PRE *and* CBD, let's just make it optional, if it's intended for CBD only, they can be removed entirely.

src/sdk/strategy.ts
```diff
@@ -0,0 +1,279 @@
...
+    startDate: Date,
+    endDate: Date,
+    conditionSet?: ConditionSet,
+    aliceSecretKey?: SecretKey,
+    bobSecretKey?: SecretKey,
+    dkgAlice?: boolean
+  ) {
+    if (dkgAlice == true) {
+      throw new TypeError('DKG Alice is not yet implemented');
+    }
```
KPrasch: These lines can be removed entirely.

src/sdk/strategy.ts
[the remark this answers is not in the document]
KPrasch: I have similar questions - it's unclear if `Strategy` is meant for PRE and CBD?

src/sdk/strategy.ts
KPrasch: at minimum, we need a blurb in the docs to inform developers these lines have no affect.

src/sdk/strategy.ts
```diff
@@ -0,0 +1,275 @@
...
+      policy,
+      encrypter,
+      decrypter,
+      this.aliceSecretKey,
+      this.bobSecretKey,
+      this.conditionSet
+    );
+  }
+
+  public static fromJSON(json: string) {
```
KPrasch, line 118: 👍🏻

test/integration/enrico.test.ts
```diff
@@ -105,4 +105,36 @@ describe('enrico', () => {
...
+    const ownsBufficornNFT = new Conditions.ERC721Ownership({
+      contractAddress: '0x1e988ba4692e52Bc50b375bcC8585b95c48AaD77',
+      parameters: [3591],
+    });
+
+    const ownsNonsenseNFT = new Conditions.ERC721Ownership({
+      contractAddress: '0x1e988ba4692e52Bc50b375bcC8585b95c48AaD77',
+      parameters: [6969],
+    });
+
```
KPrasch, line 125: 😁

test/unit/testVariables.ts
```diff
@@ -0,0 +1,12 @@
...
+export const DeployedStrategyJSON =
+  '{"policy":{"id":{"ptr":1138600},"label":"test","policyKey":{"ptr":1158168},"encryptedTreasureMap":{"ptr":2161872},"aliceVerifyingKey":"base64:A3yU8aavNj4LJ97eFAaYpU97q70oSogzBZAlo5tj/1Kj","size":3,"startTimestamp":"2001-09-09T01:46:40.000Z","endTimestamp":"2036-07-18T13:20:00.000Z","txHash":"0x1234567890123456789012345678901234567890"},"cohortConfig":{"ursulaAddresses":["0x5cf1703a1c99a4b42eb056535840e93118177232","0x7fff551249d223f723557a96a0e1a469c79cc934","0x9c7c824239d3159327024459ad69bb215859bd25"],"threshold":2,"shares":3,"porterUri":"https://porter-ibex.nucypher.community"},"aliceSecretKeyBytes":"base64:N1K+vcukPJQkVi57P5jXca5W9CwX48VEBVV0H9CYWDU=","bobSecretKeyBytes":"base64:dOs3NGmtXJMdjVMaXf3/m5PlAmqwzSGoF9XpyO4LwZk="}';
```
KPrasch, line 4: Looks like some wasm artifacts leaked into this test, unless this is intended to be negative testing for invalid data.

KPrasch (approved):
Nice Work 💥 

There are higher-level questions about how to design for PRE back-compatibility coexisting with the new CBD functionality beyond the scope of this PR, however this is a major step forward.

src/sdk/strategy.ts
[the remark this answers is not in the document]
derekpierre: ^ The question is in regards to possibly using the same SDK object (`Strategy`) for vanilla PRE **AND** CBD, or will there eventually be separate objects (eg. `CBDStrategy` / `PREStrategy`)?

Vanilla PRE only uses bob's public key and does not use bob's secret key, so at the moment `Strategy` is only useful for CBD.

KPrasch:
Note for reviewers: documentation has been relocated to the threshold gitbook.

derekpierre:
> Note for reviewers: documentation has been relocated to the threshold gitbook.

https://app.gitbook.com/o/R2meumXNNad4y1B10iL7/s/WosjlL4zUGUMlcMfuSAp/~/changes/zPVjPQ0By8rluTA9de0O/.

Ping me if you need to be invited to the Threshold Docs Gitbook.

derekpierre:
This PR has 219 commits... 😮 ... might make sense to squash some commits together before merging.

KPrasch:
I typically don't like to squash commit, but in this situation it's the best choice.

PR #99, opened 2022-10-17
Improve configuration of post deployment objects

### LINKED ISSUE #100: Remove Alice's secret key from JSON serializers and DeployedStrategy
In an effort to reduce the exposure of sensitive secret material and prevent Alice's private keys from being accidentally transmitted (PRE-adapted proof-of-concept), Alice's secret key can be removed from `Strategy`'s JSON serializer and from `DeployedStrategy` completely. 

https://github.com/nucypher/nucypher-ts/blob/sdk/src/sdk/strategy.ts#L24
https://github.com/nucypher/nucypher-ts/blob/sdk/src/sdk/strategy.ts#L32

As a result, `Strategy` will contain private information which is not intended for transport, and `DeployedStrategy` will contain only *public information*.  Doing so potentially overlaps with the objectives in https://github.com/nucypher/nucypher-ts/pull/99.

A follow-up question:  is there any reason why a (PRE-adapted-CBD) policy author will ever need to recover Alice's secret key's *after* generating KFrags and subsequently publishing the treasure map?

theref:
- Adds the ability to serialize/deserialize decrypter to/from JSON
- Fixes #100


KPrasch (commented):
Trying to understand the developer experience here - Is this feature best understood as an alternative to the client just constructing the `decrypter` themselves?

KPrasch (commented):
related to https://github.com/nucypher/nucypher-ts/issues/100

KPrasch (dismissed):

KPrasch (commented):
Question: What is the difference between a `DeployedStrategy` and a  `tDecDecrypter` now?

theref:
> Question: What is the difference between a `DeployedStrategy` and a `tDecDecrypter` now?

`DeployedStrategy` is exactly as named, it represents a Strategy that has been deployed, so it contains the policy, encrpter, and decrypter

`tDecDecrypter` is basically bob

KPrasch:
Two questions and a few suggestions that can potentially lead to some undesigning:

##### Questions

> DeployedStrategy is exactly as named, it represents a Strategy that has been deployed, so it contains the policy ...

What is meant by "policy" here?

> ... encrpter, and decrypter

What is the motivation for including enrico and bob (and furthermore JSON serializers for both) in the strategy?

##### Suggestions

I'd like to suggest that the API can be simplified and more flexible by removing the fields `encrypter` (Enrico) and `decrypter` (Bob) from `DeployedStrategy`, both of which can be constructed by the end-user from public information already present in `DeployedStrategy`.

Simplicity is gained by removing the serialization user story for `encrypter` and `decrypter`, only requiring the dApp developers to manage a side channel for one object (`DeployedStrategy)`, and fewer lines of code for us to maintain. 

Flexibility is gained by using a unique Enrico for each `encrypter` which promotes the dApp to encrypt-and-sign using a unique ephemeral keypair. Since umbral encryption is a form a PKE, the only information needed to construct an Enrico is the policy encrypting key.

https://github.com/nucypher/nucypher-ts/blob/3e59c50aa1f31a2b671c3b31faeb2c1ae0d8f001/src/characters/enrico.ts#L5-L8

Also I noticed, `aliceVerifyingKey` is being supplied to `Enrico`'s constructor here, which appears to be a bug (I'll include another comment locally)

https://github.com/nucypher/nucypher-ts/pull/99/files#diff-a5f484f9f7dbcec2ff9ca0a3a8eeb9eb2826db7bf4659f54132bf4637efb7717L200-R188

##### Example

```javascript

// fetch and deserialize
strategy = getDeployedStrategy('https://some_url')

// message sender's browser
enrico = Enrico(strategy.policyEncryptingKey)

// message receiver's browser
decrypter = Decrypter.fromStrategy(strategy)
```

Let me know if I've misunderstood anything here!

src/sdk/strategy.ts
```diff
@@ -177,36 +168,32 @@ export class DeployedStrategy {
...
     const bobSecretKey = SecretKey.fromBytes(bobSecretKeyBytes);
     const label = newPolicy.label;
     const cohort = Cohort.fromObj(cohortConfig);
-    const porterUri = cohort.configuration.porterUri;
-    const configuration = { porterUri };
-    const alice = Alice.fromSecretKey(configuration, aliceSecretKey, provider);
     const encrypter = new Enrico(
       newPolicy.policyKey,
-      alice.verifyingKey,
+      aliceVerifyingKey,
```
KPrasch, line 188: Hmm, it looks like Alice's  (secret?) keys are unintentionally being passed to Enrico here.

KPrasch (changes requested):

src/sdk/strategy.ts
theref: `const aliceVerifyingKey = PublicKey.fromBytes(policy.aliceVerifyingKey);`

theref:
> What is meant by "policy" here?

Traditional PRE policy, the SDK is built ontop of PRE so there is a still a grant.

> Simplicity is gained by removing the serialization user story for encrypter and decrypter, requiring the dApp developers to manage a side channel for only one object (DeployedStrategy) and fewer lines of code to maintain.

dApp developers only have to manage the side channel for one object currently - the DeployedStrategy. Example is :
```
// fetch and deserialize
strategy = getDeployedStrategy('https://some_url')

// message sender's browser
enrico = strategy.encrypter

// message receiver's browser
decrypter = strategy.decrypter
```

> Flexibility is gained by using a unique Enrico for each encrypter which promotes the dApp to encrypt-and-sign using a unique ephemeral keypair. Since umbral encryption is a form a PKE, the only information needed to construct an Enrico is the policy encrypting key.

The process is still the same, build an Enrico from the DeployedStrategy info. If you're suggesting that the AliceVerifyingKey doesn't need to be included then we can remove that. I can see that it's not actually used by Encrico at all. It only appears in the constructor:

```
  constructor(
    policyEncryptingKey: PublicKey,
    verifyingKey?: PublicKey,
    conditions?: ConditionSet
  ) {
    this.policyEncryptingKey = policyEncryptingKey;
    this.verifyingKey = verifyingKey ?? SecretKey.random().publicKey();
    this.conditions = conditions;
  }
```

KPrasch:
> Traditional PRE policy, the SDK is built ontop of PRE so there is a still a grant.

My point here is that "policy" is ambiguous language in concrete terms (even in PRE); is it the HRAC? the TreasureMap? the write to SubscriptionManager? Which part of strategy the policy?  I take partial credit for this misnomer, but I think it's inaccurate.

> dApp developers only have to manage the side channel for one object currently - the DeployedStrategy.

Ah, I see. The outcomes look the same, there's 2 small but important differences I can indicate: 

#1 Internally I don't see any advantage to including JSON serialization for `encrypter` or `decrypter` if the end-user can construct it programmatically / locally.

#2 If the Deployed strategy includes a serialized pre-built enrico, every encrypter will be using the sharing the enrico keypair, which is less flexible and undesirable IMO.

> If you're suggesting that the AliceVerifyingKey doesn't need to be included then we can remove that

Indeed `Enrico.verifyingKey` is optional, but more importantly the current code passes Alice's public key to be used as Enrico's private key which effectively leaks Enrico's singing material. I don't think it's actually being used right now, but nonetheless.

The idea with Enrico is that it either generates an ephemeral random secret key, or an existing enrico private key can be passed to the constructor.  It's an easy mistake to make due to name shadowing `verifyingKey` in several places, perhaps there is a better convention.

```javascript
  constructor(
    policyEncryptingKey: PublicKey,
    verifyingKey?: PublicKey,   // <<<<<< this should be PrivateKey
    conditions?: ConditionSet
  ) {
    this.policyEncryptingKey = policyEncryptingKey;
    this.verifyingKey = verifyingKey ?? SecretKey.random().publicKey();  // <<<< enrico secret key
    this.conditions = conditions;
  }
  ```

**Comment by theref** [comment]
> 1. Internally I don't see any advantage to including JSON serialization for encrypter or decrypter if the end-user can construct it programmatically / locally.

There is nothing stopping the user from using the SDK in the manner you suggest. We just haven't written examples for it. The whole idea of the SDK was to remove as much work as possible for the adopter - we don't want them to have to build encrypters/decrypters themselves

> 2. If the Deployed strategy includes a serialized pre-built enrico, every encrypter will be sharing the same enrico keypair, which is less flexible and undesirable IMO.

Ok, I see. This comes from a misunderstanding of some of Enrico's attributes, which you've subsequently highlighted. Again, there's nothing prevent an adopter from creating a new Enrico with new ephemeral keypair if they wish too. We should add examples of this.

We settled on this design of the SDK because it removes as much work as possible from the adopter, whilst giving us the flexibility to add new functionality in the future. For example, when we change to DKG, the SDK does not change for an adopter, even though the underlying construction of an encrypter/decrypter may be different. Your design puts this responsibility on the adopter, which we've been trying to avoid.

**Comment by KPrasch** [comment]
> There is nothing stopping the user from using the SDK in the manner you suggest. We just haven't written examples for it. 

That's true, but I still don't understand the motivation for maintaining JSON serializes for Enrico and Bob.  Both are role-based entities inherent to the use case that do not *need* to be portable or transported.  All of the information needed for both is already present in the `DeployedStrategy` and their instantiation can be abstracted away for ease of use.  

> Ok, I see. This comes from a misunderstanding of some of Enrico's attributes, which you've subsequently highlighted. 

It's a bug that was not identified in prior code review, which needs to be fixed.  Shall we open an issue, or do you think you manage to include a fix in this PR?

> Again, there's nothing prevent an adopter from creating a new Enrico with new ephemeral keypair if they wish too. We should add examples of this.

I agree.  Let's make some examples of this and highlight the usage and importantly the motivation for doing so. Ideally there is only one correct way to use the SDK to keep is simple).  We may need to revisit the old debate about whether or not enrico signs message plaintext before encrypting by default, of if that's a matter of adopter choice.

> The whole idea of the SDK was to remove as much work as possible for the adopter

*The whole idea of the SDK ~~was~~ is to remove as much work as possible for the adopter*

Agreed and I believe we are achieving that goal. However there are other limiting factors to consider (an extreme example: we can pre-share private keys so no one ever needs to manage their own keys!) .  

We don't need to serialize and transport entities if they can easily be expressed locally or handle private keys on bahalf adopters if the end goal is trustless design.  Putting aside all other design factors besides the ease of usage - if the API is the same or similar in terms of adopter friction, the internal construction method that has less moving parts and fewer lines of code is superior.

>  we don't want them to have to build encrypters/decrypters themselves

On the contrary, I believe that's *precisely* what is required for an end-to-end encrypted and end-to-end decentralized solution. 
Enrico, Alice, and Bob are discrete roles and authorities in the cryptosystem's design.  If they are all generated by the same party we're effectively centralizing the trust to the strategy author.

>  We settled on this design ... 

I hope there's still room for suggestions and improvement :-)

> ... of the SDK because it removes as much work as possible from the adopter, whilst giving us the flexibility to add new functionality in the future. For example, when we change to DKG, the SDK does not change for an adopter, even though the underlying construction of an encrypter/decrypter may be different.

When we introduce the DKG Bob's keys will no longer be "universal" and must be removed from the `DeployedStrategy`.  In the DKG user-story Bob will be generated locally in order to realize full trustlessness - so it seems to me the path of least resistance is to for adopters to start creating Bob locally from the strategy now.

> Your design puts this responsibility on the adopter, which we've been trying to avoid.

Hmm, I don't understand this point.  

Since Enrico and Bob are independent authorities it's inherently their responsibility to custody their own private keys, and thus It's reasonable to expect that an encrypter instantiates an Enrico and a decrypter instantiates a Bob.  Neither of these suggestions add overhead to the adopter beyond what is inherent to the nature of the tool we're building and does not add onerous code overhead.

Personally, If I'm an implementer of nucypher-ts I'd much rather be explicitly defining characters and trusting myself:
 

```javascript

// locally generated characers

enrico = Enrico(strategy.publicKey)  // enrico created by me
...
secretKey = SecretKey.random()
bob = Bob.fromStrategy(strategy, secretKey)
```

```javascript

// remotely generated characters

enrico = strategy.enrico // enrico created by another party
...
bob = strategy.bob // bob created by another party
```

theref:
I've resolved the issues with Enrico and Alice's secret keys.


I don't think the example you posted will work, so I'm struggling to understand what you're after:
```
//
// locally generated characers
//

// fetch strategy from side-channel and deserialize
strategy = getDeployedStrategy('https://some_url')

enrico = new Enrico(strategy.publicKey)  // enrico created by me
...
secretKey = SecretKey.random()  // bob created by me
bob = Bob.fromStrategy(strategy, secretKey)
```

Bob would not be able to decrypt here because Alice has to grant to a specific Bob (due to being built on top of PRE)

KPrasch:
> I've resolved the issues with Enrico and Alice's secret keys.

Excellent, I'll dismiss that review.

> Bob would not be able to decrypt here because Alice has to grant to a specific Bob (due to being built on top of PRE)

You're right I was just being illustrative.  Until we can eliminate universal bob, we need to use the bob form the strategy.  Of course this is subject to the final API design.  It's unclear what the role of bob's private keys are in the post-PRE setting:

```javascript
// fetch strategy from side-channel and deserialize
strategy = getDeployedStrategy('https://some_url')
...
enrico = new Enrico(policyEncryptingKey=strategy.publicKey)

// ... interim API (universal Bob private keys)
bob = Bob.fromStrategy(strategy)

// ... eventual API? (local private keys)
secretKey = SecretKey.random()
bob = Bob.fromStrategy(strategy, secretKey)
```

KPrasch (approved):

PR #102, opened 2022-11-05
Improve package.json as defined by package-check

theref:
https://github.com/skypackjs/package-check

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,6 +28,9 @@ jobs:
       - name: Build
         run: yarn build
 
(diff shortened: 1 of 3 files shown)


piotr-roslaniec:
That looks like an improvement, but is this package still maintained?
Could we run it on CI?

theref:
Skypack is certainly still active, they're a big CDN provider https://www.skypack.dev/ but I'm not sure about the tool

and yes, i think it's actually designed to be a CI thing

piotr-roslaniec (approved):

PR #108, opened 2022-11-10
Chain Name -> Chain ID

KPrasch:
Fixes https://github.com/nucypher/nucypher-ts/issues/103

### DIFF
--- a/src/policies/conditions.ts
+++ b/src/policies/conditions.ts
@@ -101,10 +101,10 @@ export class Condition {
   ];
 
   public static readonly SUPPORTED_CHAINS = [
-    'Ethereum Mainnet',
-    'Görli',
-    'Polygon Mainnet',
-    'Mumbai',
+    1,     // ethereum/mainnet
+    5,     // ethereum/goerli
+    137,   // polygon/mainnet
+    80001  // polygon/mumbai
   ];
 
   public readonly schema = Joi.object();
@@ -299,7 +299,7 @@ class EvmCondition extends Condition {
 
 class ERC721Ownership extends EvmCondition {
   readonly defaults = {
-    chain: 'Görli',
+    chain: 5,  // TODO: make this value configurable
     method: 'ownerOf',
     parameters: [],
     standardContractType: 'ERC721',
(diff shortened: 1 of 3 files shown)


src/policies/conditions.ts
```diff
@@ -101,10 +101,10 @@ export class Condition {
...
 
   public static readonly SUPPORTED_CHAINS = [
-    'Ethereum Mainnet',
-    'Görli',
-    'Polygon Mainnet',
-    'Mumbai',
+    1,     // ethereum/mainnet
+    5,     // ethereum/goerli
+    137,   // polygon/mainnet
+    80001  // polygon/mumbai
```
piotr-roslaniec, line 107: We could also use `ethers` to provide these values: https://docs.ethers.io/v5/api/providers/types/#providers-Network

KPrasch: Nice idea but I'm having a difficult time making it work/

```js
private static readonly SUPPORTED_CHAIN_NAMES = [
    'mainnet',
    'goerli',
    'matic',
    'maticm'
];

 public static readonly SUPPORTED_CHAINS = Condition.SUPPORTED_CHAIN_NAMES.map(chainName =>
    ethers.providers.getNetwork(chainName).chainId
);
```
    
This seems to work only for `mainnet`, despite the docs stating that it works for other networks too.  Is this API meant only for connected providers?

theref: i have used this package before https://www.npmjs.com/package/evm-chains - it's excellent.
4K weekly downloads, it's the standard way of doing things

KPrasch: After some more thought, I don't see any advantage to using a lookup by name string versus a chain ID.  Either way we're maintaining a hardcoded list of identifiers Chain IDs are less ambiguous than the network naming conventions implemented by the package maintainer.

theref: completely agree, everything internal should be chain id
i guess that package i posted is really for ui stuff, possibly useful on the demo so that a user knows they're not connected on mainnet

derekpierre (approved):
🎸

PR #113, opened 2022-11-22
Update deprecated ibex testnet in Readme

manumonti:
Ibex testnet has been deprecated. Tapir is the new adopter's testnet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ The equivalent decrypter can be created by:
 import { makeTDecDecrypter } from '@nucypher/nucypher-ts'
 
 // you can specify your own porter url here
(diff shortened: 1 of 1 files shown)


piotr-roslaniec:
There are [other usages](https://github.com/nucypher/nucypher-ts/search?q=ibex) of `ibex` that we may want to remove from repo

theref:
The whole README is actually out of date now, great find. We probably want to use some of the basic stuff from the gitbook for the README

Also, let's merge into master not tDec-EPIC

manumonti:
Since the scope of this is bigger than initially expected, I have split this PR into two issues: 

https://github.com/nucypher/nucypher-ts/issues/115

https://github.com/nucypher/nucypher-ts/issues/116

Issue #116, opened 2022-11-25
Update references to IBEX network in code


manumonti:
There are some references to the deprecated ibex network in nucypher-ts code.

https://github.com/nucypher/nucypher-ts/search?q=ibex

piotr-roslaniec:
Closed by #118

PR #118, opened 2022-11-29
Remove default chain id and references to ibex

### LINKED ISSUE #114: Default Chain ID value in conditions
I found a little bit confusing the default values taken in the [Conditions](https://docs.threshold.network/app-development/threshold-access-control-tac/references/conditions) doc:

There is no reference to what network will be taken by default. If you inspect the code, the Göerli testnet will be taken. So I'm wondering myself:

Should we have a default value for this? Or should be mandatory to specify the chain ID when creating the Conditions object? What should be the default value in the former case, Ethereum mainnet or Göerli?

_Originally posted by @manumonti [and discussion] in https://github.com/nucypher/tdec/issues/86#issuecomment-1323621431_

theref:
Fixes #114 and #116

### DIFF
--- a/src/policies/conditions.ts
+++ b/src/policies/conditions.ts
@@ -311,7 +311,6 @@ class ERC721Ownership extends EvmCondition {
 
 class ERC721Balance extends EvmCondition {
   readonly defaults = {
-    chain: 5, // TODO: make this value configurable
     method: 'balanceOf',
     parameters: [':userAddress'],
     standardContractType: 'ERC721',
--- a/src/policies/conditions.ts
+++ b/src/policies/conditions.ts
@@ -299,7 +299,6 @@ class EvmCondition extends Condition {
 
 class ERC721Ownership extends EvmCondition {
   readonly defaults = {
-    chain: 5, // TODO: make this value configurable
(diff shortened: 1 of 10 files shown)


src/policies/conditions.ts
```diff
@@ -299,7 +299,6 @@ class EvmCondition extends Condition {
 
 class ERC721Ownership extends EvmCondition {
   readonly defaults = {
-    chain: 5, // TODO: make this value configurable
```
derekpierre, line 302: Should this change also be repeated on L314 for `ERC721Balance`

test/integration/conditions.test.ts
```diff
@@ -37,6 +37,13 @@ describe('conditions schema', () => {
     expect(result.error).toEqual(undefined);
     expect(result.value.chain).toEqual(5);
   });
+
+  it('should validate chain id', async () => {
+    result = condition.validate({ chain: 10 });
```
piotr-roslaniec, line 42: We should contain all variable declarations such as `result` to the `it(...)` closure where they are being used. This prevents state-leaking across different tests.

PR #123, opened 2022-12-06
Enable function ABI in EVM conditions

piotr-roslaniec:
- Tested manually on Lynx
- Created issue: https://github.com/nucypher/nucypher-ts/issues/124
- Closes https://github.com/nucypher/tdec/issues/99


theref (approved):

src/policies/conditions.ts
```diff
@@ -288,13 +288,14 @@ class EvmCondition extends Condition {
...
-    // functionAbi: Joi.string().optional(),
+      .optional(),
+    functionAbi: Joi.object().optional(),
     method: this.makeMethod().required(),
     parameters: Joi.array().required(),
     returnValueTest: this.makeReturnValueTest(),
-  });
+  })
+    // At most one of these keys needs to be present
+    .xor('standardContractType', 'functionAbi');
```
derekpierre, line 298: 🔥

test/integration/conditions.test.ts
```diff
@@ -220,3 +220,51 @@ describe('condition context', () => {
     expect(asJson).toBeDefined();
   });
 });
+
+describe('evm condition', () => {
```
derekpierre, line 224: 👍 - nice test.

src/policies/conditions.ts
```diff
@@ -288,13 +288,14 @@ class EvmCondition extends Condition {
       .required(),
     standardContractType: Joi.string()
       .valid(...EvmCondition.STANDARD_CONTRACT_TYPES)
-      .required(),
-    // TODO: Support custom function ABIs
-    // functionAbi: Joi.string().optional(),
+      .optional(),
+    functionAbi: Joi.object().optional(),
```
derekpierre, line 292: Question Is there any way to validate a `functionAbi` dict that is provided?

derekpierre: Apologies, you already filed an issue for it - https://github.com/nucypher/nucypher-ts/issues/124. Nvm

derekpierre (approved):
🎸

src/policies/conditions.ts
piotr-roslaniec: Yes, we can use `joi` for that. It's not going to be too complicated, but I opted to address it in a separate PR since it's not critical for the release.

I'm tracking this issue here: https://github.com/nucypher/nucypher-ts/issues/124

derekpierre: Makes sense 👍

PR #129, opened 2022-12-07
Release 0.9.0

piotr-roslaniec:
-  Update examples after v0.9.0 release

### DIFF
--- a/examples/react-craco/package.json
+++ b/examples/react-craco/package.json
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "@craco/craco": "^6.4.3",
-    "@nucypher/nucypher-ts": "^0.7.0",
+    "@nucypher/nucypher-ts": "../..",
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^11.1.0",
     "@testing-library/user-event": "^12.1.10",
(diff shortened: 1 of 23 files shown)


examples/react-craco/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "@craco/craco": "^6.4.3",
-    "@nucypher/nucypher-ts": "^0.7.0",
+    "@nucypher/nucypher-ts": "0.9.0-alpha.1",
```
theref, line 15: no longer alpha i think

examples/react-webpack-5-experiments/package.json
```diff
@@ -9,8 +9,9 @@
     "build": "tsc --noEmit && rimraf build && webpack --mode production --progress"
   },
   "dependencies": {
+    "ethers": "^5.4.1",
     "file-loader": "^6.2.0",
-    "@nucypher/nucypher-ts": "^0.7.0",
+    "@nucypher/nucypher-ts": "0.9.0-alpha.1",
```
theref, line 14: same here

theref (commented):
Looks good, but we possible need to split it into two PRs.
1. Release 0.9.0
2. Upgrade all the examples to use the new Release

piotr-roslaniec:
So the reason why examples are using `0.9.0-alpha.1` is that it's pretty much compatible with `0.9.0`. I'm going to update them after releasing the v0.9.0 package on npmjs.com

theref (approved):

PR #135, opened 2022-12-08
refactor!: Pass provider to retireveAndDecrypt instead of ConditionContext

ghardin1314:
When calling `tDecDecrypter.retrieveAndDecrypt()` the current api takes in `messageKits` and `conditionContext`. The problem with this is that `messageKits` already contains `condition`(s) and there is no guarantee that `conditionContext` contains the same condition set.

What I think the real intent here is to check if any of the `condition`(s) include a `:userAddress` and therefore require a signature. This can be done just by passing a provider to this method.

This is a proposal of how this could be accomplished, but a more thorough rework of `ConditionContext` may be warranted to be cleaner

### DIFF
--- a/src/policies/conditions.ts
+++ b/src/policies/conditions.ts
@@ -1,5 +1,5 @@
 import { Conditions as WASMConditions } from '@nucypher/nucypher-core';
-import { utils as ethersUtils } from 'ethers';
+import { ethers, utils as ethersUtils } from 'ethers';
 import Joi, { ValidationError } from 'joi';
 
 import { Eip712TypedData, FormattedTypedData, Web3Provider } from '../web3';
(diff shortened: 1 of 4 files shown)


src/policies/conditions.ts
```diff
@@ -459,9 +444,9 @@ export class ConditionContext {
   }
 
   public toJson = async (): Promise<string> => {
-    const userAddressParam = this.contextParameters.find(
```
ghardin1314, line 462: I think this was just checking to see if `:userAddress` was included in any of the conditions. If I am wrong about this please let me know

piotr-roslaniec: Yes, that's correct.

PR #137, opened 2022-12-08
Expose startDate and endDate as strategy params

### LINKED ISSUE #134: Add ability to override default strategy `startDate` and `endDate`
This may just be for training wheels in the early days, but it would required be able to configure for time frames different than a month in production

https://github.com/nucypher/nucypher-ts/blob/ab2b0bec47ca25aca0d3f0e474f44b65823a7f35/src/sdk/strategy.ts#L73-L74

piotr-roslaniec:
- Fixes #134


src/sdk/strategy.ts
```diff
@@ -70,8 +89,8 @@ export class Strategy {
       label,
       threshold: this.cohort.configuration.threshold,
       shares: this.cohort.configuration.shares,
-      startDate: new Date(Date.now()),
-      endDate: new Date(Date.now() + 1000 * 60 * 60 * 24 * 30),
+      startDate: this.startDate,
+      endDate: this.endDate,
```
jdbertron, line 93: Thank you.

PR #139, opened 2022-12-08
Replace `Web3Provider` with direct usage of `ethers.provider.Web3Provider`

piotr-roslaniec:
- Fixes: https://github.com/nucypher/nucypher-ts/issues/130
- This is a breaking change in the API, so we'd best wait for a `MAJOR` release and update all relevant examples, docs, etc.

### DIFF
--- a/src/agents/subscription-manager.ts
+++ b/src/agents/subscription-manager.ts
@@ -10,13 +10,12 @@ import {
   SubscriptionManager__factory,
 } from '../../types/ethers-contracts';
 import { ChecksumAddress } from '../types';
-import { Web3Provider } from '../web3';
 
 import { DEFAULT_WAIT_N_CONFIRMATIONS, getContracts } from './contracts';
 
 export class SubscriptionManagerAgent {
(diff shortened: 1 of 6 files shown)


theref (commented):
This is becoming a bit more relevant as I start to sketch out the DKG api design. A web3 connection will obviously be required for all ritual related activities https://github.com/nucypher/nucypher-ts/issues/166

I have an open question there about how much we want to restrict the user (force the use of ethers in the background or allow them to use whatever library they like).

Either way, I like the `connectReadWrite` and `connectReadOnly` functions, we should possibly move them into a `utils` file so we can use them elsewhere in the repo

piotr-roslaniec:
I'm okay with restricting users to using `ethers` for now. If at some point they indicate a need for other web3 libs, we could easily add the `Web3Provider` facade back.

If you need to get any changes in, please let me know and we can finalize this PR.

theref (approved):

theref:
it's good for me!

Issue #140, opened 2022-12-08
Retire payment stub from TAC SDK


piotr-roslaniec:
- Remove dependency on policy creation with `SubscriptionManager`
- Mandates update of examples & docs
- This is a `MAJOR` breaking change 
- RFC: @KPrasch

piotr-roslaniec:
@arjunhassard Do we have a green light to remove `SubscriptionManager`?

piotr-roslaniec:
Notes from a discussion on Discord voice:
- Refactor policy to remove dependency on `SubscriptionManager` in CBD flow, and keep it as it is in PRE flow
- Consider future payment schemes for CBD
- Consider refactoring `nucypher-ts` into a monorepo

piotr-roslaniec:
Closed by #210

PR #141, opened 2022-12-08
Revert "Expose startDate and endDate as strategy params"

piotr-roslaniec:
- Reverts nucypher/nucypher-ts#137
- Opting to avoid changes in the API if another change is implemented #140  
- Awaits a design decision in #140


theref:
Why is this being reverted? surely it will disappear anyway once DKG is live and policies are no longer relevant to CBD

piotr-roslaniec:
This issue was created a few months back when we didn't have clarity on the policies we have now. 

These won't be relevant after the #197 merge, so we may as well close it now.

PR #143, opened 2022-12-12
Allow value of ReturnValueTest to be string or number

### LINKED ISSUE #142: Create conditions errors in code
I'm not sure if this is a documentation issue or a nucypher-ts bug.

If we check [Conditions.RpcCondition docs](https://docs.threshold.network/app-development/threshold-access-control-tac/references/conditions#conditions.rpccondition), we see there are two ways of creating this condition. But both of them return errors if used as described in docs:

1. Using RpcCondition:
```js
const const rpc = new Conditions.RpcCondition({
  chain: 1,
  method: 'eth_getBalance',
  parameters: [
      ":userAddress",
      "latest"
    ],
  returnValueTest: {
    comparator: ">=",
    value: 10000000000000
  }
});
```

2. Using Condition:
```js
const rpcConfig = {
  contractAddress: '',
  standardContractType: '',
  chain: 1,
  method: 'eth_getBalance',
  parameters: [':userAddress', 'latest'],
  returnValueTest: {
    comparator: '>=',
    value: 10000000000000,
  },
};
const rpc = Conditions.Condition(rpcConfig);
```

If we use in the code the former one, we get an error:

> "returnValueTest.value" must be a string

This error is fixed if we pass the Object key `value` as string instead of as number:

```js
  returnValueTest: {
    comparator: ">=",
    value: "10000000000000"
  }
  ```
  
  However, in the latter option, we can use both, string and number types, for the key `value`. 
  
  **Should we homogenize this behavior by allowing only one type (string or number)?**
  
  Another issue is that the latter option returns error:
  
  > Error: Request failed with status code 500
  
  It only works if we delete the two first keys of the config object:
  
  ```js
  const rpcConfig = {
    contractAddress: '', // DELETE THIS
    standardContractType: '', // DELETE THIS
    chain: 1,
    method: 'eth_getBalance',
    parameters: [':userAddress', 'latest'],
    returnValueTest: {
      comparator: '>=',
      value: 10000000000000,
    },
};
  ```
  
Also, we have the same problems in [Conditions.TimelockCondition](https://docs.threshold.network/app-development/threshold-access-control-tac/references/conditions#conditions.timelockcondition).

theref:
Fixes #142

Don't merge until we're happy with our test suite


manumonti (approved):
Looks good ✌️

piotr-roslaniec:
So either a number or a string is accepted - is it in line with how `nucypher` uses these values? I.e. are "return test" values from different tests treated as numbers and strings, or are they all converted into a string?

theref:
> So either a number or a string is accepted - is it in line with how `nucypher` uses these values? I.e. are "return test" values from different tests treated as numbers and strings, or are they all converted into a string?

in `nucypher` : `value = fields.Raw(allow_none=False, required=True)` so numbers and strings are both ok. There isn't any conversion done

derekpierre:
> in nucypher : value = fields.Raw(allow_none=False, required=True) so numbers and strings are both ok.

The value is "Raw" because `value` can be any type returned by web3 API (not just strings and numbers). For example, when a struct is returned from Solidity, or a bool, or bytes etc. See https://github.com/nucypher/nucypher/blob/development/tests/unit/conditions/test_return_value.py#L226. The test is possibly excessive, but proves the point.

piotr-roslaniec (approved):
Ok, so assuming that the type conversion from a JS type to JSON, and from JSON to a corresponding Python type works, this PR is in order

derekpierre:
> Ok, so assuming that the type conversion from a JS type to JSON, and from JSON to a corresponding Python type works, this PR is in order

Since `nucypher-ts` only deals with condition creation side of things, and takes JSON I think so, but just wanted to be clear.

derekpierre:
> > Ok, so assuming that the type conversion from a JS type to JSON, and from JSON to a corresponding Python type works, this PR is in order

Something is nagging me in my mind about this 😅 . Can we try to test out `value` being a boolean and converted to JSON on the nucypher-ts side, then processed by python processing, to check whether this works? It's as simple as generating JSON with `value` as a boolean from nucypher-ts, then using that JSON in a python test, just to be sure.


Side note: unrelated to this PR, but I noticed the value here - https://github.com/nucypher/nucypher-ts/blob/main/src/policies/conditions.ts#L320 - should not be a string, but rather an integer. Can we get that changed?

ghardin1314:
> > > Ok, so assuming that the type conversion from a JS type to JSON, and from JSON to a corresponding Python type works, this PR is in order
> 
> Something is nagging me in my mind about this sweat_smile - especially since `Joi.boolean()` is a possible type...can we try to test out `value` being a boolean and converted to JSON on the nucypher-ts side, then processed by python processing, to check whether this works? It's as simple as generating JSON with `value` as a boolean from nucypher-ts, then using that JSON in a python test, just to be sure.
> 
> Side note: unrelated to this PR, but I noticed the value here - https://github.com/nucypher/nucypher-ts/blob/main/src/policies/conditions.ts#L320 - should not be a string, but rather an integer. Can we get that changed?

Just ran into this string <> integer issue as well. I think `nucypher` should accept both strings and integers here since it is very easy to have a `tokenId` (or `paymentAmount`, etc) be a number larger than the maximum integer value of javascript (uint64 i believe)

derekpierre:
> > > Ok, so assuming that the type conversion from a JS type to JSON, and from JSON to a corresponding Python type works, this PR is in order
> 
> Something is nagging me in my mind about this 😅 - especially since `Joi.boolean()` is a possible type...can we try to test out `value` being a boolean and converted to JSON on the nucypher-ts side, then processed by python processing, to check whether this works? It's as simple as generating JSON with `value` as a boolean from nucypher-ts, then using that JSON in a python test, just to be sure.
> 
> Side note: unrelated to this PR, but I noticed the value here - https://github.com/nucypher/nucypher-ts/blob/main/src/policies/conditions.ts#L320 - should not be a string, but rather an integer. Can we get that changed?

^ @theref / @piotr-roslaniec let's see if we can resolve this PR.

PR #147, opened 2023-01-06
Add documentation tests

manumonti:
Documentation tests are intended for early detection of modifications on nucypher-ts that involve changes on the code examples in docs.

### DIFF
--- a/test/docs/cbd.test.ts
+++ b/test/docs/cbd.test.ts
@@ -1,3 +1,4 @@
+import { VerifiedKeyFrag } from '@nucypher/nucypher-core';
 import { providers } from 'ethers';
 
 import {
@@ -7,9 +8,16 @@ import {
   SecretKey,
   Strategy,
 } from '../../src';
+import { Ursula } from '../../src/characters/porter';
+import { toBytes } from '../../src/utils';
 import {
+  mockDetectEthereumProvider,
+  mockEncryptTreasureMap,
+  mockGenerateKFrags,
   mockGetUrsulas,
+  mockMakeTreasureMap,
   mockPublishToBlockchain,
+  mockRetrieveCFragsRequest,
   mockUrsulas,
   mockWeb3Provider,
 } from '../utils';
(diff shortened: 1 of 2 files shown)


test/docs/cbd.test.ts
```diff
@@ -0,0 +1,151 @@
...
+    );
+    expect(conditions.validate()).toEqual(true);
+
+    // Start of 4. Build a Strategy
+    const newStrategy = Strategy.create(newCohort, conditions);
+
+    const MMprovider = await detectEthereumProvider();
+    const mumbai = providers.getNetwork(80001);
+
+    if (MMprovider) {
```
piotr-roslaniec, line 89: The closing bracket for this `if` statement seems to be misplaced: Should probably be at line 91

manumonti: You are right, in docs (https://docs.threshold.network/app-development/threshold-access-control-tac/get-started-with-tac#4.-build-a-strategy) the closing bracket is placed after the newStrategy.deploy() statement.

But if we place there (line 91), the variables `web3Provider` and `newDeployed` will not be accessible after this line, so the subsequent use of these two variables (line 95, 111, 140...) results in compilation errors.

I have been thinking about this and my best idea is to take this little trick, but other ideas are pretty welcome.

piotr-roslaniec: My two cents:

You could create a `setup` function where you run a complete scenario as shown in docs, and then return all intermediate variables. Then, to test each documentation segment, you call `setup` and get any variable you need. You could move assertions into this `setup` function, and leave the code examples in the test body.

That would also have the added benefit of making code examples easy to copy and paste into docs (because they only contain the relevant code).

manumonti: > 

Definitely, I'm in with this idea. Working on it.

Although the closing bracket problem remains since we still need these two variables (`web3Provider` and `newDeployed`) after line 91 (lines 138 and 140).

manumonti: Another option could be to simply delete this condition from docs. I mean, delete `if (MMprovider) {` line.

This is a tutorial that shows the basic usage of our code. Maybe this kind of checking is not necessary.

What do you think @piotr-roslaniec ?

piotr-roslaniec: I think it's fine since it's a simplified (idealized) example. I.e. we're only showing a green path.

manumonti: > My two cents:
> 
> You could create a `setup` function where you run a complete scenario as shown in docs, and then return all intermediate variables. Then, to test each documentation segment, you call `setup` and get any variable you need. You could move assertions into this `setup` function, and leave the code examples in the test body.
> 
> That would also have the added benefit of making code examples easy to copy and paste into docs (because they only contain the relevant code).

Done in https://github.com/nucypher/nucypher-ts/pull/147/commits/b5f37b26b6687f04682485ddd354ec655480ab9a

piotr-roslaniec (approved):
Good job :tada:

@manumonti I took the liberty of slightly refactoring this code example. Please feel free to change it as you see fit and merge it when ready

manumonti:
> Good job 🎉
> 
> @manumonti I took the liberty of slightly refactoring this code example. Please feel free to change it as you see fit and merge it when ready

Now it looks great! Thank you

PR #193, opened 2023-04-26
Bump `nucypher-core` version

piotr-roslaniec:
- Updates to `nucypher-core@0.6.0`


test/unit/strategy.test.ts
```diff
@@ -163,8 +163,8 @@ describe('Deployed Strategy', () => {
   });
 
   it('can encrypt and decrypt', async () => {
```
piotr-roslaniec, line 165: Fails with error:
```
decryption failed: Decryption of ciphertext failed: either someone tampered with the ciphertext or you are using an incorrect decryption key.
```
on line 215:
```
const decryptedMessage = await decrypter.retrieveAndDecrypt(
  [encryptedMessageKit],
  bobProvider
);
```
This test worked before the update. It doesn't use any premade test data (`testVariables.ts`). 
@theref Does that ring any bells?

piotr-roslaniec: Looks like the state is leaking between the tests due to uncleared `jest.spy`.

test/unit/strategy.test.ts
```diff
@@ -120,16 +122,16 @@ describe('Deployed Strategy', () => {
...
     porterUri: 'https://_this.should.crash',
   };
-  const aliceSecretKey = SecretKey.fromBytes(aliceSecretKeyBytes);
-  const bobSecretKey = SecretKey.fromBytes(bobSecretKeyBytes);
+  const aliceSecretKey = SecretKey.fromBEBytes(aliceSecretKeyBytes);
+  const bobSecretKey = SecretKey.fromBEBytes(bobSecretKeyBytes);
 
   afterEach(() => {
-    jest.clearAllMocks();
+    jest.restoreAllMocks();
```
piotr-roslaniec, line 129: This was the culprit. See [this excellent](https://stackoverflow.com/questions/58151010/difference-between-resetallmocks-resetmodules-resetmoduleregistry-restoreallm) StackOverflow answer.

PR #198, opened 2023-05-10
Support `index` in `returnValueTest`

### LINKED ISSUE #138: Add support for user-defined parameters in `returnValueTest`
- Support changes from https://github.com/nucypher/nucypher/pull/3032
-  `returnTestValue` is expecting the following structure:
```
{
  "comparator": ">",
  "value": 100,
  "index": 0,
}
```
Where the `index` is the index of some element in the array returned by the `returnTest` that we want to use with the `comparator`: 
 ```
 returnTest()[index] > 100
 ```

piotr-roslaniec:
- Closes #138


theref (approved):
No you're correct, it was a simple fix - let's get it in

src/policies/conditions.ts
```diff
@@ -118,6 +118,7 @@ export class Condition {
 
   protected makeReturnValueTest() {
     return Joi.object({
+      index: Joi.number().required(),
```
derekpierre, line 121: Index is not a required parameter...https://github.com/nucypher/nucypher/pull/3032/files#diff-39ed7645b1d8386d5199d3de0d4bca1d523549567147e8c8c3f4f685b1f14da3R90

derekpierre:
I think (?) from the `nucypher-ts` side it is really just allowing it to be specified as part of the JSON - the work of actually validating/utilizing it is within the scope of the python/Ursula-code.

PR #199, opened 2023-05-10
Add support for user-defined params in `ConditionContext`

### LINKED ISSUE #173: Add support for user defined parameters in `ConditionContext`
I think we discussed some potential updates to the `ConditionContext` api (and concept in general) previously. One thing the should be added is support for user defined parameter inputs. 

`Condition`(s) allow for user supplied parameters via a `:<parameter_name>` (i.e. `:tokenId`) but there is currently no way to supply these at decryption time as the context returns only `:userAddress` when making a request to porter here: 

https://github.com/nucypher/nucypher-ts/blob/3c4d2f6186bee3e036e9f8c4d6f713e0d94a9857/src/policies/conditions.ts#L461-L471

Example of support from nucypher core library

https://github.com/nucypher/nucypher/blob/development/tests/acceptance/blockchain/conditions/test_conditions.py#L213

This is my current hack to make it work on my test project

```typescript
class MasterfileConditionConditionContext extends ConditionContext {
  private superToJson = this.toJson;

  constructor(
    conditionSet: ConditionSet,
    web3Provider: Web3Provider,
    private readonly customParams: Record<string, any>
  ) {
    super(conditionSet, web3Provider as any);
  }

  toJson = async (): Promise<string> => {
    const params = JSON.parse(await this.superToJson());
    return JSON.stringify({ ...this.customParams, ...params });
  };
}
```

piotr-roslaniec:
- Closes #173
- Includes a small refactor of tests methods

### DIFF
--- a/src/policies/conditions.ts
+++ b/src/policies/conditions.ts
@@ -192,6 +192,9 @@ const makeGuard = (
   return schema;
 };
 
+export const USER_ADDRESS_PARAM = ':userAddress';
+export const SPECIAL_CONTEXT_PARAMS = [USER_ADDRESS_PARAM];
+
 class TimelockCondition extends Condition {
   public static readonly CONDITION_TYPE = 'timelock';
 
@@ -216,8 +219,8 @@ class RpcCondition extends Condition {
     string,
     string[]
   > = {
-    eth_getBalance: [':userAddress'],
-    balanceOf: [':userAddress'],
+    eth_getBalance: [USER_ADDRESS_PARAM],
+    balanceOf: [USER_ADDRESS_PARAM],
   };
 
   public readonly schema = Joi.object({
@@ -258,7 +261,7 @@ class EvmCondition extends Condition {
   public static readonly METHODS_PER_CONTRACT_TYPE: Record<string, string[]> = {
     ERC20: ['balanceOf'],
     ERC721: ['balanceOf', 'ownerOf'],
-    ERC1155: ['balanceOf'],
+    // ERC1155: ['balanceOf'], // TODO(#131)
   };
   public static readonly PARAMETERS_PER_METHOD: Record<string, string[]> = {
     balanceOf: ['address'],
@@ -268,8 +271,8 @@ class EvmCondition extends Condition {
     string,
     string[]
   > = {
-    balanceOf: [':userAddress'],
-    ownerOf: [':userAddress'],
+    balanceOf: [USER_ADDRESS_PARAM],
+    ownerOf: [USER_ADDRESS_PARAM],
   };
 
   private makeMethod = () =>
@@ -306,7 +309,7 @@ class ERC721Ownership extends EvmCondition {
     returnValueTest: {
       index: 0,
       comparator: '==',
-      value: ':userAddress',
+      value: USER_ADDRESS_PARAM,
     },
   };
 }
@@ -314,7 +317,7 @@ class ERC721Ownership extends EvmCondition {
 class ERC721Balance extends EvmCondition {
   readonly defaults = {
     method: 'balanceOf',
-    parameters: [':userAddress'],
+    parameters: [USER_ADDRESS_PARAM],
     standardContractType: 'ERC721',
     returnValueTest: {
       index: 0,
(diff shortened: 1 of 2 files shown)


theref (approved):
Looks awesome!

KPrasch (dismissed):

theref (approved):
Looks awesome!

derekpierre (dismissed):
See prior comment.

test/integration/conditions.test.ts
```diff
@@ -21,20 +29,18 @@ describe('operator', () => {
...
   it('should validate', async () => {
     expect(result.error).toEqual(undefined);
-    expect(result.value.contractAddress).toEqual(
-      '0xC36442b4a4522E871399CD717aBDD847Ab11FE88'
-    );
+    expect(result.value.contractAddress).toEqual(TEST_CONTRACT_ADDR);
   });
 
-  result = condition.validate({ chain: 5 });
+  result = condition.validate({ chain: TEST_CHAIN_ID });
```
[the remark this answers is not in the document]
piotr-roslaniec: Sorry, this is unrelated to context variables. This is part of the test refactor mentioned in the issue description. 

I will refrain from comingling unrelated changes such as these in future PRs, seeing how it causes confusion (second time today).

src/policies/conditions.ts
```diff
@@ -443,16 +457,34 @@ export class ConditionContext {
...
+          throw new Error(`Missing custom context parameter ${key}`);
+        }
+      });
+    });
+
+    Object.keys(this.customParameters).forEach((key) => {
+      payload[key] = this.customParameters[key];
+    });
     return JSON.stringify(payload);
   };
```
piotr-roslaniec, line 481: @derekpierre Here are the changes that are most relevant to this PR: 
- Seperating variables into  `SPECIAL_CONTEXT_PARAMS` with reserved names
- Adding support for a user-provided map of custom context params

test/integration/conditions.test.ts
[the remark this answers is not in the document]
derekpierre: Oh. No, my bad then - apologies

test/integration/conditions.test.ts
```diff
@@ -262,7 +267,75 @@ describe('condition context', () => {
...
+    const evmCondition = new Conditions.EvmCondition({
+      chain: 5,
+      functionAbi: fakeFunctionAbi,
+      method: 'balanceOf',
+      contractAddress: '0x0000000000000000000000000000000000000000',
+      parameters: [USER_ADDRESS_PARAM, ':customParam'],
+      returnValueTest: {
+        index: 0,
+        comparator: '==',
+        value: USER_ADDRESS_PARAM,
```
derekpierre, line 303: Does a user-defined variable work here (with the returnValuetest) instead of `USER_ADDRESS_PARAM`? eg. if you instead had:

```suggestion
        value: ':customParam',
```

I saw the code for processing when it is in params, but I haven't seen it for return value test objects - I may have missed it somewhere (my typescript is not great 😅 ).

piotr-roslaniec: Yes, it does work here. IMO this is a desirable behavior - being able to provide custom parameters for predefined variables. 

I'm adding a test to document that behavior.

derekpierre:
Had a conversation with @piotr-roslaniec about some potential bugs in the PR for him to investigate:

1. When using context variables for arbitrary contracts they may not be correctly checked, since they are only checked for certain methods, and not arbitrary methods: <https://github.com/nucypher/nucypher-ts/blob/04e87729bf611a60106e08fafe7888a562dbb853/src/policies/conditions.ts#L154>
2. If a list of method parameters in the Condition contains a combination of context variables and actual values, the ConditionContext.toJson() function may fail - <https://github.com/nucypher/nucypher-ts/blob/04e87729bf611a60106e08fafe7888a562dbb853/src/policies/conditions.ts#L467>
3. When verifying that custom parameters used in the Condition have an associated entry in the ConditionsContext, custom variables used in the ReturnValueTest.value are not checked; only parameters are checked - https://github.com/nucypher/nucypher-ts/blob/04e87729bf611a60106e08fafe7888a562dbb853/src/policies/conditions.ts#L467

In general, we should increase testing coverage to catch some of these cases and potentially others.

piotr-roslaniec:
@derekpierre I'd like to address these issues in a separate PR if that sounds okay

derekpierre:
> @derekpierre I'd like to address these issues in a separate PR if that sounds okay

Yep. Sounds good. Just be sure to link the follow-up issue/PR to this PR.

PR #202, opened 2023-05-11
Refactor conditions into modules & classes

piotr-roslaniec:
- Introduces conditions module module in `./src/condition` directory
- Re-organizes package-level exports, `./src/index.ts`
- Refactors shared logic if `Condition` and its subclasses
- Refactors how conditions are implemented
- Separates conditions into `base` and `predefined` conditions
- No functional changes
- Note for reviewers:
  - This PR introduces breaking changes in user-facing API, specifically the package exports. To inspect these changes, see `./src/index.ts` and examples in the test. This is essentially a proposal for a new, module-based API structure. It should improve code separation and further work on CBD and PRE packaging.
- Questions: 
   - Should `Condition` be placed in `./src/conditions/base/Condition`?
   - Is using a mixin in `EvmCondition` and `RpcCondition` an overkill? With more `base` conditions being introduced (`MultichainCondition`, etc.) it should be easier to manage them with mixins.


test/docs/cbd.test.ts
```diff
@@ -1,13 +1,7 @@
...
 import { providers } from 'ethers';
 
-import {
-  Cohort,
-  Conditions,
-  ConditionSet,
-  SecretKey,
-  Strategy,
-} from '../../src';
+import { Cohort, conditions, SecretKey, Strategy } from '../../src';
```
piotr-roslaniec, line 4: An example of a module import

test/docs/cbd.test.ts
```diff
@@ -22,6 +16,12 @@ import {
   mockRetrieveCFragsRequest,
 } from '../utils';
 
+const {
+  predefined: { ERC721Ownership },
+  ConditionSet,
+  Condition,
+} = conditions;
+
```
piotr-roslaniec, line 24: This destructuring shows the contents of the module import

theref (approved):

theref:
I really like this organisation. I think as the Conditions functionality grows the modules will be a big help

manumonti (approved):
LGTM! 🙌

manumonti (approved):
LGTM! 🙌

KPrasch:
> Should Condition be placed in ./src/conditions/base/Condition?

I generally like this convention of root/base classes in a module named `base`.

> Is using a mixin in EvmCondition and RpcCondition an overkill? With more base conditions being introduced (MultichainCondition, etc.) it should be easier to manage them with mixins.

Hey if mixins are well supported I'd like to see something like that implemented (I did the same in nucypher/nucypher) but it's certainly not the only way get the level of encapsulation you're after :-)

KPrasch (approved):

src/index.ts
```diff
@@ -1,28 +1,39 @@
...
-export {
-  ConditionContext,
-  CustomContextParam,
-} from './conditions/condition-context';
 
+// Conditions
+import type { CustomContextParam } from './conditions';
+import * as conditions from './conditions';
+// TODO: Not sure how to re-export this type from the conditions module
+export { conditions, CustomContextParam };
```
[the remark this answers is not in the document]
piotr-roslaniec: The lowercase elements are modules (or objects), and the uppercase are classes

PR #207, opened 2023-05-15
Refactor conditions test suite

piotr-roslaniec:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Refactor
- [ ] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
- Refactors tests into a new directory structure
- Adds some missing test cases
- Adds Codecov to CI

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
- Test structure for condition-related tests contained redundancies and lacked structure
- It wasn't easy to review and assess the completeness of the test suite

**Notes for reviewers:**
- This is primarily a refactor and should not introduce any breaking changes to an API
- Function changes (if any) will be pointed out in the comments
- Codecov will not comment yet, since there is no baseline coverage uploaded yet (this is the first time we're uploading it)


src/conditions/condition-set.ts
```diff
@@ -11,20 +11,25 @@ export class ConditionSet {
   constructor(public readonly conditions: ReadonlyArray<ConditionOrOperator>) {}
 
   public validate() {
+    // Expects [Condition, Operator, Condition, Operator, ...], where the last element is a Condition
+
     if (this.conditions.length % 2 === 0) {
       throw new Error(
-        'conditions must be odd length, ever other element being an operator'
+        'conditions must be odd length, every other element being an operator'
```
piotr-roslaniec, line 18: In JavaScript, changing the text of an error message is not generally considered a breaking change. Nevertheless, I want to point this out as a functional change.

test/unit/conditions/context.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+      const asJson = await conditionContext
+        .withCustomParams(customParams)
+        .toJson();
+      expect(asJson).toBeDefined();
+      expect(asJson).toContain(myCustomParam);
+    });
+  });
+
+  // TODO: Fix this test. Fails with '"method" must be [balanceOf]'
```
piotr-roslaniec, line 99: Lack of proper validation is already tracked here: https://github.com/nucypher/nucypher-ts/issues/124

test/unit/testVariables.ts
```diff
@@ -63,3 +63,30 @@ export const decrypterJSON = JSON.stringify({
     'base64:A3yU8aavNj4LJ97eFAaYpU97q70oSogzBZAlo5tj/1Kj',
   bobSecretKeyBytes: 'base64:dOs3NGmtXJMdjVMaXf3/m5PlAmqwzSGoF9XpyO4LwZk=',
 });
+
+export const TEST_CONTRACT_ADDR = '0x0000000000000000000000000000000000000001';
+export const TEST_CONTRACT_ADDR_2 =
+  '0x0000000000000000000000000000000000000001';
```
manumonti, line 69: do we want that TEST_CONTRACT_ADDR have the same (fake) address than TEST_CONTRACT_ADDR_2?

Just asking because in the original condition test file (conditions.test.ts), we had this:

```
const TEST_CONTRACT_ADDR = '0xC36442b4a4522E871399CD717aBDD847Ab11FE88';
const TEST_CONTRACT_ADDR_2 = '0x5dB11d7356aa4C0E85Aa5b255eC2B5F81De6d4dA';
```

manumonti (approved):
LGTM! ✌️

I just left a question there.

test/unit/conditions/base/timelock.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    const badTimelockObj = {
+      // Intentionally replacing `returnValueTest` with an invalid test
+      returnValueTest: {
+        ...returnValueTest,
+        comparator: 'not-a-comparator',
+      },
+    };
+
+    const badTimelock = new TimelockCondition(badTimelockObj);
+    expect(() => badTimelock.toObj()).toThrow(
```
derekpierre, line 30: Given this test - any idea how this slipped through, https://discord.com/channels/866378471868727316/870383717418410066/1107101659399729192? Just curious really.

test/unit/testVariables.ts
piotr-roslaniec: Nope, that's a mistake. These should be different, i.e. the second one should be `0x0000000000000000000000000000000000000002`

test/unit/conditions/base/timelock.test.ts
piotr-roslaniec: Nothing stops anyone from breaking this code really. You can always go `{randomJson: true} as Condition` or hardcode some values into Porter client. It's best we ask them for a code example to reproduce this.

test/unit/testVariables.ts
derekpierre: Something to think about here is that if mistakenly having them be the same address did not produce a test failure, is the second one actually needed as a constant. May be / may be not i.e. perhaps the second one can just be a local variable.

derekpierre (approved):
🎸

PR #223, opened 2023-06-12
Validate `functionAbi`-related fields

### LINKED ISSUE #124: In EVM condition, validate `functionAbi` against `parameters` and `method`
- `EvmCondition` builder should extract the method name and method parameters from `functionAbi`and validate them against user-provided `parameters` and `method`

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
- 2

**What this does:**
- Adds basic validation of `EvmCondition.functionAbi` against other fields, `EvmCondition.method` and `EvmCondition.parameters` 

**Issues fixed/closed:**
- Closes #124

**Why it's needed:**
- Currently, there is no way to prevent the misuse of `functionAbi`

**Notes for reviewers:**
- See comments in the PR


src/conditions/base/evm.ts
```diff
@@ -7,6 +7,39 @@ import { returnValueTestSchema } from './schema';
...
+  type: Joi.string().required(),
+});
+
+const functionAbiSchema = Joi.object({
+  name: Joi.string().required(),
+  type: Joi.string().valid('function').required(),
+  inputs: Joi.array().items(functionAbiVariable),
+  outputs: Joi.array().items(functionAbiVariable),
+  // TODO: Should we restrict this to 'view'?
+  // stateMutability: Joi.string().valid('view').required(),
```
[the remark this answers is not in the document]
theref: That requires Ursula to spend ETH on gas costs, which we almost certainly want to avoid

src/conditions/base/evm.ts
```diff
@@ -7,6 +7,39 @@ import { returnValueTestSchema } from './schema';
 
 export const STANDARD_CONTRACT_TYPES = ['ERC20', 'ERC721'];
 
+const functionAbiVariable = Joi.object({
+  internalType: Joi.string().required(),
+  name: Joi.string().required(),
+  type: Joi.string().required(),
```
piotr-roslaniec, line 13: Should we also validate the type against a dictionary of values, such as[ `uint256`, `address`](https://docs.soliditylang.org/en/develop/abi-spec.html#types)?

I think we either need to draw a line somewhere or reuse existing ABI validator (to avoid re-implementing it from scratch here).

src/conditions/base/evm.ts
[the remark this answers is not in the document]
piotr-roslaniec: Validating here will only prevent misuse but not bad behavior. If this is a risk, then we should also check the state mutability on Ursula's side.

src/conditions/base/evm.ts
piotr-roslaniec: Same applies to `value` in `returnValueTest` (present in all `Condition`s)

src/conditions/base/evm.ts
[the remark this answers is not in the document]
derekpierre: Ursula only performs `eth_call` which is only for `read` contract calls, so presumably the call would fail if an actual transaction was needed. (cc @KPrasch )

src/conditions/base/evm.ts
derekpierre: Can the web3 library be utilized here in some way instead of us redefining the schema?

piotr-roslaniec: We could try using web3.js to try [to encode a function call](https://web3js.readthedocs.io/en/v1.2.11/web3-eth-abi.html#encodefunctioncall) and so to validate the `functionAbi`. Alternatively, use [a different parser](https://www.npmjs.com/package/@findeth/abi).

How is `functionAbi` validation done on the Python side? If the encoding (function call) fails, we can always return an error from Python. Similarly, we can allow for early execution of conditions on the client side for the purpose of validation.

Maybe this is the way to go? Parsing instead of validating + execution on the client side?

derekpierre: We indeed return an error on the python-side. It's a bit different of the python-side because it doesn't construct the JSON (other than for tests), but rather receives it.

The validation on the python-side is done here during object initialization after the JSON is received and is being converted into a ContractCondition object - https://github.com/nucypher/nucypher/blob/development/nucypher/policy/conditions/evm.py#L31.

piotr-roslaniec: After [our brief exchange on Discord](https://discord.com/channels/411401661714792449/412681039178366996/1118100730860032010) I'm leaning towards leaving a minimal verification here and revisiting this subject when scoping a related issue for `nucypher-core`

derekpierre: Agreed. I think validate the best you can. 

 `nucypher-core` work is not happening for `7.0.0`.

PR #226, opened 2023-06-15
[WIP] Support prefix notation for compound conditions

### LINKED ISSUE #224: Prefix notation for logical condition operators
Related to:
- https://github.com/nucypher/nucypher/pull/3158

Ursula-side implemenation done via https://github.com/nucypher/nucypher/pull/3140. `nucypher-ts` will need to adjust the logical operator schema appropriately.

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Implements prefix notation for compound conditions in `nucpyher-ts`. 

Related to:
- https://github.com/nucypher/nucypher/issues/3058
- https://github.com/nucypher/nucypher/pull/3140

**Issues fixed/closed:**
> - Fixes #...

Fixes #224 .

**Why it's needed:**

**Notes for reviewers:**


src/conditions/condition-set.ts
```diff
@@ -1,68 +1,53 @@
...
 
-  public static fromConditionList(list: ReadonlyArray<Record<string, string>>) {
-    return new ConditionSet(
-      list.map((ele: Record<string, string>) => {
-        if ('operator' in ele) return Operator.fromObj(ele);
-        return Condition.fromObj(ele);
-      })
-    );
+    return new ConditionSet(condition);
   }
```
piotr-roslaniec, line 51: We could try a different approach here:
```
  public static fromObj(obj: ConditionSetJSON): ConditionSet {
    // version specific logic can go here
    const underlyingConditionData = obj.condition;

    if (underlyingConditionData.operator) {
      return new ConditionSet(new CompoundCondition(underlyingConditionData));
    }

    if (underlyingConditionData.method === BLOCKTIME_METHOD) {
      return new ConditionSet(new TimeCondition(underlyingConditionData));
    }

    if (underlyingConditionData.contractAddress) {
      return new ConditionSet(new ContractCondition(underlyingConditionData));
    }

    if ((underlyingConditionData.method as string).startsWith('eth_')) {
      return new ConditionSet(new RpcCondition(underlyingConditionData));
    }

    throw new Error('Invalid condition: unrecognized condition data');
  }
  ```

Comment (derekpierre) [reply src/conditions/condition-set.ts]: ✔️

File: src/conditions/base/contract.ts [hunk 2]
```diff
@@ -2,12 +2,12 @@ import Joi from 'joi';
...
 import { ETH_ADDRESS_REGEXP } from '../const';
 
-import { RpcCondition, rpcConditionSchema } from './rpc';
+import { RpcCondition, rpcConditionRecord } from './rpc';
 
 export const STANDARD_CONTRACT_TYPES = ['ERC20', 'ERC721'];
 
 const functionAbiVariable = Joi.object({
-  internalType: Joi.string().required(),
+  internalType: Joi.string(), // TODO is this needed?
```
piotr-roslaniec, line 10: I think yes, working on a related change here: https://github.com/nucypher/nucypher-ts/pull/228

test/unit/conditions/base/contract.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    );
+
+    const { error } = badEvm.validate(badContractCondition);
+    expect(error?.message).toEqual('"contractAddress" is required');
+  });
+});
+
+describe('accepts either standardContractType or functionAbi but not both or none', () => {
+  const standardContractType = 'ERC20';
+  const functionAbi = {
```
piotr-roslaniec, line 30: There's a `testFunctionAbi` in `testVariables.ts` if we want to reuse it.

test/unit/conditions/base/contract.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    );
+  });
+});
+
+// TODO(#124)
+// it('accepts custom parameters in function abi methods', async () => {
+//     throw new Error('Not implemented');
+// });
+
+// TODO(#124)
```
piotr-roslaniec, line 102: Hmm those two should be fixed by now

test/unit/conditions/condition-set.test.ts
```diff
@@ -1,75 +1,270 @@
...
-      expect(setFromJson.conditions.length).toEqual(1);
-      expect(setFromJson.conditions[0]).toBeInstanceOf(Condition); // TODO: This should arguably be an ERC721Balance
+    it.each([
+      erc721BalanceCondition,
+      contractConditionNoAbi,
+      contractConditionWithAbi,
+      rpcCondition,
+      timeCondition,
+      compoundCondition,
+    ])('serializes to and from json', async (condition) => {
```
piotr-roslaniec, line 69: Exquisite

test/unit/conditions/base/contract.test.ts
derekpierre: This was the file that was removed in `tdec-epic` by a different PR. I re-added it and renamed it after a rebase. It is uncommented, and slightly modified now.

test/unit/conditions/condition-set.test.ts
```diff
@@ -1,75 +1,270 @@
...
+    // condition with wrong method "method" and no contract address
+    {
+      ...testTimeConditionObj,
+      method: 'doWhatIWant',
+    },
+    // rpc condition (no contract address) with disallowed method
+    {
+      ...testRpcConditionObj,
+      method: 'isPolicyActive',
+    },
```
piotr-roslaniec, line 104: What do you think about using permutations here to exhaustively test for every possible incorrect variant? Nothing from with the current implementation, I'm just interested in your thoughts.

piotr-roslaniec (approved):

test/unit/conditions/condition-set.test.ts
derekpierre: I'm intrigued - could you elaborate a bit more on what you are thinking? Just want to make sure I understand.

The goal for this test was to test the `ConditionSet.fromObj(...)` branching logic for invalid data.

piotr-roslaniec: We could generate a set of all possible permutations of the form `<field>: <value>`, where both `field` and `value` may contain valid and invalid values. From this set we remove permutations we know to be correct. Others should fail the validation, and so the test cases would be successful.

I think without a property-based testing library this is tiresome to implement. I did consider this type of testing in other places on different occasions.

theref (approved):

PR #228, opened 2023-06-16
Validate `functionAbi` using `ethers` ABI parser

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
-  1

**What this does:**
- Uses `ethers` to validate `functionABI`

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
- We've implemented somewhat basic validation for `functionABI`, but it's still possible to pass mangled/illegal JSON ABI into your conditions

**Notes for reviewers:**
- I need some input on this - It looks like there is going to be some overlap between `Joi` and `ethers`, and we have to implement some checks and balances ourselves on top of that.

### DIFF
--- a/src/conditions/base/contract.ts
+++ b/src/conditions/base/contract.ts
@@ -40,12 +40,16 @@ const functionAbiSchema = Joi.object({
   // Now we just need to validate against the parent schema
   // Validate method name
   const method = helper.state.ancestors[0].method;
-  const abiMethodName = Object.keys(asInterface.functions).find((name) =>
-    name.startsWith(`${method}(`)
-  );
-  const functionFragment = abiMethodName
-    ? asInterface.functions[abiMethodName]
-    : null;
+
+  let functionFragment;
+  try {
+    functionFragment = asInterface.getFunction(method);
+  } catch (e) {
+    return helper.message({
+      custom: `"functionAbi" contains ambiguous "${method}"`,
+    });
+  }
+
   if (!functionFragment) {
     return helper.message({
       custom: `"functionAbi" not valid for method: "${method}"`,
--- a/src/conditions/base/contract.ts
+++ b/src/conditions/base/contract.ts
@@ -46,7 +46,7 @@ const functionAbiSchema = Joi.object({
     functionFragment = asInterface.getFunction(method);
   } catch (e) {
     return helper.message({
-      custom: `"functionAbi" contains ambiguous "${method}"`,
+      custom: `"functionAbi" has no matching function for "${method}"`,
     });
   }
 
--- a/src/conditions/base/contract.ts
+++ b/src/conditions/base/contract.ts
@@ -1,3 +1,4 @@
+import { ethers } from 'ethers';
 import Joi from 'joi';
 
 import { ETH_ADDRESS_REGEXP } from '../const';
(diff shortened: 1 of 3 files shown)


theref (approved):
Great change

src/conditions/base/contract.ts
```diff
@@ -6,31 +7,51 @@ import { RpcCondition, rpcConditionRecord } from './rpc';
...
+  try {
+    asInterface = new ethers.utils.Interface([functionAbi]);
+  } catch (e: unknown) {
+    const { message } = e as Error;
+    return helper.message({
+      custom: message,
+    });
+  }
+
+  if (!asInterface.fragments) {
```
derekpierre, line 28: Should we be using `.functions` instead of `.fragments`? Just reading from here - https://docs.ethers.org/v5/api/utils/abi/interface/#Interface--properties.

Same for various logic below.

src/conditions/base/contract.ts
```diff
@@ -6,31 +7,51 @@ import { RpcCondition, rpcConditionRecord } from './rpc';
...
   // Validate method name
   const method = helper.state.ancestors[0].method;
-  if (functionAbi.name !== method) {
+  const functionFragment = asInterface.fragments.filter(
+    (f) => f.name === method
+  )[0];
+  if (!functionFragment) {
     return helper.message({
-      custom: '"method" must be the same as "functionAbi.name"',
+      custom: '"functionAbi" does not contain the method specified as "method"',
```
derekpierre, line 48: Can we output the actual method name. So something like (?):
```suggestion
      custom: `"functionAbi" not valid for method, "${method}"`,
```

src/conditions/base/contract.ts
```diff
@@ -6,31 +7,51 @@ import { RpcCondition, rpcConditionRecord } from './rpc';
...
     });
   }
 
   // Validate nr of parameters
   const parameters = helper.state.ancestors[0].parameters;
-  if (functionAbi.inputs?.length !== parameters.length) {
+  if (functionFragment.inputs.length !== parameters.length) {
```
derekpierre, line 54: 👍 Nice

test/unit/conditions/base/contract.test.ts
```diff
@@ -7,7 +7,7 @@ import {
 import { ContractCondition } from '../../../../src/conditions/base';
```
derekpierre, line 7: I'd love to see some comprehensive testing around validation of various different functionAbis, both valid and invalid, as part of this PR.

derekpierre (changes requested):
See comment about testing.

src/conditions/base/contract.ts
piotr-roslaniec: :heavy_check_mark:

piotr-roslaniec: Yes, we use them to fetch the `FunctionFragments`

src/conditions/base/contract.ts
```diff
@@ -6,31 +7,54 @@ import { RpcCondition, rpcConditionRecord } from './rpc';
...
+  // Now we just need to validate against the parent schema
   // Validate method name
   const method = helper.state.ancestors[0].method;
-  if (functionAbi.name !== method) {
+  const abiMethodName = Object.keys(asInterface.functions).find((name) =>
+    name.startsWith(`${method}(`)
+  );
+  const functionFragment = abiMethodName
+    ? asInterface.functions[abiMethodName]
+    : null;
```
derekpierre, line 48: Could this be reduced to the following (https://docs.ethers.org/v5/api/utils/abi/interface/#Interface--fragments)?

```suggestion
  const functionFragment = asInterface.getFunction(method);
```

It seems to possible throw an `ArgumentError` - do we care?

piotr-roslaniec: I think it makes sense given our conversation about the ambiguity of functions. Fixed.

src/conditions/base/contract.ts
```diff
@@ -6,31 +7,58 @@ import { RpcCondition, rpcConditionRecord } from './rpc';
...
   // Validate method name
   const method = helper.state.ancestors[0].method;
-  if (functionAbi.name !== method) {
+
+  let functionFragment;
+  try {
+    functionFragment = asInterface.getFunction(method);
+  } catch (e) {
+    return helper.message({
+      custom: `"functionAbi" contains ambiguous "${method}"`,
```
derekpierre, line 49: 👍 - minor nitpick

```suggestion
      custom: `"functionAbi" has no matching function for "${method}"`,
```

derekpierre (approved):
🎸 - good stuff. Minor nitpick suggestion.

PR #229, opened 2023-06-19
Versioned Conditions

### LINKED ISSUE #225: Ensure that versioned condition language is supported
Related to:
- https://github.com/nucypher/nucypher/issues/2977 (Issue)
- https://github.com/nucypher/nucypher/pull/3145 (PR)

Implemented on python side. Will need to adjust accordingly in `nucypher-ts`.

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Adds a version field to serialized conditions. This facilitates backwards-compatibility for any subsequent protocol changes.

Related to:
- https://github.com/nucypher/nucypher/issues/2977
- https://github.com/nucypher/nucypher/pull/3145

**Issues fixed/closed:**
Fixes #225 .

**Why it's needed:**

**Notes for reviewers:**


src/conditions/condition-expr.ts
```diff
@@ -0,0 +1,108 @@
...
+export class ConditionExpression {
+  static VERSION = '1.0.0';
+
+  constructor(
+    public readonly condition: Condition,
+    public readonly version: string = ConditionExpression.VERSION
+  ) {}
+
+  public toObj(): ConditionExpressionJSON {
+    // TODO add version here
```
piotr-roslaniec, line 30: Is this TODO relevant to this PR?

derekpierre: Ah yes, need to remove that TODO.

src/conditions/condition-expr.ts
```diff
@@ -0,0 +1,108 @@
...
+  TimeCondition,
+} from './base';
+import { BLOCKTIME_METHOD } from './base/time';
+import { CompoundCondition } from './compound-condition';
+import { ConditionContext } from './context';
+
+export type ConditionExpressionJSON = {
+  version: string;
+  condition: Record<string, unknown>;
+};
```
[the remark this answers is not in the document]
derekpierre: Does that limit the allowed properties as a whole, or ensures that at least one of those property values is present?

src/conditions/condition-expr.ts
```diff
@@ -0,0 +1,108 @@
...
+      condition: conditionData,
+    };
+  }
+
+  public static fromObj(obj: ConditionExpressionJSON): ConditionExpression {
+    const version = obj.version;
+    // version specific logic can go here
+    const receivedMajorVersion = version.split('.')[0];
+    const currentMajorVersion = ConditionExpression.VERSION.split('.')[0];
+    if (receivedMajorVersion > currentMajorVersion) {
```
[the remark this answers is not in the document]
piotr-roslaniec: We're comparing strings here, so `>` will be resolved using lexicographical order, but I suppose this is fine for us.

Accessing string operation results with `[0]` may fail silently. That could lead to a situation where we evaluate `undefined > "1"` or similar undesired behaviors. Perhaps we could validate version strings with `Regexp` to ensure that they are valid SemVers before we `split` them.

src/conditions/condition-expr.ts
```diff
@@ -0,0 +1,108 @@
...
+      version: this.version,
+      condition: conditionData,
+    };
+  }
+
+  public static fromObj(obj: ConditionExpressionJSON): ConditionExpression {
+    const version = obj.version;
+    // version specific logic can go here
+    const receivedMajorVersion = version.split('.')[0];
+    const currentMajorVersion = ConditionExpression.VERSION.split('.')[0];
```
piotr-roslaniec, line 42: This is not an improvement, but I wanted to highlight an alternative way to access `split` results using deconstruction patterns.

```suggestion
    const [major, _minor, _fix] = version.split('.');
```

src/conditions/condition-expr.ts
```diff
@@ -0,0 +1,108 @@
...
+
+      if ((underlyingConditionData.method as string).startsWith('eth_')) {
+        return new ConditionExpression(
+          new RpcCondition(underlyingConditionData),
+          version
+        );
+      }
+    }
+
+    throw new Error('Invalid condition: unrecognized condition data');
```
piotr-roslaniec, line 81: With nested constructors, this section is getting rather dense. Returning early may not increase readability here. Perhaps we could reconsider this structure? WDYT?

```suggestion
    const underlyingConditionData = obj.condition;
    let condition: Condition;

    if (underlyingConditionData.operator) {
      condition = new CompoundCondition(underlyingConditionData);
    } else if (underlyingConditionData.method) {
      if (underlyingConditionData.method === BLOCKTIME_METHOD) {
        condition = new TimeCondition(underlyingConditionData);
      } else if (underlyingConditionData.contractAddress) {
        condition = new ContractCondition(underlyingConditionData);
      } else if ((underlyingConditionData.method as string).startsWith('eth_')) {
        condition = new RpcCondition(underlyingConditionData);
      }
    }
    if (!condition) {
      throw new Error(`Invalid condition: unrecognized condition data ${JSON.stringify(underlyingConditionData)}`);
    }
    return new ConditionExpression(condition, version);
```

src/conditions/condition-expr.ts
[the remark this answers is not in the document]
piotr-roslaniec: It only defines a set of legal keys in `Record`. So if the key is present, it has to be of the given type, but if there are no keys it's still considered a valid object. I.e. `{}` is a valid `condition` field.

It's essentially adding type-checking, it does nothing for run-time validation.

src/conditions/condition-expr.ts
derekpierre: Thanks. Looks like there is a `SemVer` object I can use - https://www.npmjs.com/package/semver. I'll try it out.

src/conditions/condition-expr.ts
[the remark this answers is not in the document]
piotr-roslaniec: I think you're right, I didn't consider other keys. So the `ConditionProps` should be defined as follows:

```
type ConditionProps = 'operator' | 'method' | 'contractAddress' | string; // or replacing string with all other possible keys
```

If we use `string`, I don't think this addition would be useful enough to consider, since `'operator' | 'method' | 'contractAddress' | string;` simplifies to just `string`.

`Record<string, unknown>` shows up in a couple of different places in the codebase, so it may be worthwhile (from the perspective of idiomatic TypeScript, etc.) to turn it into a custom `type`. But perhaps this is something we could look into another time.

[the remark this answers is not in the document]
derekpierre: I think ultimately then we would be repeating the work of the underlying condition JOI schema. 

Perhaps we can use the JOI schema fields defined by the various underlying conditions instead of listing them out ourselves. But that can be considered in a separate issue/PR.

src/conditions/condition-expr.ts
derekpierre: ✔️

src/conditions/condition-expr.ts
derekpierre: ✔️

src/conditions/condition-expr.ts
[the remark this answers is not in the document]
derekpierre: ✔️

src/conditions/condition-expr.ts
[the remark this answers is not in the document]
derekpierre: Filed https://github.com/nucypher/nucypher-ts/issues/230.

piotr-roslaniec (approved):
LGTM :tada:

test/unit/conditions/condition-expr.test.ts
```diff
@@ -0,0 +1,423 @@
...
+      expect(conditionExprJson).toContain(toJSON(condition.toObj()));
+
+      const conditionExprFromJson =
+        ConditionExpression.fromJSON(conditionExprJson);
+      expect(conditionExprFromJson).toBeDefined();
+      expect(conditionExprFromJson.equals(conditionExprFromJson)).toBeTruthy();
+    });
+
+    it('incompatible version', async () => {
+      const invalidVersion = '100.0.0';
```
manumonti, line 188:
```suggestion
      const invalidVersion = '0.1.0';
```

Admittedly a nitpick, but there is a possibility that, someday, we will reach this version `100.0.0` and then this test will fail.

Using a version lower than the first version (`1.0.0` -> `0.1.0`) ensures that this will never fail.

WDYT?

manumonti (approved):
Just a little suggestion, but LGTM! 👌

test/unit/conditions/condition-expr.test.ts
derekpierre: The goal of the test is for the major version to be higher than the current version; hence semver incompatibility and the exception thrown. 

A minor version change is not an incompatibility problem since, based on semver, the change would have been made in a backwards-compatible fashion (https://semver.org/).

I don't really expect us to reach version `100.0.0`, but your underlying point about maintainability is a good one 👍 . I've updated the test to increment the major version based on the current current version.

manumonti: oh, got it. So, yes, that's a good approach.

Issue #166, opened 2023-02-01
Alter PRE-adapted tDec API design for CBD-DKG


piotr-roslaniec:
- Scope it from comments in this issue: https://github.com/nucypher/ferveo/issues/53#issuecomment-1412141648

theref:
## Ritual Initiation

We already have Cohorts and Strategies so this will stay unchanged!

```javascript
import { Cohort, Strategy } from '@nucypher/nucypher-ts';

const config = {
  threshold: 3,
  shares: 5,
  porterUri: 'https://porter-tapir.nucypher.community',
};
const newCohort = await Cohort.create(config);
const newStrategy = Strategy.create(newCohort);
```

The background work will be very different though. `Strategy.create(newCohort)` needs to call `initiateRitual` on the Coordinator contract.

This also means that `newStrategy` will be unusable until the ritual has completed successfully. Therefore the `Strategy` class needs to have a `getStatus` function that returns one of:

```solidity
enum RitualState {
    NON_INITIATED,
    AWAITING_TRANSCRIPTS,
    AWAITING_AGGREGATIONS,
    TIMEOUT,
    INVALID,
    FINALIZED
}
```

`Strategy` class will also contain the `ritualID`, this should probably be exposed.

theref:
Another option for Ritual Initiation...

We don't handle it within the strategy because this could become restrictive. Instead we just give detailed examples of how to achieve the functionality with standard tools (ethersjs, hardhat, truffle, etc):

```javascript
import { Cohort, CoordinatorABI, CoordinatorAddress } from '@nucypher/nucypher-ts';

const config = {
  threshold: 3,
  shares: 5,
  porterUri: 'https://porter-tapir.nucypher.community',
};
const newCohort = await Cohort.create(config);
const nodes = newCohort.ursulaAddresses;

import { ethers } from "ethers";
const provider = new ethers.providers.Web3Provider(window.ethereum)
const coordinatorContract = new ethers.Contract(CoordinatorAddress, CoordinatorABI, provider);
```

theref:
Another question... `Cohort` and `Strategy` are really just configuration objects, but which should handle the interface to the DKG ritual? In the comments above I outlined an example where it was contained within `Strategy`, but there's an argument that it should be part of the `Cohort`.

[...]

PR #231, opened 2023-06-28
Refactor CBD API

piotr-roslaniec:
**Type of PR:**
- Refactor

**Required reviews:** 
- 2

**What this does:**
- Replaces `Configuration` type with a raw `porterUri`
- Replaces `defaultConfiguration` with `getPorterUri` based on NuCypher network and not a `chainId`
- Moves `Porter` out of `characters` and renames it into `PorterClient`
- Removes `porterUri` and `web3Provider` from character constructors and moves them into verbs

**Issues fixed/closed:**
- Related to #166
  - https://github.com/nucypher/nucypher-ts/issues/166#issuecomment-1609886074
  - https://github.com/nucypher/nucypher-ts/issues/166#issuecomment-1611343404

**Why it's needed:**
- One of the changes necessary to transition into the new CBD API

**Notes for reviewers:**
- There are breaking changes in the API


src/characters/bob.ts
```diff
@@ -37,11 +35,11 @@ export class RemoteBob {
...
 
 export class Bob {
-  private readonly porter: Porter;
+  private readonly porter: PorterClient;
   private readonly keyring: Keyring;
 
-  constructor(config: Configuration, secretKey: SecretKey) {
-    this.porter = new Porter(config.porterUri);
+  constructor(porterUri: string, secretKey: SecretKey) {
+    this.porter = new PorterClient(porterUri);
```
derekpierre, line 42: I don't think Bob should take a Porter uri either - just like Alice.

src/characters/pre-recipient.ts
```diff
@@ -12,23 +12,22 @@ import { Condition, ConditionContext } from '../conditions';
...
 
-import { Porter } from './porter';
-
-export type PreTDecDecrypterJSON = {
+export type PreDecrypterJSON = {
   porterUri: string;
   policyEncryptingKeyBytes: Uint8Array;
   encryptedTreasureMapBytes: Uint8Array;
   publisherVerifyingKeyBytes: Uint8Array;
   bobSecretKeyBytes: Uint8Array;
```
derekpierre, line 23: If this is no longer PreTDecDecrypter, and just PreDecrypter, then bob's secret keys should no longer be needed, right?

src/config.ts
```diff
@@ -1,29 +1,17 @@
...
+type Network = 'mainnet' | 'tapir' | 'oryx' | 'lynx';
 
-export type Configuration = {
-  readonly porterUri: string;
+const PORTER_URIS: Record<Network, string> = {
+  // TODO: Make sure these are correct
+  mainnet: 'https://porter.nucypher.community',
+  tapir: 'https://porter-tapir.nucypher.community',
+  oryx: 'https://porter-oryx.nucypher.community',
+  lynx: 'https://porter-lynx.nucypher.community',
```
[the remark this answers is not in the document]
derekpierre: Is it possible to move all of the porter logic in this file to the `porter.ts` file? All of the logic is Porter related. 

Then this file is no longer needed. Unless `Network` is used elsewhere.

src/dkg.ts
```diff
@@ -102,7 +102,7 @@ export class DkgClient {
     return {
       id: ritualId,
       dkgPublicKey: DkgPublicKey.fromBytes(dkgPkBytes),
-      threshold: ritualParams.threshold,
+      threshold: ritual.dkgSize, // TODO: dkgSize is not the threshold, but in this case shares == threshold == dkgSize
```
derekpierre, line 105: I'm confused by this line...could you clarify what's going on here? Is this testnet specific code?

src/sdk/strategy/cbd-strategy.ts
```diff
@@ -66,18 +62,26 @@ export class CbdStrategy {
 
 export class DeployedCbdStrategy {
   private constructor(
-    public readonly decrypter: CbdTDecDecrypter,
+    public readonly decrypter: ThresholdDecrypter,
```
derekpierre, line 65: Ignoring for now, but we need some kind of common naming scheme for the Strategy and its encrypter/decrypter.

`DeployedCbdStrategy` having a `ThresholdDecrypter` seems a bit disjointed/disconnected - we need to tie the strategy and encrypter/decrypter together through common naming somehow. I know this is a hot-button topic across the team, so no recommendations from my end for this PR, but something we could collectively discuss and address separately.

src/sdk/strategy/pre-strategy.ts
```diff
@@ -65,26 +62,34 @@ export class PreStrategy {
...
-    const bob = new Bob(configuration, this.bobSecretKey);
+    if (shares > this.cohort.numUrsulas) {
+      throw new Error(
+        `Threshold ${threshold} cannot be greater than the number of Ursulas in the cohort ${this.cohort.numUrsulas}`
+      );
+    }
+
+    const porterUri = this.cohort.porterUri;
+    const alice = Alice.fromSecretKey(this.aliceSecretKey);
+    const bob = new Bob(porterUri, this.bobSecretKey);
```
derekpierre, line 78: Related to a previous comment - only Bob's public key should be needed for PRE, so secret key should not be needed.

src/dkg.ts
piotr-roslaniec: Yes, it's a testnet-specific code. I added clarification.

src/characters/bob.ts
piotr-roslaniec: :heavy_check_mark: 

What do you think about other characters like `Decrypter` and `Encrypter`? They are meant to persisted and transferred so it may make sense to store `porterUri` instead of using it ad-hoc.

src/config.ts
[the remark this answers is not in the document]
piotr-roslaniec: :heavy_check_mark:

src/characters/pre-recipient.ts
piotr-roslaniec: How does the `PreDecrypter` decrypt without a secret key?

src/sdk/strategy/pre-strategy.ts
piotr-roslaniec: I think I'm missing something about this transition that we're going through, i.e. from `PreTDecDecrypter` to just `PreDecrypter`. For example, I thought that Bob needs a secret key to decrypt message kits at some point.

src/characters/pre-recipient.ts
derekpierre: For PRE-tDEC Bob's secret key bytes were publicly known anyway so I thought that's why they were passed in here i.e. the Universal Bob concept. 

However, for just PRE, Bob's secret bytes should continue to be a secret to everyone except for Bob. So a few things stand out:
1. Should a `PREDecrypter` be handling Bob's secret bytes at all - that seems like a security hole. More philosophically, what is a "decrypter" in the context of PRE? For example, does a PRE decrypter return re-encrypted bytes (encrypted for Bob) that Bob can then decrypt himself instead? Basically it gets the data re-encrypted and is separated from decryption. It's kind of like the line between Porter and Bob or in Python the line between `PRERetrievalClient` vs `Bob`. In which case, it's not a "decrypter" per se.

2. If `PREDecrypter`s are persistable they should not be persisting bob's secret bytes. Bob should maintain control of his own secret key, which brings me back to point 1) above.

I might need to go back and look at the existing code, maybe I'm just missing something - I think this is where the API becomes very tricky 😅

src/characters/bob.ts
derekpierre: Seems like not. Much like Bob/Alice they *use* a Porter and therefore should probably take a porterUri as a parameter to functions that perform network actions (eg. `retrieve`). 

wdyt, since you are closer to the code than I am?

piotr-roslaniec: I think it's going to add a little bit of bloat to method signatures, but it may be a good trade-off between ergonomics  and clarity on what consists a protocol object.

src/characters/pre-recipient.ts
piotr-roslaniec: I thing I have more clarity now on where to go from there: It feels to me like `PREDecrypter` is no longer a satisfactory abstraction for decryption, and that job should be delegated to `Bob`. And in that case, maybe we should also revisit the shape and function of `PreStrategy`.

These item look like a big job, and I'd rather sketch a refactoring in a separate issue (#166), discuss it, and then fix it in another PR.

test/unit/cbd-strategy.test.ts
```diff
@@ -163,14 +163,14 @@ describe('CbdTDecDecrypter', () => {
   it('serializes to a plain object', async () => {
```
manumonti, line 163: Should we change the name of this test component? 

```
  describe('CbdTDecDecrypter', () => {
```

to 

```
  describe('TresholdDecrypter', () => {
```

manumonti (commented):
LGTM! 👌

Just a couple of thoughts/questions.

Also, a reminder that we should reflect this changes in documentation. Not sure if this is a good moment or if it is better to wait until the refactoring process ends.

src/porter.ts
```diff
@@ -9,9 +9,27 @@ import {
...
-import { Base64EncodedBytes, ChecksumAddress, HexEncodedBytes } from '../types';
-import { fromBase64, fromHexString, toBase64, toHexString } from '../utils';
+import { ConditionContext } from './conditions';
+import { Base64EncodedBytes, ChecksumAddress, HexEncodedBytes } from './types';
+import { fromBase64, fromHexString, toBase64, toHexString } from './utils';
+
+type Network = 'mainnet' | 'tapir' | 'oryx' | 'lynx';
+
+const PORTER_URIS: Record<Network, string> = {
+  // TODO: Make sure these are correct
```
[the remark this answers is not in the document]
piotr-roslaniec: I think these are correct, going to remove TODO

test/unit/cbd-strategy.test.ts
piotr-roslaniec: :heavy_check_mark:

manumonti (approved):

theref (approved):

PR #232, opened 2023-06-29
Implement local ritual verification

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
- 2

**What this does:**
- Introduces local ritual verification using hard-coded keys

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
- Allows client to verify ritual correctness before performing decryption etc.

**Notes for reviewers:**
- Uses hard-coded keys from ritual on `lynx`. Usage of keys from the coordinator is blocked by https://github.com/nucypher/nucypher-contracts/pull/77
- Tested manually on `lynx`
- Do we want to perform ritual verification at a different moment? Before encryption?
- Do we want to check the ritual state at any moment? Before encryption? We're using  pre-made ritual here but in the future, it would make more sense
- This PR is on hold and blocked by 
  - https://github.com/nucypher/nucypher/issues/3167
  - https://github.com/nucypher/nucypher-ts/issues/234

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [1.0.0-alpha.2](https://github.com/nucypher/nucypher-ts/compare/v1.0.0-alpha.1...v1.0.0-alpha.2) (2023-08-02)
+
+
+### ⚠ BREAKING CHANGES
+
+* porter uris are based on networks, not chain ids
+* replaced configuration with raw porter uri
+
+### Features
+
+* implement local ritual verificaiton ([39241c0](https://github.com/nucypher/nucypher-ts/commit/39241c0cacc2b29bf11a5f4c88a6d53bb8ea4375))
+* porter uris are based on networks, not chain ids ([d911481](https://github.com/nucypher/nucypher-ts/commit/d911481d0f1b8f80e022b2c32de5aaa53607a2f9))
+* replaced configuration with raw porter uri ([fcd7fc0](https://github.com/nucypher/nucypher-ts/commit/fcd7fc0cb505b71a6e709ae37c1c53f9b2261f53))
+
 ## [1.0.0-alpha.1](https://github.com/nucypher/nucypher-ts/compare/v1.0.0-alpha.0...v1.0.0-alpha.1) (2023-07-10)
 
 
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@nucypher/nucypher-ts",
   "author": "Piotr Roslaniec <p.roslaniec@gmail.com>",
-  "version": "1.0.0-alpha.1",
+  "version": "1.0.0-alpha.2",
   "license": "GPL-3.0-only",
   "repository": {
     "type": "git",


src/agents/coordinator.ts
```diff
@@ -24,12 +24,23 @@ export interface CoordinatorRitual {
 export type DkgParticipant = {
   provider: string;
   aggregated: boolean;
+  // TODO: How do I get the transcript from the Coordinator?
+  transcript: Transcript;
```
derekpierre, line 28: I assume this TODO is no longer relevant? Yep, this is how you get the transcript

src/characters/cbd-recipient.ts
```diff
@@ -73,16 +78,36 @@ export class CbdTDecDecrypter {
...
+      web3Provider,
+      this.ritualId
+    );
+    if (ritualState !== DkgRitualState.FINALIZED) {
+      throw new Error(
+        `Ritual with id ${this.ritualId} is not finalized. Ritual state is ${ritualState}.`
+      );
+    }
+
+    const isLocallyVerified = await DkgClient.verifyRitual(
```
derekpierre, line 96: I'm not sure whether this needs to be done every time `retrieve` is called. Perhaps this should be optional i.e. provided as a parameter to `retrieve(...)`. It's unclear to me whether the default value should be `True` of `False` (cc @cygnusv ).

src/dkg.ts
```diff
@@ -106,24 +112,57 @@ export class DkgClient {
...
+      ritualId
+    );
+
+    // TODO: Does this check make sense here? Or do we delegate it to the Coordinator contract?
+    // for (const p of participants) {
+    //   // Not every participant has submitted a transcript
+    //   if (!p.aggregated) {
+    //     return false;
+    //   }
+    // }
```
derekpierre, line 131: Perhaps you should just check the state of the ritual here - the expectation would that the ritual state is finalized. 

I believe that would be sufficient, and negate any specific checks for the participant.

src/dkg.ts
```diff
@@ -106,24 +112,57 @@ export class DkgClient {
...
+        fromHexString(
+          '6000000000000000ace9d7567b26dafc512b2303cfdaa872850c62b100078ddeaabf8408c7308b3a43dfeb88375c21ef63230fb4008ce7e908764463c6765e556f9b03009eb1757d179eaa26bf875332807cc070d62a385ed2e66e09f4f4766451da12779a09036e'
+        )
+      ),
+      '0xb15d5A4e2be34f4bE154A1b08a94Ab920FfD8A41': FerveoPublicKey.fromBytes(
+        fromHexString(
+          '60000000000000008b373fdb6b43e9dca028bd603c2bf90f0e008ec83ff217a8d7bc006b585570e6ab1ce761bad0e21c1aed1363286145f61134ed0ab53f4ebaa05036396c57f6e587f33d49667c1003cd03b71ad651b09dd4791bc631eaef93f1b313bbee7bd63a'
+        )
+      ),
+    };
```
derekpierre, line 160: Can we do this in a test with mocks etc. instead of having hardcoded values in actual code and subsequently going stale?

src/dkg.ts
```diff
@@ -85,15 +90,16 @@ export class DkgRitual {
...
-    threshold: number;
-  }): Promise<DkgRitual> {
+  public static async initializeRitual(
+    web3Provider: ethers.providers.Web3Provider,
+    ritualParams: {
+      shares: number;
+      threshold: number;
+    }
+  ): Promise<DkgRitual> {
     const ritualId = 2;
```
derekpierre, line 101: I assume all of this is only for testing. Is this something that can be moved to tests / mocked in tests instead of including it in actual code?

src/agents/coordinator.ts
piotr-roslaniec: :heavy_check_mark:

src/characters/cbd-recipient.ts
piotr-roslaniec: I think `true` is a sensible default since the security of the user depends.

src/dkg.ts
piotr-roslaniec: Until we fix https://github.com/nucypher/nucypher-contracts/pull/77 we can:
1) Expose this as a paremeter

src/characters/cbd-recipient.ts
derekpierre: Maybe. Technically if the ritual made it to completion, then the validation was crowd-sourced to all of the nodes in the cohort and was fine..., no? This is for the unlikely case that the entire cohort was malicious - since any one non-malicious Ursula would have prevented the DKG ritual from completing.

src/dkg.ts
piotr-roslaniec: My assumption was that for the time being we're going to operate on a set of premade rituals, and right now we only have one

src/characters/cbd-recipient.ts
piotr-roslaniec: We can only know if the participants contributed the correct transcripts once we perform the verification. I don't think there is any other mechanism in place at the moment. For example, how would a non-malicious Ursula prevent the ritual from completing? I don't see this implemented in the `Coordinator` contract.

Notice how without verification it's also possible for malicious Ursulas to shut down any ritual without accountability.

I think in order to address this we need to discuss the scope of configuration sets ("packages") we want to ship and then backtrack how rituals are supposed to work on the client side from there.

src/dkg.ts
piotr-roslaniec: :heavy_check_mark:

src/dkg.ts
piotr-roslaniec: On second thought I think it's reasonable to expose `ritualId` as a parameter at least within the SDK

piotr-roslaniec: Implemented in #233

src/characters/cbd-recipient.ts
piotr-roslaniec: Would moving the ritual verification step to `CbdTDecDecrypter.create` would be a good temporary solution?

derekpierre: > For example, how would a non-malicious Ursula prevent the ritual from completing? I don't see this implemented in the Coordinator contract.

This is done by the honest Ursula providing a correct aggregated transcript that does match the others (from malicious nodes) - https://github.com/nucypher/nucypher-contracts/blob/main/contracts/contracts/coordination/Coordinator.sol#L228. This is why currently the protocol only needs one honest Ursula out of the `n` nodes in the cohort. Subsequently, if all aggregated transcripts match, then ritual state is then set to finalized. Knowing that the state is finalized, means that either all nodes were malicious (probabilistically unlikely for larger-sized cohorts though not 0%) or that everything was good/correct. 

To be clear, local verification is good as an option 👍 , I don't know/am unsure that we want it to be done every time which is the current set-up; should someone be explicit (default parameter is `False` and they have to specifically set it to `True`) about checking it versus being implicit (default to `True` and they have to specifically set it to `False`). In the long term, local verification would not be used at all because the protocol would be even more robust in the future (slashing, proofs, on-chain verification etc.), so having the default be `False` and adopters having to change to `True` could make sense - it's just something we should think about.... what do others think @cygnusv , @KPrasch ?

Side question: how long does the local verification take?

> Would moving the ritual verification step to CbdTDecDecrypter.create would be a good temporary solution?

Interesting. That would seem to be a better spot since the verification would only be done once per encrypter, and is more hidden from the user. That being said, I still have questions about doing local verification in general, but it would definitely be an improvement, and seemingly more hidden from the user. Nice suggestion.

piotr-roslaniec: > This is done by the honest Ursula providing a correct aggregated transcript 

Thanks for explaining that, I just started working on `Coordinator` changes and didn't have a clear picture there.

> Side question: how long does the local verification take?

The complexity ceiling is a single BLS12-381 pairing, so an order of tens of milliseconds.

manumonti (approved):
LGTM! ✌️

src/dkg.ts
derekpierre: My vote would be the pause until https://github.com/nucypher/nucypher-contracts/pull/77 is dealt with.  This PR would then be a downstream change. Note that PR 77 would also have downstream effects on `nucypher`. 

I'd like to avoid hard-coding things that we know will change in actual code; tests are one thing, but in actual code it should be avoided if possible.

piotr-roslaniec: I agree. Let's keep this PR reviewed open and reconsider merging after downstream changes are dealt with. I will document them where appropriate.

piotr-roslaniec:
So the original issue was caused by rebasing branches in the wrong order, i.e. `local-verification` should have been rebased over `alpha`. Instead, I rebased `alpha` over `local-verification`.

PR #233, opened 2023-06-30
Add ritual initialization to client

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
- 2

**What this does:**
- Replaces hardcoded ritual with a proper ritual initialization in `DkgClient.initializeRitual`
- Implements accessory methods, such as waiting for ritual initialization to finish

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
- Currently, users may only use premade rituals and only under certain constraints. See https://github.com/nucypher/nucypher-ts/pull/232
- With this PR, users should be able to initialize rituals on their own using `Cohort` and `Strategy` APIs 

**Notes for reviewers:**
- Based on #232
- Tested on `lynx` with hand-picked `Cohort`. Documented a case where ritual initialization fails when `Cohort` doesn't meet staking requirements here: #235


src/sdk/strategy/cbd-strategy.ts
```diff
@@ -30,15 +30,13 @@ export class CbdStrategy {
   }
 
   public async deploy(
-    web3Provider: ethers.providers.Web3Provider
+    web3Provider: ethers.providers.Web3Provider,
+    blockUntilDeployed = true
```
piotr-roslaniec, line 34: Is this a sensible default?

manumonti: It doesn't sound bad to me, but maybe I'm not the best person to say it.

manumonti (approved):
LGTM! 👏

manumonti (approved):

src/sdk/strategy/cbd-strategy.ts
theref: i like it, you don't want to get in the situation of attempting to use a ritual that never successfully completed

theref (approved):
Looks good!

PR #239, opened 2023-07-07
Fix incorrect `Condition` parsing

piotr-roslaniec:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 2

**What this does:**
- Fixes a bug where conditions would be incorrectly deserialized from their string format, leading to `ConditionContext` attempting to access `undefined` fields 
- Changes how `ConditionContext` is prepared for Porter to fix this issue passing silently in tests

**Why it's needed:**

**Notes for reviewers:**
- Need to back-port this to `alpha` tree

### DIFF
--- a/src/characters/pre-recipient.ts
+++ b/src/characters/pre-recipient.ts
@@ -8,7 +8,7 @@ import {
 } from '@nucypher/nucypher-core';
 import { ethers } from 'ethers';
 
-import { Condition, ConditionContext } from '../conditions';
+import { ConditionContext, ConditionExpression } from '../conditions';
 import { Keyring } from '../keyring';
 import { PolicyMessageKit } from '../kits/message';
 import { RetrievalResult } from '../kits/retrieval';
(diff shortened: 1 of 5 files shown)


src/characters/pre-recipient.ts
```diff
@@ -103,14 +103,11 @@ export class PreTDecDecrypter {
     const conditions = messageKits
       .map((mk) => mk.conditions)
       .filter((condition): condition is Conditions => !!condition)
-      .map((condition) => JSON.parse(condition.toString()))
-      .reduce((acc: Record<string, string>[], val) => acc.concat(val), []);
+      .map((condition) => ConditionExpression.fromJSON(condition.toString()))
+      .reduce((acc: ConditionExpression[], val) => acc.concat(val), [])
+      .map((condExpr: ConditionExpression) => condExpr.condition);
```
piotr-roslaniec, line 108: This out of date condition handling was the root cause

theref (approved):

manumonti (approved):
LGTM! 👏

Thanks for fixing this so quickly.

derekpierre (approved):
🎸

PR #251, opened 2023-07-17
Timeout

### LINKED ISSUE #249: Handle ritual initialization timeout
- When initializing the ritual, handle a case where ritual participants may not contribute in a timely manner
- Check [ritual timeout](https://github.com/nucypher/nucypher-contracts/blob/main/contracts/contracts/coordination/Coordinator.sol#L76-L77) when waiting for `EndRitual` event

theref:
**Type of PR:**
- [ ] Feature

**Required reviews:** 
- [ ] 2

Fixes #249


manumonti (commented):
The code looks good to me (I made a little suggestion).

Thinking aloud, since the "timer" that runs in `nucypher-ts` and the "timer" than runs in `Coordinator contract` are not the same and probably, due to the network delays, one of them will reach the deadline before the other... I wonder if the difference between these two deadlines is significant or not 🤔

src/agents/coordinator.ts
```diff
@@ -80,6 +80,15 @@ export class DkgCoordinatorAgent {
     return Coordinator.rituals(ritualId);
   }
 
+  public static async getTimeout(
+    provider: ethers.providers.Web3Provider
+): Promise<number> {
+    const Coordinator = await this.connectReadOnly(provider);
+    const timeout = await Coordinator.timeout();
```
piotr-roslaniec, line 87: Timeout is the number of seconds counted since the ritual started, right? What happens if my transaction gets stuck (low gas etc.) but I'm still counting from when I send the transaction? Should we instead count the timeout from the "ritual start" event or some other cut-off?

theref: No, `timeout` is a variable on the Coordinator https://github.com/nucypher/nucypher-contracts/blob/da35b9d9f13ddebd7ce7bbbc3fbc7d5fb0d95411/contracts/contracts/coordination/Coordinator.sol#L61

It's set at Coordinator level and applies to all rituals

piotr-roslaniec: I think I'm missing something here. Timeout is defined as [`ritual.initTimestamp + timeout < block.timestamp`](https://github.com/nucypher/nucypher-contracts/blob/da35b9d9f13ddebd7ce7bbbc3fbc7d5fb0d95411/contracts/contracts/coordination/Coordinator.sol#L76-L77), and the initial value of `ritual.initTimestamp` is taken from [`block.timestamp`](https://github.com/nucypher/nucypher-contracts/blob/da35b9d9f13ddebd7ce7bbbc3fbc7d5fb0d95411/contracts/contracts/coordination/Coordinator.sol#L127). So it seems to me like the `timeout` denotes some number of seconds, and the timeout occurs after we reach some number of seconds after the ritual started.

Do you see any edge cases here? If we use `setTimeout`, is it going to match exactly the calculation performed in ritual state checks? I.e. `ritual.initTimestamp + timeout < block.timestamp`

theref: >`timeout` denotes some number of seconds, and the timeout occurs after we reach some number of seconds after the ritual started.
Yes, agree

`setTimeout` is changing the config of Coordinator, so not sure that it helps us.

But i see what your original comment implied now. When we do:
```ts
const ritualId = await DkgCoordinatorAgent.initializeRitual(
      web3Provider,
      ursulas.sort()
    );
```
The fact that we `await` doesn't always mean that the transaction went through? and it could therefore be sitting around in the mempool whilst the `Promise.race` starts counting down. Is that correct?

piotr-roslaniec: `initializeRitual` is blocked until the ritual is started (and some event is emitted, etc.) so it may not be an issue after all.

I was trying to give a concrete example of something that @manumonti already hinted on in his comment:
> Thinking aloud, since the "timer" that runs in nucypher-ts and the "timer" than runs in Coordinator contract are not the same and probably, due to the network delays, one of them will reach the deadline before the other... I wonder if the difference between these two deadlines is significant or not

I didn't make a full analysis on what are the different edge cases we may have here, but one other that comes to mind is where we're awaiting for `timeout` and the `Coordinator` admin changes the value of `timeout`. I think this case and some basket of other cases can be handled by replacing `setTimeout` by `setInterval` and by redoing the contract calculation `ritual.initTimestamp + timeout < block.timestamp` in JS. Or alternatively, querying the contract for the ritual state. The latter seems to be more robust.

src/dkg.ts
```diff
@@ -122,15 +122,25 @@ export class DkgClient {
     );
 
     if (waitUntilEnd) {
-      const isSuccessful = await DkgClient.waitUntilRitualEnd(
-        web3Provider,
-        ritualId
-      );
-      if (!isSuccessful) {
+      const timeout = await DkgCoordinatorAgent.getTimeout(web3Provider);
```
derekpierre, line 125: Haven't thought this through fully just yet, but you may be able to leverage the Coordinator contract to do the bulk of the timeout work for you. See https://github.com/nucypher/nucypher-contracts/blob/main/contracts/contracts/coordination/Coordinator.sol#L75.

The ritual state will return TIMEOUT if the Ritual has not completed within the timeout window. Of course, you don't want to loop hitting the Coordinator contract, so only hitting it periodically could be simpler.

Basically the end case for waiting is you receive `EndRitual` or the returned state from the Coordinator contract is INVALID / TIMEOUT / FINALIZED.

src/dkg.ts
```diff
@@ -122,15 +122,25 @@ export class DkgClient {
...
+      const bufferedTimeout = timeout * 1.1;
```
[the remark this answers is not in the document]
theref: yeah i was trying to avoid potential infinite loops, but i agree that the addition of 10% is too much, it was really there as a placeholder to see how people react

derekpierre:
Repeating my comment from Discord (https://discord.com/channels/866378471868727316/866378471868727319/1131960750836031618) to this PR:

Just so that I get the full picture:
- call `DKGCoordinatorAgent.initializeRitual(...)` which doesn't return until there is at least one block confirmation
- then you basically wait for the `EndRitual` event - this is the ideal path - but of course you want to be robust. There is the case that the ritual times out in which case there is not `EndRitual` event, so you want to timeout waiting for the `EndRitual` event
- If you wait the ritual timeout, you can then call the `Coordinator` contract to determine the status of the ritual - https://github.com/nucypher/nucypher-contracts/blob/main/contracts/contracts/coordination/Coordinator.sol#L75. In all likelihood by this time the Contract would return a final state (`TIMEOUT`, `INVALID`, or `FINALIZED`), but the concern (which is a good one :thumbsup: ) is that you may have some slight discrepancy in the time that the Contract started it's timer and the time that you did.

Assuming this is the correct premise...

You can probably use the ritual's `initTimestamp` value stored in the Ritual object in the contract (https://github.com/nucypher/nucypher-contracts/blob/main/contracts/contracts/coordination/Coordinator.sol#L49) to calculate when you need to wait until - which is basically `timestamp_to_wait_util = initTimestamp + ritual timeout` and then wait until `current_block_time` > `time_to_wait_until` and then call `getRitualState()` and the proper status will be returned. The `getRitualState()` function uses the same calc - https://github.com/nucypher/nucypher-contracts/blob/main/contracts/contracts/coordination/Coordinator.sol#L84C20-L84C35

wdyt?

src/dkg.ts
```diff
@@ -122,15 +122,45 @@ export class DkgClient {
...
+      const initTimestamp = await DkgCoordinatorAgent.getRitualInitTime(
         web3Provider,
         ritualId
       );
-      if (!isSuccessful) {
+      const timeout = await DkgCoordinatorAgent.getTimeout(web3Provider);
+      const endTime = initTimestamp + timeout;
+
+      // Wait until the current time is past the endTime
+      while (Math.floor(Date.now() / 1000) < endTime) {
```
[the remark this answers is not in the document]
theref: will `nowTimestamp` be updated at the end of every loop? i thought it would be calculated once and then remain static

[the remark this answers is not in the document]
piotr-roslaniec: Oh yeah, you're right. I just wanted to suggest naming this variable.

derekpierre (changes requested):
Chatted with James a bit, about some issues we may need to address in this PR before merging:

1. It seems that `onRitualEndEvent(...)` uses `Coordinator.once(...)` which always (?) expects the `EndRitual` event to occur. However in the case of `TIMEOUT` state there is no EndRitual event - timeouts don't produce an `EndRitual` event because  a tx is required to emit an event, and timeouts by nature imply no response, and therefore no tx.

2. Currently the code waits for clock time to expire, then waits for block time to expire(the contract uses block time for timeout), then tries to get the `EndRitual` event. We need to be able to short-circuit this check if the ritual finishes quickly i.e. it's possible/likely (😅 ) that the ritual completes (`FINALIZED` or `INVALID`) before the timeout, but it seems the code still waits the entire timeout which is unnecessary.

piotr-roslaniec:
Ritual initialization will be disabled during 7.0.0-beta. Shall we close this PR?

PR #263, opened 2023-08-11
Implement the TACo API module

### LINKED ISSUE #166: Alter PRE-adapted tDec API design for CBD-DKG

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
- 3

**What this does:**
- Implements a simplified API version for TACo

**Issues fixed/closed:**
- Fixes #166 

**Why it's needed:**

**Notes for reviewers:**
- Rebased over https://github.com/nucypher/nucypher-ts/pull/286
- Please see `test/unit/taco.test.ts` for usage examples

### DIFF
--- a/src/agents/coordinator.ts
+++ b/src/agents/coordinator.ts
@@ -120,8 +120,8 @@ export class DkgCoordinatorAgent {
     const Coordinator = await this.connectReadOnly(provider);
     const dkgPublicKeyBytes = dkgPublicKey.toBytes();
     const pointStruct: BLS12381.G1PointStruct = {
-      word0: dkgPublicKeyBytes.slice(0, 16),
-      word1: dkgPublicKeyBytes.slice(16, 32),
+      word0: dkgPublicKeyBytes.slice(0, 32),
+      word1: dkgPublicKeyBytes.slice(32, 48),
     };
     return await Coordinator.getRitualIdFromPublicKey(pointStruct);
   }
--- a/src/agents/coordinator.ts
+++ b/src/agents/coordinator.ts
@@ -1,5 +1,5 @@
-import { SessionStaticKey } from '@nucypher/nucypher-core';
-import { ethers } from 'ethers';
+import { DkgPublicKey, SessionStaticKey } from '@nucypher/nucypher-core';
+import { BigNumberish, ethers } from 'ethers';
 
 import {
   Coordinator,
@@ -13,12 +13,16 @@ import { DEFAULT_WAIT_N_CONFIRMATIONS, getContract } from './contracts';
 
 export interface CoordinatorRitual {
   initiator: string;
-  dkgSize: number;
   initTimestamp: number;
+  endTimestamp: number;
   totalTranscripts: number;
   totalAggregations: number;
-  publicKey: BLS12381.G1PointStructOutput;
+  authority: string;
+  dkgSize: number;
+  threshold: number;
   aggregationMismatch: boolean;
+  accessController: string;
+  publicKey: BLS12381.G1PointStructOutput;
   aggregatedTranscript: string;
 }
 
@@ -59,10 +63,18 @@ export class DkgCoordinatorAgent {
   public static async initializeRitual(
     provider: ethers.providers.Provider,
     signer: ethers.Signer,
-    providers: ChecksumAddress[]
+    providers: ChecksumAddress[],
+    authority: string,
+    duration: BigNumberish,
+    accessController: string
   ): Promise<number> {
     const Coordinator = await this.connectReadWrite(provider, signer);
-    const tx = await Coordinator.initiateRitual(providers);
+    const tx = await Coordinator.initiateRitual(
+      providers,
+      authority,
+      duration,
+      accessController
+    );
     const txReceipt = await tx.wait(DEFAULT_WAIT_N_CONFIRMATIONS);
     const [ritualStartEvent] = txReceipt.events ?? [];
     if (!ritualStartEvent) {
@@ -95,14 +107,23 @@ export class DkgCoordinatorAgent {
     const Coordinator = await this.connectReadOnly(provider);
     // We leave `initiator` undefined because we don't care who the initiator is
     // We leave `successful` undefined because we don't care if the ritual was successful
-    const eventFilter = Coordinator.filters.EndRitual(
-      ritualId,
-      undefined,
-      undefined
-    );
-    Coordinator.once(eventFilter, (_ritualId, _initiator, successful) => {
+    const eventFilter = Coordinator.filters.EndRitual(ritualId, undefined);
+    Coordinator.once(eventFilter, (_ritualId, successful) => {
       callback(successful);
     });
+  }
+
+  public static async getRitualIdFromPublicKey(
+    provider: ethers.providers.Provider,
+    dkgPublicKey: DkgPublicKey
+  ): Promise<number> {
+    const Coordinator = await this.connectReadOnly(provider);
+    const dkgPublicKeyBytes = dkgPublicKey.toBytes();
+    const pointStruct: BLS12381.G1PointStruct = {
+      word0: dkgPublicKeyBytes.slice(0, 16),
+      word1: dkgPublicKeyBytes.slice(16, 32),
+    };
+    return await Coordinator.getRitualIdFromPublicKey(pointStruct);
   }
 
   private static async connectReadOnly(provider: ethers.providers.Provider) {
--- a/src/characters/cbd-recipient.ts
+++ b/src/characters/cbd-recipient.ts
@@ -14,7 +14,6 @@ import { ethers } from 'ethers';
 
 import { DkgCoordinatorAgent, DkgParticipant } from '../agents/coordinator';
 import { ConditionContext } from '../conditions';
-import { DkgRitual } from '../dkg';
 import { PorterClient } from '../porter';
 import { fromJSON, objectEquals, toJSON } from '../utils';
 
@@ -31,22 +30,22 @@ export class ThresholdDecrypter {
     private readonly threshold: number
   ) {}
 
-  public static create(porterUri: string, dkgRitual: DkgRitual) {
+  public static create(porterUri: string, ritualId: number, threshold: number) {
     return new ThresholdDecrypter(
       new PorterClient(porterUri),
-      dkgRitual.id,
-      dkgRitual.dkgParams.threshold
+      ritualId,
+      threshold
     );
   }
 
   // Retrieve and decrypt ciphertext using provider and condition expression
   public async retrieveAndDecrypt(
-    provider: ethers.providers.Provider,
+    web3Provider: ethers.providers.Provider,
     thresholdMessageKit: ThresholdMessageKit,
     signer?: ethers.Signer
   ): Promise<Uint8Array> {
     const decryptionShares = await this.retrieve(
-      provider,
+      web3Provider,
       thresholdMessageKit,
       signer
     );
--- a/src/sdk/strategy/cbd-strategy.ts
+++ b/src/sdk/strategy/cbd-strategy.ts
@@ -81,7 +81,7 @@ export class DeployedCbdStrategy {
     const decrypter = ThresholdDecrypter.create(
       porterUri,
       dkgRitual.id,
-      dkgRitual.dkgParams.threshold
+      dkgRitual.threshold
     );
     return new DeployedCbdStrategy(decrypter, dkgRitual.dkgPublicKey);
   }
--- a/src/sdk/strategy/cbd-strategy.ts
+++ b/src/sdk/strategy/cbd-strategy.ts
@@ -78,7 +78,11 @@ export class DeployedCbdStrategy {
   ) {}
 
   public static create(dkgRitual: DkgRitual, porterUri: string) {
-    const decrypter = ThresholdDecrypter.create(porterUri, dkgRitual);
+    const decrypter = ThresholdDecrypter.create(
+      porterUri,
+      dkgRitual.id,
+      dkgRitual.dkgParams.threshold
+    );
     return new DeployedCbdStrategy(decrypter, dkgRitual.dkgPublicKey);
   }
 
--- a/src/sdk/strategy/cbd-strategy.ts
+++ b/src/sdk/strategy/cbd-strategy.ts
@@ -44,7 +44,7 @@ export class CbdStrategy {
     //   // Given that we just initialized the ritual, this should never happen
     //   throw new Error('Ritual ID is undefined');
     // }
-    const dkgRitual = await DkgClient.getExistingRitual(provider, ritualId);
+    const dkgRitual = await DkgClient.getRitual(provider, ritualId);
     return DeployedCbdStrategy.create(dkgRitual, this.cohort.porterUri);
   }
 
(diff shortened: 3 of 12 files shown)


src/taco.ts
```diff
@@ -0,0 +1,84 @@
...
+  const decrypter = ThresholdDecrypter.create(
+    porterUri,
+    messageKit.ritualId,
+    messageKit.threshold
+  );
+  const condExpr = ConditionExpression.fromAad(messageKit.aad);
+  // TODO: We need web3Provider to fetch participants from Coordinator to make decryption requests.
+  //  Removing this dependency is tied to release of ThresholdMessageKit
+  //  Blocked by changes to nucypher-core and nucypher:
+  //  https://github.com/nucypher/nucypher/pull/3194
```
piotr-roslaniec, line 72: Even after merging these changes, we'll still need `web3Provider` to get the ACP signature

theref (approved):
This looks great, exactly what adopters want to use

src/taco.ts
```diff
@@ -0,0 +1,84 @@
...
+} from '@nucypher/nucypher-core';
+import { ethers } from 'ethers';
+
+import { ThresholdDecrypter } from './characters/cbd-recipient';
+import { Condition, ConditionExpression } from './conditions';
+import { DkgClient } from './dkg';
+import { getPorterUri } from './porter';
+import { toBytes } from './utils';
+
+export interface TacoMessageKit {
```
derekpierre, line 14: Was this a stand-in until `nucypher-core` 0.12.0 was released? Any reason why `nucypher-core` `ThresholdMessageKit` cannot be used?

src/characters/cbd-recipient.ts
```diff
@@ -35,24 +35,24 @@ export class ThresholdDecrypter {
...
-      dkgRitual.dkgParams.threshold
+      ritualId,
+      threshold
     );
   }
 
   // Retrieve and decrypt ciphertext using provider and condition expression
   public async retrieveAndDecrypt(
-    provider: ethers.providers.Web3Provider,
+    web3Provider: ethers.providers.Web3Provider,
```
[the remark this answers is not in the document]
piotr-roslaniec: Yes, however, #270 was developed after this PR so at the time these changes weren't redundant

src/taco.ts
piotr-roslaniec: Yes, but it seems to be that we still need some extra information not covered by `ThresholdMessageKit`: We need to know the `ritualId` and `threshold` to construct `DecryptionRequests`.

src/taco.ts
```diff
@@ -0,0 +1,82 @@
...
+  thresholdMessageKit: ThresholdMessageKit;
+  conditionExpr: ConditionExpression;
+  // TODO: How do we get rid of these two fields? We need them for decrypting
+  // We ritualId in order to fetch the DKG participants and create DecryptionRequests for them
+  ritualId: number;
+  // We need to know the threshold in order to create DecryptionRequests
+  threshold: number;
```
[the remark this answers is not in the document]
piotr-roslaniec: Do we want to expose `class TacoMessageKit.requireSigner` so that decrypting users can detect when they need to pass signer in?

src/taco.ts
derekpierre: This is covered via latest changes in the Coordinator contract.

There is/will be (once deployed) the ability to get the ritual id from the public key via the Coordinator contract - see https://github.com/nucypher/nucypher-contracts/pull/107/files#diff-044dc40960596bb6ed1e69a1a50bfff628a2ca07775e737eec923b13e857fb43R366 (https://github.com/nucypher/nucypher-contracts/pull/107).

`threshold` is also stored in the Ritual stuct in the Coordinator as part of that change as well; once you know the ritual id, you can get the threshold https://github.com/nucypher/nucypher-contracts/pull/107/files#diff-044dc40960596bb6ed1e69a1a50bfff628a2ca07775e737eec923b13e857fb43R67.

src/taco.ts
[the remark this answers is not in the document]
derekpierre: Given my comment above, I question whether this `TacoMesageKit` object is needed at all.

conditionExpr is in the ACP in the ThresholdMessageKit, and I don't believe ritualId and threshold are needed because they can be obtained from the Coordinator contract along with the ThresholdMessageKit.

My concern lies in returning objects that aren't subject to versioning, and the caller potentially serializing them. Ideally, encrypt would return a `ThresholdMessageKit` just like python.

Of course, if additional information is actually needed in the `ThresholdMessageKit` itself we can explore that.

derekpierre (dismissed):
See comment about `TacoMessageKit`.

src/agents/coordinator.ts
```diff
@@ -57,18 +61,23 @@ export class DkgCoordinatorAgent {
...
-    const tx = await Coordinator.initiateRitual(providers);
-    const txReceipt = await tx.wait(DEFAULT_WAIT_N_CONFIRMATIONS);
-    const [ritualStartEvent] = txReceipt.events ?? [];
-    if (!ritualStartEvent) {
-      throw new Error('Ritual start event not found');
-    }
-    return ritualStartEvent.args?.ritualId;
+    throw new Error('Not implemented');
+
+    // const Coordinator = await this.connectReadWrite(provider, signer);
```
piotr-roslaniec, line 73: I will update it in a minute. It's out of scope for 7.0.0 and irrelevant to this PR so it can wait.

src/agents/coordinator.ts
```diff
@@ -95,16 +107,25 @@ export class DkgCoordinatorAgent {
...
 
+  public static async getRitualIdFromPublicKey(
+    provider: ethers.providers.Provider,
+    dkgPublicKey: DkgPublicKey
+  ): Promise<number> {
+    const Coordinator = await this.connectReadOnly(provider);
+    const dkgPublicKeyBytes = dkgPublicKey.toBytes();
+    const pointStruct: BLS12381.G1PointStruct = {
+      word0: dkgPublicKeyBytes.slice(0, 16),
+      word1: dkgPublicKeyBytes.slice(16, 32),
```
derekpierre, line 124: A G1Point is 48 bytes from the contract is defined as [follows](https://github.com/nucypher/nucypher-contracts/blob/main/contracts/contracts/lib/BLS12381.sol#L14).

Shouldn't `word0` by `(0,32)` and `word1` `(32,48)`?

derekpierre: I think we need to start thinking about "acceptance" tests in `nucypher-ts` that interact with actual contracts (cc @cygnusv , @KPrasch ).

src/sdk/strategy/cbd-strategy.ts
```diff
@@ -78,7 +78,11 @@ export class DeployedCbdStrategy {
   ) {}
 
   public static create(dkgRitual: DkgRitual, porterUri: string) {
-    const decrypter = ThresholdDecrypter.create(porterUri, dkgRitual);
+    const decrypter = ThresholdDecrypter.create(
+      porterUri,
+      dkgRitual.id,
+      dkgRitual.dkgParams.threshold
```
derekpierre, line 84: Is threshold still needed, even though it is now part of the ritual struct?

src/taco.ts
```diff
@@ -0,0 +1,55 @@
...
+  provider: ethers.providers.Provider,
+  message: string,
+  condition: Condition,
+  ritualId: number
+): Promise<ThresholdMessageKit> => {
+  const dkgRitual = await DkgClient.getFinalizedRitual(provider, ritualId);
+  return await encryptLight(message, condition, dkgRitual.dkgPublicKey);
+};
+
+export const encryptLight = async (
```
derekpierre, line 22: Not sure how I feel about `Light` as the term 😅 . Can this just be an overloaded function with the same name?

test/unit/conditions/context.test.ts
```diff
@@ -90,20 +90,16 @@ describe('context parameters', () => {
     };
     const condition = new ContractCondition(conditionObj);
     const conditionExpr = new ConditionExpression(condition);
-    const conditionContext = conditionExpr.buildContext(provider, {}, signer);
     expect(conditionExpr.contextRequiresSigner()).toBe(true);
```
derekpierre, line 93: I think it's good to test the building works i.e.
```suggestion
    expect(conditionExpr.contextRequiresSigner()).toBe(true);

    // check that conditionContext created successfully
    expect(conditionExpr.buildContext(provider, {}, signer)...to succeed
    
    // try to create condition context without signer
    expect(conditionExpr.buildContext(provider, {})....to fail
```

test/unit/conditions/context.test.ts
```diff
@@ -90,20 +90,16 @@ describe('context parameters', () => {
...
   });
 
   it('detects if a signer is not required', () => {
     const condition = new RpcCondition(testRpcConditionObj);
     const conditionExpr = new ConditionExpression(condition);
-    const conditionContext = conditionExpr.buildContext(provider, {}, signer);
     expect(JSON.stringify(condition.toObj()).includes(USER_ADDRESS_PARAM)).toBe(
       false
     );
     expect(conditionExpr.contextRequiresSigner()).toBe(false);
```
derekpierre, line 102: Similar comment as above, for completeness:
```suggestion
    expect(conditionExpr.contextRequiresSigner()).toBe(false);

    // check that conditionContext created successfully even if signer passed
    expect(conditionExpr.buildContext(provider, {}, signer)...to succeed
    
    // try to create condition context without signer
    expect(conditionExpr.buildContext(provider, {})....to succeed
```

derekpierre (commented):
Needs another rebase over #286.

src/agents/coordinator.ts
```diff
@@ -13,12 +13,16 @@ import { DEFAULT_WAIT_N_CONFIRMATIONS, getContract } from './contracts';
...
   totalAggregations: number;
-  publicKey: BLS12381.G1PointStructOutput;
+  authority: string;
+  dkgSize: number;
+  threshold: number;
   aggregationMismatch: boolean;
+  accessController: string;
+  publicKey: BLS12381.G1PointStructOutput;
   aggregatedTranscript: string;
 }
```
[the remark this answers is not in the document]
piotr-roslaniec: The contents of this struct are taken from types generated from contracts by `typechain`. I can trim them down to a size.

src/taco.ts
piotr-roslaniec: There is no support function overloading in JS/TS. I didn't have better ideas for an appropriate name, maybe we could use the `blockchainy` naming from the original TACo API issue.

src/agents/coordinator.ts
piotr-roslaniec: IMHO acceptance tests with contracts are unnecessary. We're only supporting read operations (at this point), and setting those up is not different than existing mocks. Yes, this is a bug that went unnoticed. But it would have been caught easily by QA which we do on every release. 

I think the better solution to this particular problem is to enhance `BLS12381.G1PointStructOutput` generated type with a wrapper that performs validation. The `word0` and `word` fields are represented as strings by `typechain` in generated typings, which also prevented TS compiler from catching this issue.

derekpierre: > But it would have been caught easily by QA which we do on every release.

The goal of the "acceptance" tests is to catch these things before live QA. It is inefficient to catch these kind of bugs during live QA on a testnet. Doesn't mean we will catch everything in those tests, but it does reduce the need for subsequent fixing/releases after bugs are found during live testing.

I don't know what specifically the "acceptance" tests would entail and what it would require from the codebase to be set up, but having everything tested with mocks limits the breadth of testing and bugs that we can find in tests - *who's testing the mocks* is essentially the argument.

"Acceptance" tests is a broad term, and maybe there is more appropriate term here, but we should investigate how we can perform tests closer to what the actual tesntet/production environment would be.

piotr-roslaniec: I understand the advantages of acceptance testing, I just don't think they are necessary at this point. We should probably get them before the release, but not until we're done with monorepo or adapting `viem`: https://github.com/nucypher/nucypher-ts/pull/271

That being said, I'd be happy to accept a PR that implements acceptance testing if anyone wants to chip in.

derekpierre: Definitely not something needed for this PR or this instant - but something we should nonetheless investigate. Filed #295 .

src/sdk/strategy/cbd-strategy.ts
piotr-roslaniec: The client needs to know how many shares to fetch from the nodes (`/cbd_decrypt` responses)

derekpierre: Yes, but now if you have the ritual id, you can get the value of the `threshold` from the Coordinator agent when doing the retrieve, or am I misunderstanding.

manumonti (approved):
LGTM. Great work! 👏

test/unit/conditions/context.test.ts
```diff
@@ -91,6 +91,10 @@ describe('context parameters', () => {
     const condition = new ContractCondition(conditionObj);
     const conditionExpr = new ConditionExpression(condition);
     expect(conditionExpr.contextRequiresSigner()).toBe(true);
+    expect(conditionExpr.buildContext(provider, {}, signer)).toBeDefined();
+    expect(() => conditionExpr.buildContext(provider, {})).toThrow(
+      `Cannot use ${USER_ADDRESS_PARAM} as a parameter without a signer`
```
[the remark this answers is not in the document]
piotr-roslaniec: Yes, I prefer this one

test/unit/conditions/context.test.ts
```diff
@@ -114,7 +120,7 @@ describe('context parameters', () => {
     const conditionExpr = new ConditionExpression(condition);
     expect(conditionExpr.contextRequiresSigner()).toBe(true);
     expect(() => conditionExpr.buildContext(provider, {}, undefined)).toThrow(
-      `Condition contains ${USER_ADDRESS_PARAM} context variable and requires a signer to populate`
+      `Cannot use ${USER_ADDRESS_PARAM} as a parameter without a signer`
```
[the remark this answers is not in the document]
piotr-roslaniec: Yes, I did that on purpose - I remembered updating this line previously and decided to pick this wording.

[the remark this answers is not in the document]
derekpierre: How about?

```
Signer required to satisfy ${USER_ADDRESS_PARAM} context variable in condition.
```

Two thoughts here:
1. It's already a done deal - the condition is already present so it's not that the condition can't use `:userAddress`, it's that the context can't be built.
2. It may not be a "parameter" in the context of a function call, it could be part of the return value test. In general, our terminology here has been "context variables".

src/sdk/strategy/cbd-strategy.ts
piotr-roslaniec: If you inspect the usage of `DeployedCbdStrategy.create` you can see that this is exactly what happens:
- https://github.com/nucypher/nucypher-ts/blob/taco-api/src/sdk/strategy/cbd-strategy.ts#L47-L48
- https://github.com/nucypher/nucypher-ts/blob/taco-api/src/sdk/strategy/cbd-strategy.ts#L95-L96

derekpierre: Right 👍 . I mixed up my layers.

Side question: any reason we need the `DkgRitualParameters` object? This line could look nice as:
```typescript
    const decrypter = ThresholdDecrypter.create(
      porterUri,
      dkgRitual.id,
      dkgRitual.threshold
```
? Anyways, just something to think about.

derekpierre (approved):
🎸 - nice work

src/taco.ts
manumonti: I don't have a good name for it, but I would prefer something like 

```typescript
export const encryptWithDKG = async (
``` 

than the original one

PR #267, opened 2023-08-21
Update Conditions API

### LINKED ISSUE #94: Since its a ts project, consider using Zod instead of joi?
I noticed that you're using joi in condition assertions.  Its much more fully developed ground up for typescript.  @piotr-roslaniec

### LINKED ISSUE #230: Consider narrowing the strings allowed for condition records in ConditionExpressionJSON
See conversation from PR review here - https://github.com/nucypher/nucypher-ts/pull/229#discussion_r1235352880.

Maybe we can use the schema fields defined for underlying conditions to limit the record keys that can be used with `ConditionExpressionJSON` object.

### LINKED ISSUE #258: Enhancements to Conditions API
Currently working with conditions is very difficult. There is no way to get type inference on what the schema of the condition should be based on the condition type (evm, timelock, etc). Here are my suggestions to improve this.

1. Add `conditionType` discriminating field to the condition schema itself

Right now it is very hard to tell from a JSON condition, what type it is. Adding a new top level field to the `Condition` schema itself would make this much easier. Not sure exactly who owns the condition schema definition right now, as if I pass extra fields like this to Porter I get an error. Example would be

```JSON
{
  "conditionType": "evm" | "timelock" | ...,
  "method": "...",
  "parameters": [ ],
  ...
}
```

This also helps immensely on the ts side as you can use a [discriminated union](https://www.fullstory.com/blog/discriminated-unions-and-exhaustiveness-checking-in-typescript/) to parse from JSON

2. Add types/interfaces to the constructor of each type of Condition

Right now the constructors of all types of conditions share the same interface as the base condition which is `value: Record<string, unknown>`. This provides very little feedback to the developer about what they are actually supposed to pass here and leads to a lot of trial/error and frustration. 

To fix this, the actual type of the schema should be inferred in the constructor. For example the type for RpcCondition should look something like:

```typescript

interface RpcConditionProps {
  method: 'eth_getBalance' | 'balanceOf'
  parameters: [':userAddress']
  chain: number
  returnValueTest: ReturnValueTestProps
}

class RpcCondition extends Condition {
  constructor(value: RpcConditionProps) {
    super(value)
  }
}
```

These types can be written manually to match the Joi schema, or generated using something like (Joi to Typescript)[https://github.com/mrjono1/joi-to-typescript]

My real recommendation is to switch from Joi to Zod so the types can be inferred directly from the schema itself. You can see an example [here](https://github.com/ghardin1314/tac-poc/blob/43dabdd9c26a558e903eafe507d81e4b2300fe36/libs/cbd-ts/src/lib/conditions.ts#L37C1-L44C63)

I think these 2 changes will massively improve DX for working with conditions. More than happy to submit a PR for refactoring, but @piotr-roslaniec mentioned there may be big changes to the condition schema so would like to wait until that is finalized

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
- 3 

**What this does:**
- Replaces Joi with Zod, for condition validation
- Adds `conditionType` discriminating field to conditions
- Exposes conditions interface, `*Props` 

**Issues fixed/closed:**
- Fixes #258
- Fixes #230
- Fixes #94

**Why it's needed:**
- Improves developer experience for TS users
- Improves type-safety

**Notes for reviewers:**
- Downstream change addressed in https://github.com/nucypher/nucypher/pull/3201
- Marking this as a breaking change because of breaking changes in conditions API
- Slightly changed the module structure - `Condition` is now at `.conditions.Condition` instead of `.conditions.base.Condition`
- See how condition constructors are beings used now in `conditions/base/*.test.ts`
- See how the type-hints work now in the screenshot below

image


src/conditions/condition-expr.ts
```diff
@@ -28,13 +32,28 @@ export class ConditionExpression {
...
+      case 'time':
+        return new TimeCondition(obj as TimeConditionProps);
+      case 'contract':
+        return new ContractCondition(obj as ContractConditionProps);
+      case 'compound':
+        return new CompoundCondition(obj as CompoundConditionProps);
+      default:
+        throw new Error(`Invalid conditionType: ${obj.conditionType}`);
+    }
+  }
```
piotr-roslaniec, line 55: RFC: Do we really need access to concrete types now that we have the `conditionType` in every `Condition`? For example, what is there to gain for our users from having an `RpcCondition` instead of a `Condition` instance?

piotr-roslaniec:
Hi @ghardin1314, you may be interested in reviewing this PR.

piotr-roslaniec:
Wishing I used [valibot](https://github.com/fabian-hiller/valibot) instead. Perhaps in a different life.

src/conditions/base/contract.ts
```diff
@@ -1,87 +1,90 @@
...
 
 import { ETH_ADDRESS_REGEXP } from '../const';
 
-import { RpcCondition, rpcConditionRecord } from './rpc';
+import { rpcConditionSchema } from './rpc';
 
-export const STANDARD_CONTRACT_TYPES = ['ERC20', 'ERC721'];
+// TODO: Consider replacing with `z.unknown`:
+//    Since Solidity types are tied to Solidity version, we may not be able to accurately represent them in Zod.
+//    Alternatively, find a TS Solidity type lib.
```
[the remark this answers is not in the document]
ghardin1314: Potentially use this?

https://abitype.dev/api/zod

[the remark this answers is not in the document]
ghardin1314: Potentially use this?

https://abitype.dev/api/zod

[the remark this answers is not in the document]
piotr-roslaniec: Yes, it's likely. I've already started looking into `viem`: https://github.com/nucypher/nucypher-ts/pull/271

manumonti (approved):
This is a great job 👏. In really like this approach.

src/conditions/condition-expr.ts
theref: I'm also +1 for having this simplified

src/conditions/base/contract.ts
```diff
@@ -1,87 +1,90 @@
...
-    const { message } = e as Error;
-    return helper.message({
-      custom: message,
-    });
-  }
+const functionAbiVariableSchema = z
+  .object({
+    name: z.string(),
+    type: z.enum(EthBaseTypes),
+    internalType: z.enum(EthBaseTypes), // TODO: Do we need to validate this?
```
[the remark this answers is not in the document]
theref: yeah, it seems likely that `ethers` or `viem` could produce a useful error for the user if the abi isn't valid.

theref (approved):

src/conditions/condition-expr.ts
piotr-roslaniec: So just to reiterate: "I, as a TS developer, don't need to know the concrete type of my `condition` because I already know it's `conditionType`". Does that make sense?

src/conditions/base/time.ts
```diff
@@ -1,23 +1,15 @@
-import Joi from 'joi';
+import { z } from 'zod';
 
-import { omit } from '../../utils';
+import { rpcConditionSchema } from './rpc';
 
-import { RpcCondition, rpcConditionRecord } from './rpc';
+// TimeCondition is an RpcCondition with the method set to 'blocktime' and no parameters
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const { parameters: _, ...restShape } = rpcConditionSchema.shape;
```
derekpierre, line 7: Interesting; so this is how you remove parameters from being a schema entry using zod.

src/conditions/compound-condition.ts
```diff
@@ -1,31 +1,24 @@
...
-      rpcConditionSchema,
-      timeConditionSchema,
-      contractConditionSchema,
-      Joi.link('#compoundCondition')
+export const compoundConditionSchema: z.ZodSchema = z.object({
+  conditionType: z.literal('compound').default('compound'),
+  operator: z.enum(['and', 'or']),
+  operands: z
+    .array(
+      z.lazy(() =>
```
derekpierre, line 12: Cool!

src/conditions/condition-expr.ts
derekpierre: > So just to reiterate: "I, as a TS developer, don't need to know the concrete type of my condition because I already know it's conditionType". Does that make sense?

Not sure I follow. What would be the outcome of such a plan?

test/unit/conditions/base/condition.test.ts
```diff
@@ -11,37 +11,45 @@ import {
...
+    const result = condition.validate(invalidOverride);
+    expect(result.error).toBeDefined();
+    expect(result.data).toBeUndefined();
+    expect(result.error?.format()).toMatchObject({
+      chain: {
+        _errors: [
+          'Invalid literal value, expected 137',
+          'Invalid literal value, expected 80001',
+          'Invalid literal value, expected 5',
+          'Invalid literal value, expected 1',
```
derekpierre, line 49: All of these are included in the error? I guess zod does do something like an array of literals so that the error would be like `"Invalid literal value, not one of 137, 80001, 5, 1"`

test/unit/conditions/condition-expr.test.ts
```diff
@@ -211,50 +212,6 @@ describe('condition set', () => {
       }
     );
 
-    it.each([
```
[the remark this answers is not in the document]
derekpierre: Perhaps you can repurpose to test to check what happens when the `condition_type` is either not defined, or defined but the JSON does not match what's expected based on the type.

src/conditions/condition.ts
```diff
@@ -0,0 +1,63 @@
...
+  } {
+    const newValue = {
+      ...this.value,
+      ...override,
+    };
+    const result = this.schema.safeParse(newValue);
+    if (result.success) {
+      return { data: result.data };
+    }
+    return { error: result.error };
```
derekpierre, line 42: Not sure I follow why we need to return ConditionProps or Error here.


Is the returned data dictionary used somehow, didn't see how in the tests.

Are we relying on users of the API to actively call validate? What I'm wondering is how come we don't just validate in the constructor and raise an error if validation fails. `validate` would essentially be an internal function and not expected to be called externally at all. Or am I missing part of the picture?

test/unit/conditions/base/condition.test.ts
piotr-roslaniec: I'm afraid this is what Zod does when you use `z.literal`. I've been looking for a way around that without much success, but I can give it another try.

derekpierre: In my original comment, I meant to say *"I guess zod does [not] do something..."*

If you've already looked and it's not obvious then no worries. Thanks!

src/conditions/condition-expr.ts
piotr-roslaniec: Let me rephrase that: 
1) I just deserialized a condition. I want to know what type of condition it is. I can just check (by hand or programmatically) what is the value of `conditionType`. That's all good, the problem is solved.
2) But now, why would I need to know the concrete TypeScript type of my conditions? And if I ever need to know that, I follow steps in 1) and cast my condition to the desired type, `condition as RpcCondition` (or any other condition, based on `conditionType`). So, essentially what the code above is doing.

src/conditions/base/rpc.ts
```diff
@@ -1,39 +1,30 @@
...
+  index: z.number().optional(),
+  comparator: z.enum(['==', '>', '<', '>=', '<=', '!=']),
+  value: z.union([z.string(), z.number(), z.boolean()]),
+});
 
-const rpcMethodSchemas: Record<string, Joi.Schema> = {
-  eth_getBalance: Joi.array().items(ethAddressOrUserAddressSchema).required(),
-  balanceOf: Joi.array().items(ethAddressOrUserAddressSchema).required(),
-};
+export type ReturnValueTestProps = z.infer<typeof returnValueTestSchema>;
```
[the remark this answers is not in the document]
derekpierre: 👍 - something common would be great

src/conditions/base/rpc.ts
```diff
@@ -1,39 +1,30 @@
...
-  method: Joi.string()
-    .valid(...Object.keys(rpcMethodSchemas))
-    .required(),
-  parameters: makeParameters(),
-  returnValueTest: returnValueTestSchema.required(),
-};
+export const rpcConditionSchema = z.object({
+  conditionType: z.literal('rpc').default('rpc'),
+  chain: z.union([
+    z.literal(137),
```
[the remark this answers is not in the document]
piotr-roslaniec: There was a small issue with typings here (TypeScript couldn't infer a correct type for a Zod object), but it's fixed now

src/conditions/condition.ts
piotr-roslaniec: I see the point of a "throwing constructor", but I'd like for the user to have a public `validate` method that doesn't throw and return the errors in an orderly manner. That would allow for programmatic feedback on the condition structure. For example, imagine someone building a visual condition builder - It would make sense for them to parse these errors and return something actionable to the caller (end-user).

Perhaps we could settle on a "throwing constructor" and have `validate` as a public static method to call in case someone wants to pre-emptively validate their condition object?

src/conditions/condition-expr.ts
piotr-roslaniec: [ ] Move to `Condition.fromObj`

src/types.ts
```diff
@@ -8,3 +8,10 @@ export enum ChainId {
   GOERLI = 5,
   MAINNET = 1,
 }
+
+export const SUPPORTED_CHAIN_IDS = [
```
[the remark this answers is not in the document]
piotr-roslaniec: Moved `SUPORTED_CHAIN_IDS` to `const.ts` and `ChainId` to `web3.ts`

src/conditions/base/index.ts
```diff
@@ -1,4 +1,43 @@
...
+  }
+}
+
+export class TimeCondition extends Condition {
+  constructor(value: TimeConditionProps) {
+    super(timeConditionSchema, value);
+  }
+}
+
+export {
```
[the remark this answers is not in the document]
piotr-roslaniec: I think props yes (because they make typing a bit easier), schemas - not so much (no direct value add for users).

src/conditions/condition.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (error) {
+      throw new Error(`Invalid condition: ${JSON.stringify(error.issues)}`);
+    }
+    this.value = data;
+  }
+
+  public static validate(
+    schema: ConditionSchema,
+    value: ConditionProps,
+    override: Partial<ConditionProps> = {}
```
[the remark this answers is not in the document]
piotr-roslaniec: Hmmm I don't think we need those anymore

src/conditions/base/index.ts
```diff
@@ -1,4 +1,40 @@
...
+
+export class TimeCondition extends Condition {
+  constructor(value: TimeConditionProps) {
+    super(timeConditionSchema, value);
+  }
+}
+
+export { type ContractConditionProps } from './contract';
+export { type RpcConditionProps } from './rpc';
+export { type TimeConditionProps } from './time';
```
derekpierre, line 40: Should we also export CompoundConditionProps?

piotr-roslaniec: It's already exported from here: https://github.com/nucypher/nucypher-ts/pull/267/files#diff-c69f0d621a87085de1005a17d7157e7bf6fe13316cf6003268fff6af266fee11R11

test/unit/conditions/compound-condition.test.ts
```diff
@@ -1,85 +1,121 @@
...
 
-    expect(orCondition.operator).toEqual('or');
-    expect(orCondition.operands).toEqual([
-      ownsBufficornNFT,
-      testTimeConditionObj,
-    ]);
+    expect(result.error).toBeUndefined();
+    expect(result.data).toEqual({
+      ...conditionObj,
+      conditionType: 'compound',
```
arjunhassard, line 21: For the documentation and Conditions Schema, should we consider `compound` to be a fourth condition type, or on a different 'level' from `contract`, `rpc` and `time`? And does the constructor tell the developer that they're creating a compound condition set?

arjunhassard (approved):
Nice work @piotr-roslaniec 
Thank you @ghardin1314 for the DX improvements

test/unit/conditions/compound-condition.test.ts
piotr-roslaniec: I think `CompoundCondition` is somewhat different since (by definition) it can't be used on its own. That's why I didn't put it into `conditions/base` which is supposed to contain some basic building blocks for conditions.

Yes, the constructor expects `CompoundConditionProps` and provides type hints for TS developers.

PR #270, opened 2023-08-23
Seperate `signer` from `provider` in user-facing API

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
- 2

**What this does:**
- Separates signer and provider abstractions in user-facing APIs

**Issues fixed/closed:**
- Fixes https://github.com/nucypher/nucypher-ts/issues/268

**Why it's needed:**
> E.g., if this PR addresses a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
- Please [see tests](https://github.com/nucypher/nucypher-ts/pull/270/files#diff-e9da6655845be3cade2499964c990f9fc9bf47768165f5cb1d0b361d5c74145fR89-R91) for the relevant changes in the user-facing API
- Now, a lot of methods will require two parameters instead of one (`signer`, `provider`). This is a mild inconvenience, and I could argue that we could simply take one of these instead and then [check for the connection with the other](https://docs.ethers.org/v5/api/signer/#Signer-connect). But in that case, we would have to throw an error when the user mishandles the API. We could also introduce our own type, `SignerWithProvider`, and use it where appropriate. IMHO these considerations are best left for `taco` API, and not for this PR in particular.
- This change will be superseded by changes in https://github.com/nucypher/nucypher-ts/pull/271 in the future
- [ ] Rebase once this PR is merged: https://github.com/nucypher/nucypher-ts/pull/286


manumonti (approved):
Code looks good to me. 👍

About the convenience of separating `Web3Provider` into `Provider` and `Signer`, I would agree. For instance, recent versions of ethers have introduced this strong division (see "Coming from Web3.js" notice in https://docs.ethers.org/v5/api/providers/provider/#Provider).

IMHO this stablish a clear line between methods that requires a provider with signer (like MetaMask does) and others that can use just a provider (like just using Infura or Alchemy endpoint). So, the developer have a cleaner and less-confussing vision about what's going on under TACo/CBD.

I think that with some good examples in documentation we can easily let clear for adopters how to get a signer from a provider (as Piotr mentioned).

Although maybe this is not the best place to discuss it, I would prefer not to choose the option of creating a `SignerWithProvider` object, since IHHO this add some complexity to the code, that can result in confusion. The terms `Provider` and `Signer` are pretty common in web3 development and I think most of the developers are familiar with them.

jdbertron:
Thanks for the explanation.

theref (approved):

src/conditions/condition-expr.ts
```diff
@@ -85,9 +85,10 @@ export class ConditionExpression {
   }
 
   public buildContext(
-    provider: ethers.providers.Web3Provider
+    provider: ethers.providers.Provider,
+    signer: ethers.Signer
```
derekpierre, line 89: Does `buildContext` get called if there are no context variables in the condition, eg. blocktime condition?

What if there are only user-defined context variables since they don't need anything signed?

Basically, I'm wondering if we properly handle signer being undefined if no signing is needed for the context...either because the condition does need it (in which case buildContext may not even be called), or if there are only user-defined custom context variables in which case there is nothing to be signed.

src/agents/subscription-manager.ts
```diff
@@ -66,14 +67,15 @@ export class PreSubscriptionManagerAgent {
...
+    signer: ethers.Signer
   ) {
-    return await this.connect(web3Provider, web3Provider.getSigner());
+    return await this.connect(provider, signer);
   }
 
   private static async connect(
     provider: ethers.providers.Provider,
-    signer?: ethers.providers.JsonRpcSigner
+    signer?: ethers.Signer
```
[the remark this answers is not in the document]
piotr-roslaniec: These are equivalent, I will update them for consistency.

src/conditions/condition-expr.ts
piotr-roslaniec: We could check `ConditionExpression` for the presence of context parameters and avoid passing `signer` in if it doesn't contain one. But we still need to make `signer` available in the user-facing API in case we need it (https://github.com/nucypher/nucypher-ts/blob/sep-signer-provider/src/characters/cbd-recipient.ts#L70)

I was considering this and other cases where the signer is optional. Which one of these approaches do we prefer?
1. Make the signer optional and throw an error if it turns out we actually need it
2. Split API into two user flows and make it required in one of them and remove it from the other.
I'm leaning towards the first option. I'm assuming that API users can detect/handle these cases correctly.
If we'll agree on 1), I will simply add optionality and relevant errors in this PR.

piotr-roslaniec: I tentatively implemented these changes.

derekpierre: I think 1) is reasonable. 

> I'm assuming that API users can detect/handle these cases correctly.

Is there a public API/utility function that can detect whether a signer is needed based on the condition?

piotr-roslaniec: I think it's the `ThresholdDecrypter.retrieveAndDecrypt`, which will be exposed by `taco.decrypt`

derekpierre: I guess I'm wondering is if I'm an API user, 
- if a signer is needed and one is not provided I get prompted somehow 
OR 
- do I need to inspect the condition myself to determine if `:userAddress` is in it and therefore I know I need to provide a signer
OR
- is there a function on a condition object, or the decrypter, or some other utility that provides me with a bool result as to whether I need the signer or not based on the condition
OR
...

TL;DR 😅  If signer is optional, how does the API user determine whether to pass it or not?

piotr-roslaniec: > if a signer is needed and one is not provided I get prompted somehow

There are only two ways that the user can learn that:
- From a function signature (it's "marked" as required, throws an error if missing during TS compilation etc.)
- From run-time errors - The parameter is marked as optional, but there is some dependency on it being present under certain conditions (context parameter is present or not). And because of this hidden dependency, we have to throw an exception in runtime instead

> do I need to inspect the condition myself to determine if :userAddress is in it and therefore I know I need to provide a signer

If you want to avoid having errors thrown in runtime, then yes. 

> is there a function on a condition object, or the decrypter, or some other utility that provides me with a bool result as to whether I need the signer or not based on the condition

No, but we can add it.

derekpierre: If `signer` is optional then I think providing such a facility would be good. cc @KPrasch , @cygnusv in case they have a different view.

piotr-roslaniec: I've added a `ConditionExpression.contextRequiresSigner` helper function and related utilities. This is something end-users may use to check if they need to use a signer for their conditions (to build the context). We should document in an advanced usage guide.

test/unit/conditions/context.test.ts
```diff
@@ -77,6 +78,32 @@ describe('context parameters', () => {
...
+
+  it('detects if a signer is not required', () => {
+    const condition = new RpcCondition(testRpcConditionObj);
+    const conditionExpr = new ConditionExpression(condition);
+    const conditionContext = conditionExpr.buildContext(provider, {}, signer);
+    expect(JSON.stringify(condition.toObj()).includes(USER_ADDRESS_PARAM)).toBe(
+      false
+    );
+    expect(conditionExpr.contextRequiresSigner()).toBe(false);
+    expect(conditionContext.requiresSigner()).toBe(false);
```
derekpierre, line 104: How come we need the check on both the `ConditionExpr` and the `Context` object? I would have assumed it's only needed in `ConditionExpr`.

piotr-roslaniec: We don't necessarily need to check it - it's just when a wrapper for exposing this functionality is higher up in the API. I think it makes sense to add this check to the tests to document this API.

derekpierre: I guess I really meant "expose" and not "check"

What's the case where the call `requiresSigner()` is performed on the `Context` instead of the `Condition`? Having it in the `Context` doesn't mesh with my understanding - the `Context` is just the dictionary of values, but it is the condition that determines whether a signer is needed or not.

piotr-roslaniec: I think I had some use case in mind but on second thought it's already covered by `ContextCondition.validate`. I think we can make `ContextCondition.requiresSigner` private then.

derekpierre: If still needed internally for validation - then yes, private would be better 👍 .

PR #273, opened 2023-08-24
Update to `nucypher-core@0.12.0`

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
- 1

**What this does:**
- Bumps `nucypher-core` version
- Tests changes in https://github.com/nucypher/nucypher-core/pull/74

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
> What should reviewers focus on? 
> Is there a particular commit/function/section of your PR that requires more attention from reviewers?

PR #286, opened 2023-08-30
Apply PR suggestions for #273

### LINKED ISSUE #269: Incorporate latest ACP/ThresholdMessageKit objects from `nucypher-core`
Related to changes from:
- https://github.com/nucypher/nucypher-core/pull/74
- https://github.com/nucypher/nucypher/pull/3194

Relevant areas:
- encryption now returns `ThresholdMessageKit`
- decryption takes `ThresholdMessageKit`
- `ThresholdDecryptionRequest` now uses `CiphertextHeader` and `AccessControlPolicy` objects

piotr-roslaniec:
**Type of PR:**
- Other

**Required reviews:** 
- 2

**What this does:**
- Fixes PR comments from #273 
- Closes #269 
- Closes https://github.com/nucypher/nucypher-ts/issues/293

### DIFF
--- a/src/characters/cbd-recipient.ts
+++ b/src/characters/cbd-recipient.ts
@@ -64,9 +64,9 @@ export class ThresholdDecrypter {
       provider,
       this.ritualId
     );
-    const wasmContext = await ConditionContext.fromAccessControlPolicy(
+    const wasmContext = await ConditionContext.fromConditions(
       provider,
-      thresholdMessageKit.acp,
+      thresholdMessageKit.acp.conditions,
       signer
     ).toWASMContext();
     const { sharedSecrets, encryptedRequests } =
--- a/src/characters/cbd-recipient.ts
+++ b/src/characters/cbd-recipient.ts
@@ -13,7 +13,7 @@ import {
 import { ethers } from 'ethers';
 
 import { DkgCoordinatorAgent, DkgParticipant } from '../agents/coordinator';
-import { ConditionExpression } from '../conditions';
+import { ConditionContext } from '../conditions';
 import { DkgRitual } from '../dkg';
 import { PorterClient } from '../porter';
 import { fromJSON, objectEquals, toJSON } from '../utils';
@@ -42,13 +42,11 @@ export class ThresholdDecrypter {
   // Retrieve and decrypt ciphertext using provider and condition expression
   public async retrieveAndDecrypt(
     provider: ethers.providers.Provider,
-    conditionExpr: ConditionExpression,
     thresholdMessageKit: ThresholdMessageKit,
     signer?: ethers.Signer
   ): Promise<Uint8Array> {
     const decryptionShares = await this.retrieve(
       provider,
-      conditionExpr,
       thresholdMessageKit,
       signer
     );
@@ -59,7 +57,6 @@ export class ThresholdDecrypter {
   // Retrieve decryption shares
   public async retrieve(
     provider: ethers.providers.Provider,
-    conditionExpr: ConditionExpression,
     thresholdMessageKit: ThresholdMessageKit,
     signer?: ethers.Signer
   ): Promise<DecryptionShareSimple[]> {
@@ -67,15 +64,18 @@ export class ThresholdDecrypter {
       provider,
       this.ritualId
     );
-    const contextStr = await conditionExpr
-      .buildContext(provider, {}, signer)
-      .toJson();
-    const { sharedSecrets, encryptedRequests } = this.makeDecryptionRequests(
-      this.ritualId,
-      new Context(contextStr),
-      dkgParticipants,
-      thresholdMessageKit
-    );
+    const wasmContext = await ConditionContext.fromAccessControlPolicy(
+      provider,
+      thresholdMessageKit.acp,
+      signer
+    ).toWASMContext();
+    const { sharedSecrets, encryptedRequests } =
+      await this.makeDecryptionRequests(
+        this.ritualId,
+        wasmContext,
+        dkgParticipants,
+        thresholdMessageKit
+      );
 
     const { encryptedResponses, errors } = await this.porter.cbdDecrypt(
       encryptedRequests,
@@ -117,21 +117,21 @@ export class ThresholdDecrypter {
     );
   }
 
-  private makeDecryptionRequests(
+  private async makeDecryptionRequests(
     ritualId: number,
-    conditionContext: Context,
+    wasmContext: Context,
     dkgParticipants: Array<DkgParticipant>,
     thresholdMessageKit: ThresholdMessageKit
-  ): {
+  ): Promise<{
     sharedSecrets: Record<string, SessionSharedSecret>;
     encryptedRequests: Record<string, EncryptedThresholdDecryptionRequest>;
-  } {
+  }> {
     const decryptionRequest = new ThresholdDecryptionRequest(
       ritualId,
       FerveoVariant.simple,
       thresholdMessageKit.ciphertextHeader,
       thresholdMessageKit.acp,
-      conditionContext
+      wasmContext
     );
 
     const ephemeralSessionKey = this.makeSessionKey();
--- a/src/characters/cbd-recipient.ts
+++ b/src/characters/cbd-recipient.ts
@@ -1,26 +1,22 @@
 import {
-  AccessControlPolicy,
-  AuthenticatedData,
-  Ciphertext,
   combineDecryptionSharesSimple,
   Context,
   DecryptionShareSimple,
-  decryptWithSharedSecret,
   EncryptedThresholdDecryptionRequest,
   EncryptedThresholdDecryptionResponse,
   FerveoVariant,
   SessionSharedSecret,
   SessionStaticSecret,
   ThresholdDecryptionRequest,
+  ThresholdMessageKit,
 } from '@nucypher/nucypher-core';
 import { ethers } from 'ethers';
-import { keccak256 } from 'ethers/lib/utils';
 
 import { DkgCoordinatorAgent, DkgParticipant } from '../agents/coordinator';
 import { ConditionExpression } from '../conditions';
-import { DkgClient, DkgRitual } from '../dkg';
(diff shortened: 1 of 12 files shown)


src/characters/cbd-recipient.ts
```diff
@@ -46,55 +42,25 @@ export class ThresholdDecrypter {
   // Retrieve and decrypt ciphertext using provider and condition expression
   public async retrieveAndDecrypt(
     provider: ethers.providers.Provider,
-    signer: ethers.Signer,
     conditionExpr: ConditionExpression,
```
derekpierre, line 45: You shouldn't need `conditionExpr` here. The condition is already in the threshold message kit object as part of the `acp`.

src/characters/cbd-recipient.ts
```diff
@@ -46,55 +42,25 @@ export class ThresholdDecrypter {
...
-    ciphertext: Ciphertext
+    thresholdMessageKit: ThresholdMessageKit,
+    signer?: ethers.Signer
   ): Promise<Uint8Array> {
-    const acp = await this.makeAcp(provider, signer, conditionExpr, ciphertext);
-
     const decryptionShares = await this.retrieve(
       provider,
       conditionExpr,
```
derekpierre, line 51: Same comment here as above and other places.

src/characters/cbd-recipient.ts
```diff
@@ -154,20 +119,19 @@ export class ThresholdDecrypter {
...
   } {
     const decryptionRequest = new ThresholdDecryptionRequest(
       ritualId,
       FerveoVariant.simple,
-      ciphertext.header,
-      acp,
-      new Context(contextStr)
+      thresholdMessageKit.ciphertextHeader,
+      thresholdMessageKit.acp,
+      conditionContext
```
derekpierre, line 134: 👍

src/characters/enrico.ts
```diff
@@ -51,26 +53,33 @@ export class Enrico {
...
+
+    const headerHash = keccak256(ciphertext.header.toBytes());
+    const authorization = this.keyring.signer.sign(arrayify(headerHash));
+    const acp = new AccessControlPolicy(
+      authenticatedData,
+      authorization.toBEBytes()
     );
-    return { ciphertext, aad };
+
+    return new ThresholdMessageKit(ciphertext, acp);
```
derekpierre, line 83: 👍

test/utils.ts
```diff
@@ -336,56 +348,71 @@ export const fakeTDecFlow = ({
 
     const decryptionShare = aggregate.createDecryptionShareSimple(
       dkg,
-      ciphertext.header,
-      aad,
+      thresholdMessageKit.ciphertextHeader,
+      authenticatedData.aad(),
```
derekpierre, line 352: The authenticated data should never be used outside of encryption - it's only use is to ensure a consistent aad is used for encryption and decryption since the ACP can only be constructed after the data is encrypted (Ciphertext is produced).

```suggestion
      thresholdMessageKit.acp.aad(),
```

test/utils.ts
```diff
@@ -336,56 +348,71 @@ export const fakeTDecFlow = ({
...
 
-  // The client should have access to the public parameters of the DKG
-  const plaintext = decryptWithSharedSecret(ciphertext, aad, sharedSecret);
+  const plaintext = thresholdMessageKit.decryptWithSharedSecret(sharedSecret);
   if (!bytesEqual(plaintext, message)) {
     throw new Error('Decryption failed');
   }
-  return { decryptionShares, sharedSecret, plaintext };
+  return {
+    authenticatedData,
```
derekpierre, line 365: Should not need this - see comment above.

src/characters/cbd-recipient.ts
```diff
@@ -46,71 +41,44 @@ export class ThresholdDecrypter {
...
-      acp
-    );
+    const conditionExpr = thresholdMessageKit.acp.conditions
+      ? ConditionExpression.fromWASMConditions(
+          thresholdMessageKit.acp.conditions
+        )
+      : ConditionExpression.fromObj({
+          version: ConditionExpression.VERSION,
+          condition: {},
+        });
```
piotr-roslaniec, line 73: We need to somehow handle the case where `thresholdMessageKit.acp.conditions` is missing

derekpierre: It should never have non-defined conditions, right?

Conditions should always be defined - that's the change here - https://github.com/nucypher/nucypher-core/pull/80

piotr-roslaniec: Ok, I'm going to keep a work-around in there until we merge & release that change in `nucypher-core`.

derekpierre: 👍 that's fair - but let's track it so it doesn't get lost in the shuffle.

piotr-roslaniec: Tracked here: https://github.com/nucypher/nucypher-ts/issues/293

test/unit/cbd-strategy.test.ts
```diff
@@ -104,20 +103,20 @@ describe('CbdDeployedStrategy', () => {
...
       .encryptMessageCbd(message);
 
     // Setup mocks for `retrieveAndDecrypt`
-    const { decryptionShares } = fakeTDecFlow({
+    const { decryptionShares } = await fakeTDecFlow({
       ...mockedDkg,
       message: toBytes(message),
-      aad,
-      ciphertext,
+      conditionExpr,
```
derekpierre, line 114: Let's do the same thing here, and only use `thresholdMessagerKit` which contains the conditions anyway.

I'm only pointing it out because sometimes devs look to tests for usage patterns.

test/unit/conditions/context.test.ts
```diff
@@ -90,20 +90,16 @@ describe('context parameters', () => {
     };
     const condition = new ContractCondition(conditionObj);
     const conditionExpr = new ConditionExpression(condition);
-    const conditionContext = conditionExpr.buildContext(provider, {}, signer);
```
derekpierre, line 93: Perhaps you can put this after the `contextRequiresSigner` expect on L93, to prove that the condition context can be build if a signer is passed. Perhaps also add a check that a failure occurs if the signer is not provided...?

Of course if this is tested somewhere else then don't bother.

test/unit/conditions/context.test.ts
```diff
@@ -90,20 +90,16 @@ describe('context parameters', () => {
...
     expect(conditionExpr.contextRequiresSigner()).toBe(true);
-    expect(conditionContext.requiresSigner()).toBe(true);
   });
 
   it('detects if a signer is not required', () => {
     const condition = new RpcCondition(testRpcConditionObj);
     const conditionExpr = new ConditionExpression(condition);
-    const conditionContext = conditionExpr.buildContext(provider, {}, signer);
```
derekpierre, line 101: Similar comment as above for completeness. After the `contextRequiresSigner` check on L105, you can ensure that context can be successfully built with/without a signer since one isn't needed.

test/utils.ts
```diff
@@ -304,22 +307,22 @@ interface FakeDkgRitualFlow {
   sharesNum: number;
   threshold: number;
   receivedMessages: ValidatorMessage[];
-  ciphertext: Ciphertext;
-  aad: Uint8Array;
   dkg: Dkg;
   message: Uint8Array;
+  dkgPublicKey: DkgPublicKey;
+  conditionExpr: ConditionExpression;
+  thresholdMessageKit: ThresholdMessageKit;
```
derekpierre, line 314: Same comment here about just using `thresholdMessageKit`, even for testing.

src/characters/cbd-recipient.ts
```diff
@@ -46,71 +42,40 @@ export class ThresholdDecrypter {
...
-      .buildContext(provider, {}, signer)
-      .toJson();
-    const { sharedSecrets, encryptedRequests } = this.makeDecryptionRequests(
-      this.ritualId,
-      ciphertext,
-      contextStr,
-      dkgParticipants,
-      acp
-    );
+    const wasmContext = await ConditionContext.fromAccessControlPolicy(
```
derekpierre, line 67: Seems like  passing the entire acp is unnecessary as opposed to only the conditions, so `fromAccessControlPolicy(...)` seems unnecessary.

This could instead be:
```typescript
ConditionContext.fromConditions(provider, thresholdMessageKit.acp.conditions, signer)
```

src/conditions/context/context.ts
```diff
@@ -49,7 +50,10 @@ export class ConditionContext {
       }
     });
 
-    if (this.requiresSigner() && !this.signer) {
+    const conditionRequiresSigner = this.conditions.some((c) =>
+      c.requiresSigner()
+    );
+    if (conditionRequiresSigner && !this.signer) {
       throw new Error(
         `Cannot use ${USER_ADDRESS_PARAM} as custom parameter without a signer`
```
derekpierre, line 58: Can we change this to something like the following?
```suggestion
        `Condition contains ${USER_ADDRESS_PARAM} context variable and requires a signer to populate`
```

Open to alternative wording.

test/unit/conditions/context.test.ts
```diff
@@ -90,20 +90,32 @@ describe('context parameters', () => {
...
+      returnValueTest: {
+        ...testReturnValueTest,
+        value: USER_ADDRESS_PARAM,
+      },
+    };
+    const condition = new ContractCondition(conditionObj);
+    const conditionExpr = new ConditionExpression(condition);
+    expect(conditionExpr.contextRequiresSigner()).toBe(true);
+    expect(() => conditionExpr.buildContext(provider, {}, undefined)).toThrow(
+      `Cannot use ${USER_ADDRESS_PARAM} as custom parameter without a signer`
```
derekpierre, line 117: See comment about updating the wording of this error.

theref (approved):

manumonti (approved):

manumonti:
LGTM! ✌️ But I left a nitpicky question

derekpierre (approved):
🎸

PR #299, opened 2023-09-21
Improve DX when working with WASM dependencies

piotr-roslaniec:
**Type of PR:**
- Feature
- Refactor

**Required reviews:** 
-  2

**What this does:**
- Introduces changes from https://github.com/nucypher/nucypher-core/pull/83
- Now, instead of using a WASM-aware bundler, the `nucypher-ts` user will have to call the `await initialize()` method imported from `nucypher-ts`
- Updates CJS and ES build config
- Updates examples, removes examples deprecated by this change

**Issues fixed/closed:**
- Refers to https://github.com/nucypher/nucypher-core/pull/83
- Fixes https://github.com/nucypher/nucypher-ts/issues/120
- Fixes https://github.com/nucypher/nucypher-ts/issues/257
- Fixes https://github.com/nucypher/nucypher-ts/issues/256
- Fixes https://github.com/nucypher/nucypher-ts/issues/259
- Fixes https://github.com/nucypher/nucypher-ts/issues/255
- Refers to: https://github.com/nucypher/nucypher-ts/issues/272

**Why it's needed:**
- Removes dependency on WASM-aware bundler for `nucypher-ts` users

**Notes for reviewers:**
- These changes will break compatibility with the previous `nucypher-ts` version because of how the integration works now.
- Based on https://github.com/nucypher/nucypher-ts/pull/297 - Please review it first
- The relevant changes are in the [last commit](https://github.com/nucypher/nucypher-ts/pull/299/commits/0b7ba57a20d05cd2af753ee1afe9acd89e13fd13)
- CI uses a pre-release of `@nucypher/nucypher-core` package
- Is `await initialize()` an improvement from having to configure bundler? You can learn more about the trade-offs in the related PR: https://github.com/nucypher/nucypher-core/pull/83
- Please see `./examples` to compare and contrast this change with the previous DX
- Will address documentation changes post monorepo merge - https://github.com/nucypher/nucypher-ts/issues/300


examples/nextjs/README.md
```diff
@@ -0,0 +1,48 @@
+This is a [Next.js](https://nextjs.org/) project bootstrapped with
+[`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
+
+## Getting Started
```
manumonti, line 4: Before running the dev server, it is necessary to install the packages, right? I tried with npm install, but I got an error:

```
% npm i                                                                                                                                                                                                                    
npm ERR! code EUNSUPPORTEDPROTOCOL
npm ERR! Unsupported URL Type "workspace:": workspace:*
```

But it worked with pnpm: `pnpm install`

Should we add `pnpm install` to this section?

examples/nextjs/README.md
```diff
@@ -0,0 +1,48 @@
...
+yarn dev
+# or
+pnpm dev
+# or
+bun dev
+```
+
+Open [http://localhost:3000](http://localhost:3000) with your browser to see the
+result.
+
```
manumonti, line 20: Should we add a note about Metamask (if it is being used) should have Mumbai as the network selected ?

manumonti (changes requested):
LGTM overall, just a few thoughts

manumonti (approved):

examples/react/src/App.tsx
```diff
@@ -1,49 +1,66 @@
...
+function App() {
+  const [isInit, setIsInit] = useState<boolean>(false);
+  const [provider, setProvider] = useState<
+    ethers.providers.Web3Provider | undefined
+  >();
+  const [alice, setAlice] = useState<Alice | undefined>();
+  const [bob, setBob] = useState<Bob | undefined>();
+  const [policy, setPolicy] = useState<EnactedPolicy>();
+
+  const initNucypher = async () => {
```
cygnusv, line 22: This is an internal function, right? Name is not very descriptive, but if it's only a local thing, it's fine.

KPrasch (approved):
I do think `await initialize()` is an improvement, although I do wish that the error raised when executing nucypher-core functions without calling it was more informative. Is it possible to catch and re-raise? In any case good documentation will mitigate appropriately in my opinion.  Appoved!

🤠

package.json
```diff
@@ -4,6 +4,7 @@
   "license": "GPL-3.0-only",
   "author": "Piotr Roslaniec <p.roslaniec@gmail.com>",
   "scripts": {
+    "postinstall": "pnpm build || true && echo 'Ignoring build errors on setup until we publish the first version'",
```
cygnusv, line 7: We should track this

cygnusv (approved):

package.json
cygnusv: Opened #307

piotr-roslaniec: Tracked here: https://github.com/nucypher/nucypher-ts/issues/308

examples/react/src/App.tsx
piotr-roslaniec: Yes, it's only used in this example as a helper method

package.json
KPrasch: 😆

PR #301, opened 2023-09-22
Refactor packages APIs

piotr-roslaniec:
**Type of PR:**
- Refactor
- Other

**Required reviews:** 
- 2

**Issues fixed/closed:**
- Closes https://github.com/nucypher/nucypher-ts/issues/272
- Closes https://github.com/nucypher/nucypher-ts/issues/274
- Closes https://github.com/nucypher/nucypher-ts/issues/303

**What this does:**
- Continues refactoring of monorepo packages
- Configures tests to use inlined WASM when using `@nucypher/nucypher-core`
- Refactors from character-based to functional API
  - [x] In the `taco` package 
  - ~~In the `pre` package~~ Not a priority
- Updates `taco` for compatibility with `dkg-alpha-13`, including `examples`
- Adds a `demos/taco-demo`

**Notes for reviewers:**
- Based on https://github.com/nucypher/nucypher-ts/pull/299 - Please review it first
- There are some [other demos](https://github.com/piotr-roslaniec/nucypher-ts/blob/refactor-packages/demos/README.md) that were not migrated yet, tracked here: #308


examples/taco/nodejs/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+dotenv.config();
+
+const rpcProviderUrl = process.env.RPC_PROVIDER_URL;
+if (!rpcProviderUrl) {
+  throw new Error('RPC_PROVIDER_URL is not set.');
+}
+
+const privateKey = process.env.PRIVATE_KEY;
+if (!privateKey) {
+  throw new Error('PRIVATE_KEY is not set.');
```
cygnusv, line 22: We have a hardcoded example key for Enrico in `nucypher`. See https://github.com/nucypher/nucypher/blob/4b978025a262f240a9afb9869257c070a33d767d/tests/constants.py#L88

Maybe we can use this here as a fallback?

examples/taco/nodejs/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  const messageKit = await encrypt(
+    provider,
+    message,
+    hasPositiveBalance,
+    ritualId,
+    signer,
+  );
+
+  console.log('Decrypting message...');
+  const porterUri = getPorterUri('lynx'); // Test network
```
cygnusv, line 59: How about we create some constants somewhere like `TacoNetwork.DEV_TESTNET = 'lynx'` and `TacoNetworkTESTNET = 'tapir'`, and we reduce a bit the cognitive load for adopters? After all, how necessary is that adopters know that the testnet is called lynx/tapir?

Thoughts? @piotr-roslaniec @KPrasch @derekpierre

examples/taco/webpack-5/src/index.html
```diff
@@ -0,0 +1,15 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <title>nucypher-ts example</title>
+    <script src="index.js"></script>
+  </head>
+  <body>
+    <noscript
+      >This page contains webassembly and javascript content, please enable
+      javascript in your browser.</noscript
+    >
```
cygnusv, line 12:
```suggestion
    <noscript>This page contains webassembly and javascript content, please enable
      javascript in your browser.</noscript>
```

packages/shared/abis/Coordinator.json
```diff
@@ -1,1375 +1,1423 @@
 [
```
cygnusv, line 1: Opened https://github.com/nucypher/nucypher-ts/issues/310 to use our NPM package instead of manually tracking here. The NPM package probably needs updating, though (cc @manumonti)

packages/shared/src/contracts/registry.ts
```diff
@@ -13,7 +13,7 @@ const POLYGON: Contracts = {
 
 const MUMBAI: Contracts = {
   SUBSCRIPTION_MANAGER: '0xb9015d7b35ce7c81dde38ef7136baa3b1044f313',
-  COORDINATOR: '0x0f019Ade1D34399D946CF2f161386362655Dd1A4',
+  COORDINATOR: '0x8E49989F9D3aD89c8ab0de21FbA2E00C67ca872F',
```
cygnusv, line 16: Opened https://github.com/nucypher/nucypher-ts/issues/310. The bulk of this module should  be externalized to the NPM package. See also 
https://github.com/nucypher/nucypher-contracts/issues/124 for on-going discussion

cygnusv (commented):
First pass. Looking good!

demos/README.md
```diff
@@ -0,0 +1,7 @@
+# `@nucypher/*` Demos
+
+- [`taco-demo`](./taco-demo) - A demo of the `@nucypher/taco` library.
+- [`taco-nft-demo`](./taco-nft-demo) - A demo an NFT-based condition using the `@nucypher/taco` library.
+- [`nucypher-ts-demo`](https://github.com/nucypher/nucypher-ts-demo) - A demo of PRE in the `nucypher-ts` library.
+- [`tdec-sandbox`](https://github.com/nucypher/tdec-sandbox) - A demo of tDec in the `nucypher-ts` library.
+- [`tdec-nft-example`](https://github.com/nucypher/tdec-nft-example) - A demo of tDec in the `nucypher-ts` library.
```
KPrasch, line 7: In the spirit of the monorepo, shall we also internalize these demos to this repo?

KPrasch (approved):
It's so refreshing to see the demos so buttoned up and organized.  🤠

examples/taco/nodejs/src/index.ts
KPrasch: It's another level of abstraction above domains which at this level of public exposure makes sense to me! 👍🏻 

I suggest that we adopt the nucypher/nucypher convention to call these "domains" instead of "networks" for accuracy:

```typescript
TacoDomain.DEVNET; TacoDomain.TESTNET
```

examples/taco/nextjs/src/app/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const provider = new ethers.providers.Web3Provider(window.ethereum!, 'any');
+    await provider.send('eth_requestAccounts', []);
+    const signer = provider.getSigner();
+
+    const {chainId} = await provider.getNetwork();
+    if (chainId !== 80001) {
+      // Switch to Matic Mumbai testnet
+      await window.ethereum!.request!({
+        method: 'wallet_switchEthereumChain',
+        params: [{chainId: '0x13881'}],
```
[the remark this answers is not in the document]
piotr-roslaniec: I'd prefer to keep `examples` simpler and less opinionated and have `demos` show more fully-featured apps instead.

examples/taco/nodejs/src/index.ts
piotr-roslaniec: On one hand, it makes sense - This Enrico is already configured on testnet (allow-listed and so on) but on the other hand, I'm not sure what happens when a number of users start testing their apps with this default character.

examples/taco/nodejs/src/index.ts
piotr-roslaniec: I agree with both suggestions, tracking it in https://github.com/nucypher/nucypher-ts/issues/311

demos/README.md
piotr-roslaniec: Internalizing `nucypher-ts-demo` will require releasing `@nucypher/pre`, which I think we should postpone. `tdec-sandbox` and `tdec-nft-example` will be deprecated (archived) and replaced by existing `demos`.

KPrasch: I also generally agree with the division of examples/demos mentioned above 👍🏻

examples/taco/nodejs/src/index.ts
derekpierre: Agreed 👍

theref (approved):

demos/README.md
derekpierre: > tdec-sandbox and tdec-nft-example will be deprecated (archived) and replaced by existing demos.

^ If we are going to deprecate/archive these, should we remove these in the readme then?

demos/taco-demo/README.md
```diff
@@ -0,0 +1,31 @@
...
+pnpm start
+```
+
+## Usage
+
+In order to run this demo will need a MetaMask with an account funded with some
+$MATIC.
+
+In order to connect with the network, the demo uses a public instances of
+[Porter](https://docs.nucypher.com/en/latest/application_development/web_development.html#running-porter).
```
derekpierre, line 16: Can we point to the threshold docs instead? The Porter docs will improve over time.

```suggestion
[Porter](https://docs.threshold.network/app-development/threshold-access-control-tac/porter).
```

demos/taco-demo/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Polygon
+
+`@nucypher/taco` is in an early release. We recommend **not** using it in
+production _just yet_.
```
derekpierre, line 21: Do we need this for a demo readme?

demos/taco-demo/src/App.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const encryptMessage = async (message: string) => {
+    if (!condition) {
+      return;
+    }
+    setLoading(true);
+
+    await switchNetwork(Mumbai.chainId);
+
+    const provider = new ethers.providers.Web3Provider(window.ethereum);
+    const ritualId = 2; // Replace with your own ritual ID
```
derekpierre, line 43: Is it possible to make this some sort of browser var or some variable that users can easily set somewhere - some kind of .env file equivalent for local scripts? Even if localized to a specific file.

Whatever process for updating this number should end up in the readme/demo instructions.

demos/taco-demo/src/App.tsx
```diff
@@ -0,0 +1,118 @@
+import {
+  conditions,
+  decrypt,
+  encrypt,
+  getPorterUri,
+  initialize,
+  ThresholdMessageKit,
+} from '@nucypher/taco';
```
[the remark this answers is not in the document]
derekpierre: It does look very clean 🎸

demos/taco-nft-demo/README.md
```diff
@@ -0,0 +1,31 @@
...
+In order to run this demo will need a MetaMask with an account funded with some
+$MATIC.
+
+In order to connect with the network, the demo uses a public instances of
+[Porter](https://docs.nucypher.com/en/latest/application_development/web_development.html#running-porter).
+
+### Polygon
+
+`@nucypher/taco` is in an early release. We recommend **not** using it in
+production _just yet_.
```
derekpierre, line 21: Same question as in the other demo.

demos/taco-nft-demo/src/NFTConditionBuilder.tsx
```diff
@@ -0,0 +1,122 @@
...
+  enabled: boolean;
+}
+
+export const NFTConditionBuilder = ({
+  condition,
+  setConditions,
+  enabled,
+}: Props) => {
+  const { library } = useEthers();
+  const goerliNFTAddress = '0x932Ca55B9Ef0b3094E8Fa82435b3b4c50d713043'; // https://goerli-nfts.vercel.app/
```
derekpierre, line 17: Tapir is going to use sepolia (lynx will still use goerli) - so we'll need a new NFT address on sepolia. 

The both will still use Mumbai, of if using an NFT on mumbai is simpler then that works too.

(cc @KPrasch )

examples/taco/nextjs/src/app/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+        comparator: '>',
+        value: 0,
+      },
+    });
+    const ritualId = 2; // Replace with your own ritual ID
+    const messageKit = await encrypt(provider, message, hasPositiveBalance, ritualId, signer);
+
+    console.log('Decrypting message...');
+    const porterUri = getPorterUri('lynx'); // Test network
+    const decryptedMessage = await decrypt(provider, messageKit, signer, porterUri);
```
derekpierre, line 93: These API calls look great!

examples/taco/nextjs/src/app/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+      conditionType: 'rpc',
+      chain: 5,
+      method: 'eth_getBalance',
+      parameters: [':userAddress', 'latest'],
+      returnValueTest: {
+        comparator: '>',
+        value: 0,
+      },
+    });
+    const ritualId = 2; // Replace with your own ritual ID
```
derekpierre, line 88: Similar question about making this some sort of UI input/variable

examples/taco/nodejs/src/index.ts
derekpierre: I don't think we have much of a choice for `tapir` 🤔 , unless:
1. we make ritual initiation public (for say tapir)
2. grant initiator roles to those who ask

only specific Enricos would be allowed (global allow list) to encrypt for rituals.

If neither 1 or 2 are feasible, at the very least we can probably provide more than one test authorized Enrico account private key.

packages/taco/src/conditions/base/rpc.ts
```diff
@@ -11,7 +11,9 @@ export const rpcConditionSchema = z.object({
   conditionType: z.literal(RpcConditionType).default(RpcConditionType),
   chain: createUnionSchema(SUPPORTED_CHAIN_IDS),
   method: z.enum(['eth_getBalance', 'balanceOf']),
-  parameters: z.array(EthAddressOrUserAddressSchema),
+  parameters: z.array(
+    z.union([EthAddressOrUserAddressSchema, z.string(), z.number()]),
```
derekpierre, line 15: Why not just `EthAddressOrUserAddressSchema`? Why also number or string?

PS. love how that typing reads.

packages/taco/src/taco.ts
```diff
@@ -1,34 +1,88 @@
...
-  return await encryptLight(message, condition, dkgRitual.dkgPublicKey);
+
+  const mk = await encryptWithPublicKey(
+    message,
+    condition,
+    dkgRitual.dkgPublicKey,
+    authSigner,
+  );
+
+  const isAuthorized = DkgCoordinatorAgent.isEncryptionAuthorized(
```
derekpierre, line 48: I think this check should be optional, and **not done by default**.

In some/most cases it is ok for Enrico to encrypt before being authorized. Authorization can be granted after encryption and the authorization check only really applies at decryption time (this was a prior question from BqEth).

I get the notion of being proactive about authorization before encryption, but they are different in their timing.

(cc @cygnusv , @KPrasch ).

packages/pre/src/policy.ts
```diff
@@ -121,7 +124,7 @@ export class BlockchainPolicy {
   public async generatePreEnactedPolicy(
     ursulas: readonly Ursula[],
   ): Promise<PreEnactedPolicy> {
-    if (ursulas.length != this.verifiedKFrags.length) {
+    if (ursulas.length !== this.verifiedKFrags.length) {
```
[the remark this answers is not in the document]
KPrasch: Even though in this case it seems pretty clear that we are only comparing two integers, I believe this is a matter of convention around type saftey.

packages/taco/src/taco.ts
KPrasch: Indeed, it's use-case specific, ritual-specific, and may actually be performed by different actors or user stories altogether.

demos/taco-demo/src/App.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  useEffect(() => {
+    initialize();
+    switchNetwork(Mumbai.chainId);
+  }, []);
+
+  const encryptMessage = async (message: string) => {
+    if (!condition) {
+      return;
+    }
```
KPrasch, line 37: it's not obvious to me what the intended flow control is here, are you using `condition` as some sort of status indication?  Why do you return nothing here instead of raising an error?

derekpierre (commented):
Nice work @piotr-roslaniec! Left some comments/suggestions that we should address before merging.

examples/taco/react/src/App.tsx
```diff
@@ -0,0 +1,107 @@
...
+      returnValueTest: {
+        comparator: '>',
+        value: 0,
+      },
+    });
+    const ritualId = 2; // Replace with your own ritual ID
+    const messageKit = await encrypt(provider, message, hasPositiveBalance, ritualId, signer);
+
+    console.log('Decrypting message...');
+    const porterUri = getPorterUri('lynx'); // Test network
```
KPrasch, line 92: see #311

packages/pre/src/policy.ts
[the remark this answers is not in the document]
piotr-roslaniec: In practice, you would almost always use `!==` in TypeScript so using `!=` could be confusing, i.e. there is no reason to not discriminate the data type here.

packages/shared/src/contracts/ethers-typechain/factories/Coordinator__factory.ts
```diff
@@ -8,1377 +8,1425 @@ import type { Coordinator, CoordinatorInterface } from "../Coordinator";
 
 const _abi = [
   {
+    type: "constructor",
```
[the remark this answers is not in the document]
piotr-roslaniec: This code is generates from ABIs, so it will automatically get updated when we switch to #310

packages/taco/src/conditions/base/rpc.ts
piotr-roslaniec: The second parameter to `eth_getBalance` is `blockNumer`, which is a string (as per JSON RPC schema). I thought that accepting a number here would be more intuitive. Does that make sense? Or should we expose strict interfaces (lifted straight from JSON RPC)?

demos/taco-demo/src/App.tsx
piotr-roslaniec: `condition` could be `undefined` in this scope, but that should never happen. This check is for type safety: `tsc` will coalesce `Condition | undefined` type into `undefined` in this scope. I don't think this short demo needs to implement a red path with error messages, validation, etc.

packages/taco/src/conditions/base/rpc.ts
derekpierre: Oh right I got you.

But then we ideally want something like:
```
parameters = (EthAddressOrUserAddressSchema) OR (EthAddressOrUserAddressSchema, Union[string, int]])
```

So something like the following then (my zod is bad)?
```
z.array(
     z.union([
           EthAddressOrUserAddressSchema,
           z.array(
                EthAddressOrUserAddressSchema,
                z.union([
                     z.string(),
                     z.number()]
               ])
          )
   ])
)
```

If the above is the simplest way (perhaps not), might be cleaner to define a custom schema type somewhere else, similar to how we did `EthAddressOrUserAddressSchema`

jMyles:
Running the tests all together or one at a time, they all seem to fail simialrly:

```
 FAIL  |packages| taco/test/ritual.test.ts [ packages/taco/test/ritual.test.ts ]
TypeError: Cannot set property crypto of #<Object> which has only a getter
 ❯ Object.<anonymous> ../node_modules/.pnpm/@nucypher+nucypher-core@0.13.0-alpha.1/node_modules/@nucypher/nucypher-core/dist/node/index_node.cjs:6844:19
```

Is this because the version of nucypher-core needs to be ticked somewhere?  Does this depend on logic introduced in nucypher-core 83?

demos/taco-nft-demo/src/NFTConditionBuilder.tsx
piotr-roslaniec: I will update the demos when we release TACo on Sepolia. I always have to test them in cases like this anyway.

cygnusv (approved):
I reviewed code and ran the demo locally. Worked like a charm. Awesome work, Piotr! We're almost there

examples/taco/nodejs/src/index.ts
piotr-roslaniec: Decided to not pursue this change in the PR, tracking it here: https://github.com/nucypher/nucypher-ts/issues/313

packages/taco/src/conditions/base/rpc.ts
```diff
@@ -11,9 +11,11 @@ export const rpcConditionSchema = z.object({
   conditionType: z.literal(RpcConditionType).default(RpcConditionType),
   chain: createUnionSchema(SUPPORTED_CHAIN_IDS),
   method: z.enum(['eth_getBalance', 'balanceOf']),
-  parameters: z.array(
-    z.union([EthAddressOrUserAddressSchema, z.string(), z.number()]),
-  ),
+  parameters: z.union([
+    z.array(EthAddressOrUserAddressSchema).length(1),
+    // Using tuple here because ordering matters
+    z.tuple([EthAddressOrUserAddressSchema, z.union([z.string(), z.number()])]),
```
derekpierre, line 17: Great - 🚀

packages/taco/test/conditions/base/rpc.test.ts
```diff
@@ -34,4 +35,52 @@ describe('validation', () => {
       },
     });
   });
+
+  describe('parameters', () => {
```
derekpierre, line 39: Love the testing!

PR #314, opened 2023-10-06
Use `@nucypher/nucypher-contracts` contract registry

### LINKED ISSUE #310: Use @nucypher/nucypher-contracts NPM package instead of custom registry
Instead of manually tracking addresses in `packages/shared/src/contracts/registry.ts` module and ABIs in `packages/shared/abis`, we should externalize that to our NPM package.

For the moment, we need to track addresses and ABIs for:
* New Lynx deployments for Coordinator, GlobalAllowlist and TacoApplication
* Existing SubscriptionManager (Polygon Mainnet and Mumbai)

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
- 2

**What this does:**

**Issues fixed/closed:**
- Fixes #310

**Why it's needed:**

**Notes for reviewers:**
- [ ] Get contract addresses, supported networks, etc. from `@nucypher/nucypher-contracts` too
- Consider not storing generated code in the repo


packages/shared/package.json
```diff
@@ -36,12 +36,13 @@
...
     "lint:fix": "pnpm lint --fix",
     "package-check": "package-check",
-    "typechain": "typechain --out-dir=./src/contracts/ethers-typechain --target=ethers-v5 \"abis/**/*.json\"",
+    "typechain": "ts-node scripts/typechain.ts",
     "typedoc": "typedoc"
   },
   "dependencies": {
     "@ethersproject/abi": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
+    "@nucypher/nucypher-contracts": "^0.5.0",
```
KPrasch, line 45:
```suggestion
    "@nucypher/nucypher-contracts": "^0.5.1",
```

packages/shared/package.json
```diff
@@ -36,12 +36,13 @@
...
     "lint:fix": "pnpm lint --fix",
     "package-check": "package-check",
-    "typechain": "typechain --out-dir=./src/contracts/ethers-typechain --target=ethers-v5 \"abis/**/*.json\"",
+    "typechain": "ts-node scripts/typechain.ts",
     "typedoc": "typedoc"
   },
   "dependencies": {
     "@ethersproject/abi": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
+    "@nucypher/nucypher-contracts": "^0.6.0",
```
KPrasch, line 45:
```suggestion
    "@nucypher/nucypher-contracts": "^7.0.0-rc.1",
```

packages/shared/scripts/typechain.ts
```diff
@@ -0,0 +1,78 @@
...
+const parseContractRegistry = (registry: ContractRegistry): Contract[] =>
+  Object.keys(registry)
+    .map((chainId) => {
+      const networkRegistry = registry[chainId];
+      return Object.keys(networkRegistry).map((contractName): Contract => {
+        const contract = networkRegistry[contractName];
+        return { name: contractName as ContractName, abi: contract.abi };
+      });
+    })
+    .flat();
```
KPrasch, line 24: Nice to see this data model from nucypher-contracts adopted here.

packages/shared/abis/Coordinator.json
derekpierre, line 1: 👍 Happy to see these files go away.

demos/taco-demo/src/App.tsx
```diff
@@ -62,12 +64,12 @@ export default function App() {
     setDecryptionErrors([]);
 
     const provider = new ethers.providers.Web3Provider(window.ethereum);
-    const porterUri = getPorterUri('lynx');
     const decryptedMessage = await decrypt(
       provider,
+      domains.DEV,
       encryptedMessage,
+      getPorterUri(domains.DEV),
```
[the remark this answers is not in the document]
piotr-roslaniec: We could, but we would be closing the doors on the "bring your own Porter" idea. I wasn't sure if providing a default here would make sense.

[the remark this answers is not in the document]
derekpierre: You're right. Ignore.

packages/shared/src/contracts/agents/global-allow-list.ts
```diff
@@ -0,0 +1,56 @@
+import { ethers } from 'ethers';
+
+import { Domain } from '../../porter';
+import { ChecksumAddress } from '../../types';
+import { GlobalAllowList, GlobalAllowList__factory } from '../ethers-typechain';
+import { DEFAULT_WAIT_N_CONFIRMATIONS, getContract } from '../registry';
+
+export class GlobalAllowListAgent {
+  public static async registerEncrypters(
```
derekpierre, line 9: Doesn't have to be this PR, but you may want to add a follow-up issue to add the ability to check if an encrypter is already registered i.e. `isAddressAuthorized(ritualid, address)` I think the contract method is called.

derekpierre: You could also use it to check if the encrypter is already authorized before performing the authorization below.

packages/taco/src/taco.ts
```diff
@@ -92,6 +108,23 @@ export const decrypt = async (
 
 export const isAuthorized = async (
   provider: ethers.providers.Provider,
+  domain: Domain,
   messageKit: ThresholdMessageKit,
   ritualId: number,
-) => DkgCoordinatorAgent.isEncryptionAuthorized(provider, ritualId, messageKit);
+) => DkgCoordinatorAgent.isEncryptionAuthorized(provider, domain, ritualId, messageKit);
+
+export const registerEncrypters = async (
```
derekpierre, line 116: This is very specific to the GlobalAllowList. I think we need to be careful here. Rituals may use the GlobalAllowList but I don't believe they have to. We'd need to first check here whether the `accessController` for the associated ritual is the global allow list contract before registering the encrypter. It doesn't have to be, although more often than not, it may be. 

cc @KPrasch

packages/shared/src/contracts/agents/global-allow-list.ts
piotr-roslaniec: These two methods are [exposed by `taco`](https://github.com/nucypher/nucypher-ts/pull/314/files#diff-94274562e7c2dacfa0fd63a28f2d134b6c4b4a918761053b83b816b6a065adcbR109-R116). I think the latter (registering encrypters) could use the former under the hood.

packages/taco/src/taco.ts
piotr-roslaniec: Yes, the encryptor onboarding is a bit more complicated and it depends on https://github.com/nucypher/nucypher-ts/pull/306

We could hide this method from the `taco` API until we sort it out.

derekpierre: 👍 I think that would be prudent.

cygnusv (approved):

derekpierre (approved):
🎸 - good work. 

Approved, but let's hide the `registerEncrypters` function from the API for now until we sort out the proper logic. Let's ensure there's a follow-up issue for tracking.

examples/taco/webpack-5/src/index.html
```diff
@@ -6,7 +6,8 @@
     <script src="index.js"></script>
   </head>
   <body>
-    <noscript>This page contains webassembly and javascript content, please enable
+    <noscript
+      >This page contains webassembly and javascript content, please enable
```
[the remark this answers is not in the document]
piotr-roslaniec: I guess I could play around with linter rules, but I don't think these files are read as frequently.

packages/shared/src/contracts/ethers-typechain/common.ts
```diff
@@ -40,5 +40,5 @@ export type GetContractTypeFromFactory<F> = F extends MinEthersFactory<
   : never;
 
 export type GetARGsTypeFromFactory<F> = F extends MinEthersFactory<any, any>
-  ? Parameters<F["deploy"]>
+  ? Parameters<F['deploy']>
```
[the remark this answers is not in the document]
piotr-roslaniec: It's an oversight on my part. They should be generated and formatted in one step (to prevent line changes), but I haven't set it up yet.

PR #316, opened 2023-10-09
Taco Node-js Demo as GH Action

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Creates a GH action that runs the node-js example on Lynx
   * Action is triggered by pushing PRs, manually, and every 20 minutes.
   * It will give us valuable information about state of Lynx testnet and API stability

**Notes for reviewers:**
Sister PR for Python API: https://github.com/nucypher/nucypher/pull/3274

Improvements for later:
   * The 20 minutes schedule can be reconsidered soon, once we gain a bit of confidence in both Lynx and the GH action workflow.
   * The workflow can be improved with more informative diagnosis in case of failures
   * For the moment, it uses many hard-coded values (ritual ID, encryptor signing key, RPC provider, etc). There's lot of room for improvement there too.

### DIFF
--- a/.github/workflows/lynx.yml
+++ b/.github/workflows/lynx.yml
@@ -1,0 +1,53 @@
+name: '🔎 Lynx Testnet Example'
+
+on:
+  schedule:
+    - cron: "0/20 * * * *"  # Every 20 minutes
+  pull_request:
+    # branches:
+    #   - main
+    #   - development
+    #   - epic-*
+  workflow_dispatch:
+
+env:
+  RPC_PROVIDER_URL: ${{ vars.INFURA_MUMBAI }}
+  PRIVATE_KEY: ${{ vars.TESTNET_ENRICO_PRIVATE_KEY }}
+
+jobs:
+  build_project:
+    name: '🔎 Lynx Testnet Example on Node ${{ matrix.node }} and ${{ matrix.os }}'
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        node: [ '18.x' ]
+        os: [ ubuntu-latest ]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 8.1
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'pnpm'
+
+      - name: Lynx testnet node-js script installation
+        working-directory: ./examples/taco/nodejs 
+        run: pnpm install
+
+      - name: Lynx testnet node-js script environment
+        working-directory: ./examples/taco/nodejs 
+        run: |
+          echo RPC_PROVIDER_URL=${{ vars.INFURA_MUMBAI }} > .env
+          echo PRIVATE_KEY=${{ vars.TESTNET_ENRICO_PRIVATE_KEY }} >> .env
+          cat .env
+
+      - name: Lynx testnet node-js script execution
+        id: lynx_script
+        working-directory: ./examples/taco/nodejs 
+        run: pnpm start
--- a/.github/workflows/lynx.yml
+++ b/.github/workflows/lynx.yml
@@ -10,9 +10,9 @@ on:
     #   - epic-*
   workflow_dispatch:
 
-env:
-  RPC_PROVIDER_URL: ${{ vars.INFURA_MUMBAI }}
-  PRIVATE_KEY: ${{ vars.TESTNET_ENRICO_PRIVATE_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/lynx.yml
```diff
@@ -0,0 +1,52 @@
...
+  pull_request:
+    # branches:
+    #   - main
+    #   - development
+    #   - epic-*
+  workflow_dispatch:
+
+env:  # TODO: Use variables when GH supports it for forks. See https://github.com/orgs/community/discussions/44322
+  RPC_PROVIDER_URL: "https://polygon-mumbai.infura.io/v3/3747007a284045d483c342fb39889a30"
+  PRIVATE_KEY: "0x900edb9e8214b2353f82aa195e915128f419a92cfb8bbc0f4784f10ef4112b86"
```
manumonti, line 15: Are we sure we want to expose this infura endpoint and this private key?

If someone take this private key and drain all the founds, this workflow will not pass, right?

Also, if someone drain the free queries of the infura endpoint, also this will fail.

Maybe I'm missing something.

manumonti: As far as I know, there is an option to use secrets in GitHub:

https://docs.github.com/en/actions/using-workflows/sharing-workflows-secrets-and-runners-with-your-organization#sharing-secrets-and-variables-within-an-organization

manumonti (changes requested):
Looks good in general, I just left a concern about publishing that keys.

.github/workflows/lynx.yml
cygnusv: Re: Infura endpoint, it is configured with allowlist for Taco contracts and certain JSON-RPC methods. We could even consider adding rate limits and origin restrictions, if necessary.

Re: private key, this is a test private key that's already part of the nucypher codebase.

Re: secrets in GH, the problem is that these are not accessible by pull requests triggered from forks, which is not convenient for us. That's even the case for non-secret GH variables! See for a recent support discussion in GH: https://github.com/orgs/community/discussions/44322
In any case, these particular private key and infura endpoint can be published because of the reasons mentioned earlier, so for the moment it's OK.

.github/workflows/lynx.yml
```diff
@@ -0,0 +1,52 @@
+name: '🔎 Lynx Testnet Example'
+
+on:
+  schedule:
+    - cron: "0/20 * * * *"  # Every 20 minutes
+  pull_request:
+    # branches:
+    #   - main
+    #   - development
+    #   - epic-*
```
derekpierre, line 10: Is this intentional? Does this mean for any PR then?

examples/taco/nodejs/src/index.ts
```diff
@@ -46,7 +47,7 @@ const runExample = async () => {
     hasPositiveBalance.requiresSigner(),
     'Condition requires signer',
   );
-  const ritualId = 2; // Replace with your own ritual ID
+  const ritualId = 1; // Replace with your own ritual ID
```
derekpierre, line 50: Similar comment in python demo job. We should make this modifiable without needing to change the code. Obviously not necessarily for this PR, but something to follow-up on.

.github/workflows/lynx.yml
derekpierre: I'm fine with the restricted Infura endpoint. Another option you could consider is a public endpoint for mumbai...e.g. https://rpc-mumbai.polygon.technology (from https://wiki.polygon.technology/docs/pos/reference/rpc-endpoints/)

manumonti: Ok, I still think that someone draining the wallet can cause some trouble, but we can proceed with this and see if this "attack" happens or not.

manumonti (approved):
LGTM! ✌️

.github/workflows/lynx.yml
cygnusv: Yes, at least for the moment since we're working on the monorepo branch. But I also think that what's the problem about running this for every PR?

derekpierre: No real issue - just wondering if this was commented out on purpose.

derekpierre (approved):
🎸

examples/taco/nodejs/src/index.ts
KPrasch: Agreed, this needs to be an input parameter at some layer.

cygnusv: I don't mind we do that, but is that a comment specific to running the example as a GH action, or for the example script in general?

PR #323, opened 2023-10-17
Expose custom context params in TACo API

### LINKED ISSUE #322: Allow passing custom params when decrypting
The current `taco` API does not easily allow for passing custom context params when decrypting. The core of the problem is that the `retrieve` function calls `ConditionContext.fromConditions`

https://github.com/nucypher/nucypher-ts/blob/26c6fb51b7659827cc8f9df716b7d828c1af69d2/packages/taco/src/tdec.ts#L83-L87

which defaults the custom params to an empty object with no opportunity to override with `withCustomParams` before calling `toWASMContext()`

https://github.com/nucypher/nucypher-ts/blob/26c6fb51b7659827cc8f9df716b7d828c1af69d2/packages/taco/src/conditions/context/context.ts#L161

This prevents adding any user supplied custom params at decryption time (which we need for our usecase)

_side note:_

When there is a function that takes many inputs like `retrieveAndDecrypt` (especially when some of the inputs have the same type like ritualId and threshold) I think it would be nicer to pass a single object. This way you get named params during input. Not a deal breaker or anything just a suggestion

https://github.com/nucypher/nucypher-ts/blob/26c6fb51b7659827cc8f9df716b7d828c1af69d2/packages/taco/src/tdec.ts#L50C1-L57C28

```typescript
export const retrieveAndDecrypt = async (params: {
  porterUri: string,
  thresholdMessageKit: ThresholdMessageKit,
  ritualId: number,
  threshold: number,
  signer?: ethers.Signer
  },
): Promise<Uint8Array> => {
...
}
```

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
- 1

**Issues fixed/closed:**
- Fixes #322

### DIFF
--- a/packages/taco/src/conditions/context/context.ts
+++ b/packages/taco/src/conditions/context/context.ts
@@ -153,11 +153,12 @@ export class ConditionContext {
   public static fromConditions(
     provider: ethers.providers.Provider,
     conditions: WASMConditions,
+    customParameters: Record<string, CustomContextParam>,
     signer?: ethers.Signer,
   ): ConditionContext {
     const innerConditions = [
       ConditionExpression.fromWASMConditions(conditions).condition,
     ];
-    return new ConditionContext(provider, innerConditions, {}, signer);
+    return new ConditionContext(provider, innerConditions, customParameters, signer);
   }
 }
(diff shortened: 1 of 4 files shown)


manumonti (approved):
LGTM! 👍

packages/taco/src/conditions/context/context.ts
```diff
@@ -153,11 +153,12 @@ export class ConditionContext {
   public static fromConditions(
     provider: ethers.providers.Provider,
     conditions: WASMConditions,
+    customParameters: Record<string, CustomContextParam>,
```
derekpierre, line 156: Should this be optional?

derekpierre:
@piotr-roslaniec is there the possibility of adding a test for this addition?

derekpierre:
^ To clarify - perhaps not something for this PR but nonetheless something to think about.

PR #353, opened 2023-10-20
Fix contract registry not parsing contract artifacts correctly

piotr-roslaniec:
**Type of PR:**
- Bugfix

**Required reviews:**
- 1

**What this does:**
- Fixes not using the default import in contract registry import (from `nucypher-contracts` package)
- Fixes linter
- Releases new RC version

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**
- You can see the actual fix [here](https://github.com/nucypher/nucypher-ts/pull/353/files#r1367686471). The remainder of this PR are chores that have to do fix fixing the lining and releasing the fixed RC version.

### DIFF
--- a/packages/shared/src/contracts/registry.ts
+++ b/packages/shared/src/contracts/registry.ts
@@ -1,6 +1,6 @@
-import * as lynxRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/lynx.json';
-import * as mainnetRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/mainnet.json';
-import * as tapirRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/tapir.json';
+import lynxRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/lynx.json';
+import mainnetRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/mainnet.json';
+import tapirRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/tapir.json';
 
 import { Domain } from '../porter';
 import { ChecksumAddress } from '../types';
--- a/packages/shared/test/registry.spec.ts
+++ b/packages/shared/test/registry.spec.ts
@@ -1,0 +1,20 @@
+import { describe, expect, it } from 'vitest';
+
+import { ChainId, ContractName, getContract } from '../src';
+
+const testCases: [string, number, ContractName][] = [
+  ['lynx', ChainId.MUMBAI, 'Coordinator'],
+  ['lynx', ChainId.MUMBAI, 'GlobalAllowList'],
+  ['tapir', ChainId.MUMBAI, 'Coordinator'],
+  ['tapir', ChainId.MUMBAI, 'GlobalAllowList'],
+];
+
+describe('registry', () => {
+  for (const testCase of testCases) {
+    const [domain, chainId, contract] = testCase;
+    it(`should for domain ${domain}, chainId ${chainId}, contract ${contract}`, () => {
+      const contractAddress = getContract(domain, chainId, contract);
+      expect(contractAddress).toBeDefined();
+    });
+  }
+});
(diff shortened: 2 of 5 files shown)


packages/shared/src/contracts/registry.ts
```diff
@@ -1,6 +1,6 @@
-import * as lynxRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/lynx.json';
-import * as mainnetRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/mainnet.json';
-import * as tapirRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/tapir.json';
+import lynxRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/lynx.json';
+import mainnetRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/mainnet.json';
+import tapirRegistryJson from '@nucypher/nucypher-contracts/deployment/artifacts/tapir.json';
```
piotr-roslaniec, line 3: The root cause of the issue

KPrasch (approved):

packages/shared/test/registry.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  ['tapir', ChainId.MUMBAI, 'GlobalAllowList'],
+  ['tapir', ChainId.MUMBAI, 'SubscriptionManager'],
+];
+
+describe('registry', () => {
+  for (const testCase of testCases) {
+    const [domain, chainId, contract] = testCase;
+    it(`should for domain ${domain}, chainId ${chainId}, contract ${contract}`, () => {
+      const contractAddress = getContract(domain, chainId, contract);
+      expect(contractAddress).toBeDefined();
```
derekpierre, line 19: @piotr-roslaniec I think we can harden this test a bit more since it is only testing existence at the moment.
1. Compare the same contract from each domain and ensure the addresses are different i.e. address for Coordinator from `lynx` is not equal to Coordinator address from `tapir` etc.
2. Is it possible to obtain contract addresses from chain `5` (`lynx`) from chain `11155111` (`tapir`) - eg. `TACoApplication`? We don't have to do anything with the contract address (this can be a `toBeDefined()`), but the fact that we can get it proves the correct registry is used.

piotr-roslaniec: I think I prefer to move these tests along with typings to `@nucypher/nucypher-contracts` where they really belong. I only added them here so as not to trip over this issue again.

Related issue: https://github.com/nucypher/nucypher-contracts/issues/161

derekpierre: Not sure I follow. I thought the goal here (in this test) is to test `getContract(...)` which is a `nucypher-ts` function, and the issue was that it was not resolving the artifacts correctly. Currently, the test is only testing that the contract address is defined, it's not actually ensuring that `getContract(...)` resolves the address correctly based on `domain`. For example, the test can pass if for instance the same contract registry was used for `lynx`/`tapir` domains. 

Now of course we don't want to check the address value itself because it can change for testnet over time, but what we want to make sure is that the resolution is doing what we expect, hence my recommendation above ensuring the returned addresses are at least different. 

That is irrespective of `nucypher-contracts`.

piotr-roslaniec: The reason why the `getContract` function exists, to begin with, is that there is no API in the contract registry that can guarantee a behavior that I'm looking for: "Give me a contract that matches this network/domain/etc.".

I intend to implement this API in the `@nucypher/nucypher-contracts` package and deprecate the existing `getContract` method.

derekpierre: Is the expectation that this will be completed for `7.0.0`?

piotr-roslaniec: I don't see this as a priority for `7.0.0`

derekpierre: Ok. If it's potentially not going into 7.0.0, I'd recommend updating the test since `getContract` is what we'll be using,. Whenever `getContract` goes away then we can remove these tests.

KPrasch: > I intend to implement this API in the @nucypher/nucypher-contracts package and deprecate the existing getContract method.

good idea 💡

PR #366, opened 2023-10-25
Encode bytes in `Context` params to hex

piotr-roslaniec:
**Type of PR:**
- Bugfix

**Required reviews:**
- 1

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/packages/shared/src/utils.ts
+++ b/packages/shared/src/utils.ts
@@ -24,9 +24,9 @@ export const toBase64 = (bytes: Uint8Array): string =>
 export const fromBase64 = (str: string): Uint8Array =>
   Uint8Array.from(atob(str), (c) => c.charCodeAt(0));
 
-export const base64ToU8Receiver = (_key: string, value: unknown) => {
-  if (typeof value === 'string' && value.startsWith('base64:')) {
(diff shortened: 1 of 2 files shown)


packages/shared/src/utils.ts
```diff
@@ -44,22 +44,22 @@ const sortedReplacer = (_key: string, value: unknown) => {
...
   if (value instanceof Uint8Array) {
-    return `base64:${toBase64(value)}`;
+    return `0x${toHexString(value)}`;
   }
   return value;
 };
 
 const sortedSerializingReplacer = (_key: string, value: unknown): unknown => {
-  const serializedValue = u8ToBase64Replacer(_key, value);
+  const serializedValue = u8ToHexReplacer(_key, value);
```
derekpierre, line 55: Would this also implicitly apply to `paremeters` sent in the condition i.e. if a `parameter` value was in bytes...?

piotr-roslaniec: Yes, this is what all JSON serialization methods are using in `nucypher-ts`.

derekpierre: 👍

PR #367, opened 2023-10-25
Quality of life updates to conditions

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:**
- 2

**What this does:**
- Adds `or` and `and` shorthand methods to `CompoundCondition`
- Removes mandatory `conditionType` from conditions constructors. It's now inferred inside the constructor. The condition props still contain the `conditionType` field to ensure it's validation.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**
- Based on https://github.com/nucypher/nucypher-ts/pull/365, please review it first

### DIFF
--- a/examples/taco/nextjs/src/app/page.tsx
+++ b/examples/taco/nextjs/src/app/page.tsx
@@ -73,7 +73,6 @@ function App() {
 
     console.log('Encrypting message...');
     const hasPositiveBalance = new conditions.RpcCondition({
-      conditionType: 'rpc',
       chain: 5,
       method: 'eth_getBalance',
       parameters: [':userAddress', 'latest'],
--- a/packages/taco/src/conditions/base/index.ts
+++ b/packages/taco/src/conditions/base/index.ts
@@ -5,14 +5,24 @@ import {
 } from '../compound-condition';
 import { Condition, ConditionProps } from '../condition';
 
-import { ContractConditionProps, contractConditionSchema, ContractConditionType } from './contract';
+import {
+  ContractConditionProps,
+  contractConditionSchema,
+  ContractConditionType,
+} from './contract';
 import { RpcConditionProps, rpcConditionSchema, RpcConditionType } from './rpc';
-import { TimeConditionProps, timeConditionSchema, TimeConditionType } from './time';
+import {
+  TimeConditionProps,
+  timeConditionSchema,
+  TimeConditionType,
+} from './time';
 
-type OmitConditionType<T> = Omit<T, 'conditionType'>
+type OmitConditionType<T> = Omit<T, 'conditionType'>;
 
 // Exporting classes here instead of their respective schema files to
 // avoid circular dependency on Condition class.
+
+type ConditionOrProps = Condition | ConditionProps;
 
 export class CompoundCondition extends Condition {
   constructor(value: OmitConditionType<CompoundConditionProps>) {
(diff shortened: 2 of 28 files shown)


packages/taco/src/conditions/base/index.ts
```diff
@@ -1,37 +1,70 @@
...
+  constructor(value: OmitType<CompoundConditionProps>) {
+    super(compoundConditionSchema, {
+      conditionType: CompoundConditionType,
+      ...value,
+    });
+  }
+
+  private static withOperator(
+    operands: ConditionProps[],
+    operator: 'or' | 'and',
```
derekpierre, line 27: Is there already an issue filed to deal with `not` compound conditions in 7.0.0? This is completed on the `nucypher` side -https://github.com/nucypher/nucypher/pull/3293

packages/taco/src/conditions/base/index.ts
```diff
@@ -1,37 +1,70 @@
...
+import { Condition, ConditionProps } from '../condition';
 
-import { ContractConditionProps, contractConditionSchema } from './contract';
-import { RpcConditionProps, rpcConditionSchema } from './rpc';
-import { TimeConditionProps, timeConditionSchema } from './time';
+import { ContractConditionProps, contractConditionSchema, ContractConditionType } from './contract';
+import { RpcConditionProps, rpcConditionSchema, RpcConditionType } from './rpc';
+import { TimeConditionProps, timeConditionSchema, TimeConditionType } from './time';
+
+type OmitType<T> = Omit<T, 'conditionType'>
```
derekpierre, line 12: `OmitType<...>` reads a little odd for me, like "omit the type that's being provided to it" almost eg. OmitType<Foo> almost reads like "omit the Foo type" (to me).

What do you think about `OmitConditionType<...>` or `OmitConditionTypeFrom<...>`? or something like that?

packages/taco/test/conditions/base/contract.test.ts
```diff
@@ -43,6 +43,11 @@ describe('validation', () => {
       },
     });
   });
+
+  it('infers condition type from constructor', ()=> {
+    const condition = new ContractCondition(testContractConditionObj);
+    expect(condition.value.conditionType).toEqual(ContractConditionType);
+  })
```
derekpierre, line 50: Seems like `testContractConditionObj` does contain `conditionType` in it, does it also make sense to additionally test removing the value from `testContractConditionObj` and then passing it in? This ensures that the Omit functionality is ok if the `conditionType` isn't provided. Does that make sense or would that situation never happen? If it makes sense, similar comment for other analogous tests that infer type.

packages/taco/test/conditions/compound-condition.test.ts
```diff
@@ -166,4 +165,36 @@ describe('validation', () => {
     expect(result.error).toBeDefined();
     expect(result.data).toBeUndefined();
   });
+
+  it('accepts shorthand for "or" operator', () => {
```
derekpierre, line 169: Nice test.

packages/taco/src/conditions/base/index.ts
piotr-roslaniec: Filed an issue here: https://github.com/nucypher/nucypher-ts/issues/369

packages/taco/src/conditions/base/index.ts
piotr-roslaniec: I thought it was a bit wordy, but it make sense

derekpierre (approved):
🎸 - looks good.

There is a failing CI job.

examples/taco/nextjs/src/app/page.tsx
```diff
@@ -73,16 +73,15 @@ function App() {
 
     console.log('Encrypting message...');
     const hasPositiveBalance = new conditions.RpcCondition({
-      conditionType: 'rpc',
-      chain: 5,
+      chain: 80001,
```
derekpierre, line 76: @piotr-roslaniec / @KPrasch Does changing the chain from 5 -> 80001 affect any default behaviour / expectations of the demo...?

piotr-roslaniec: It doesn't but it makes the example more consistent (everything is happening on the same chain).

KPrasch: The tapir nodes are not evaluating chain 5 for conditions, but 80001 is used for both testnet and devnet.  I figure this gives the examples/demos the best shot at working "out of the box" regardless of which testnet you use.

KPrasch (approved):

PR #379, opened 2023-11-03
Fix `lynx` release

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:**
-  2

**What this does:**
- Updates demos & demo config for `lynx` network
- Removes optional `index` field

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/packages/taco/src/conditions/base/contract.ts
+++ b/packages/taco/src/conditions/base/contract.ts
@@ -4,7 +4,7 @@ import { z } from 'zod';
 import { ETH_ADDRESS_REGEXP } from '../const';
 
 import { rpcConditionSchema } from './rpc';
-import { ParamOrContextParamSchema } from './shared';
+import { paramOrContextParamSchema } from './shared';
 
 // TODO: Consider replacing with `z.unknown`:
 //    Since Solidity types are tied to Solidity version, we may not be able to accurately represent them in Zod.
@@ -81,7 +81,7 @@ export const contractConditionSchema = rpcConditionSchema
     standardContractType: z.enum(['ERC20', 'ERC721']).optional(),
     method: z.string(),
     functionAbi: functionAbiSchema.optional(),
-    parameters: z.array(ParamOrContextParamSchema),
+    parameters: z.array(paramOrContextParamSchema),
   })
   // Adding this custom logic causes the return type to be ZodEffects instead of ZodObject
   // https://github.com/colinhacks/zod/issues/2474
--- a/packages/taco/src/conditions/base/contract.ts
+++ b/packages/taco/src/conditions/base/contract.ts
@@ -4,7 +4,7 @@ import { z } from 'zod';
 import { ETH_ADDRESS_REGEXP } from '../const';
 
 import { rpcConditionSchema } from './rpc';
-import { ParamOrContextParamSchema, returnValueTestSchema } from './shared';
+import { ParamOrContextParamSchema } from './shared';
 
 // TODO: Consider replacing with `z.unknown`:
 //    Since Solidity types are tied to Solidity version, we may not be able to accurately represent them in Zod.
@@ -82,9 +82,6 @@ export const contractConditionSchema = rpcConditionSchema
     method: z.string(),
     functionAbi: functionAbiSchema.optional(),
     parameters: z.array(ParamOrContextParamSchema),
-    returnValueTest: returnValueTestSchema.extend({
-      value: ParamOrContextParamSchema,
-    }),
   })
   // Adding this custom logic causes the return type to be ZodEffects instead of ZodObject
   // https://github.com/colinhacks/zod/issues/2474
--- a/packages/taco/src/conditions/base/contract.ts
+++ b/packages/taco/src/conditions/base/contract.ts
@@ -4,6 +4,7 @@ import { z } from 'zod';
 import { ETH_ADDRESS_REGEXP } from '../const';
 
 import { rpcConditionSchema } from './rpc';
+import { ParamOrContextParamSchema, returnValueTestSchema } from './shared';
 
 // TODO: Consider replacing with `z.unknown`:
 //    Since Solidity types are tied to Solidity version, we may not be able to accurately represent them in Zod.
@@ -80,7 +81,10 @@ export const contractConditionSchema = rpcConditionSchema
     standardContractType: z.enum(['ERC20', 'ERC721']).optional(),
     method: z.string(),
     functionAbi: functionAbiSchema.optional(),
-    parameters: z.array(z.unknown()),
+    parameters: z.array(ParamOrContextParamSchema),
+    returnValueTest: returnValueTestSchema.extend({
+      value: ParamOrContextParamSchema,
+    }),
   })
   // Adding this custom logic causes the return type to be ZodEffects instead of ZodObject
   // https://github.com/colinhacks/zod/issues/2474
--- a/packages/taco/src/conditions/base/rpc.ts
+++ b/packages/taco/src/conditions/base/rpc.ts
@@ -20,7 +20,7 @@ export const rpcConditionSchema = z.object({
     // Using tuple here because ordering matters
     z.tuple([EthAddressOrUserAddressSchema, paramOrContextParamSchema]),
   ]),
-  returnValueTest: returnValueTestSchema,
+  returnValueTest: returnValueTestSchema, // Update to allow multiple return values after expanding supported methods
 });
 
 export type RpcConditionProps = z.infer<typeof rpcConditionSchema>;
--- a/packages/taco/src/conditions/base/rpc.ts
+++ b/packages/taco/src/conditions/base/rpc.ts
@@ -3,7 +3,12 @@ import { z } from 'zod';
 import { SUPPORTED_CHAIN_IDS } from '../const';
 import createUnionSchema from '../zod';
 
-import { EthAddressOrUserAddressSchema, returnValueTestSchema } from './shared';
+import {
+  ContextParamOrNumberSchema,
+  EthAddressOrUserAddressSchema,
+  ParamOrContextParamSchema,
+  returnValueTestSchema,
+} from './shared';
 
 export const RpcConditionType = 'rpc';
 
@@ -12,11 +17,13 @@ export const rpcConditionSchema = z.object({
   chain: createUnionSchema(SUPPORTED_CHAIN_IDS),
   method: z.enum(['eth_getBalance', 'balanceOf']),
   parameters: z.union([
-    z.array(EthAddressOrUserAddressSchema).length(1),
+    z.array(EthAddressOrUserAddressSchema).nonempty(),
     // Using tuple here because ordering matters
-    z.tuple([EthAddressOrUserAddressSchema, z.union([z.string(), z.number()])]),
+    z.tuple([EthAddressOrUserAddressSchema, ParamOrContextParamSchema]),
   ]),
-  returnValueTest: returnValueTestSchema,
+  returnValueTest: returnValueTestSchema.extend({
+    value: ContextParamOrNumberSchema,
+  }),
 });
 
 export type RpcConditionProps = z.infer<typeof rpcConditionSchema>;
--- a/packages/taco/src/conditions/base/rpc.ts
+++ b/packages/taco/src/conditions/base/rpc.ts
@@ -4,7 +4,6 @@ import { SUPPORTED_CHAIN_IDS } from '../const';
 import createUnionSchema from '../zod';
 
 import {
-  ContextParamOrNumberSchema,
   EthAddressOrUserAddressSchema,
   ParamOrContextParamSchema,
   returnValueTestSchema,
@@ -21,9 +20,7 @@ export const rpcConditionSchema = z.object({
     // Using tuple here because ordering matters
     z.tuple([EthAddressOrUserAddressSchema, ParamOrContextParamSchema]),
   ]),
-  returnValueTest: returnValueTestSchema.extend({
-    value: ContextParamOrNumberSchema,
-  }),
+  returnValueTest: returnValueTestSchema,
 });
 
 export type RpcConditionProps = z.infer<typeof rpcConditionSchema>;
--- a/packages/taco/src/conditions/base/rpc.ts
+++ b/packages/taco/src/conditions/base/rpc.ts
@@ -14,7 +14,7 @@ export const RpcConditionType = 'rpc';
 export const rpcConditionSchema = z.object({
   conditionType: z.literal(RpcConditionType).default(RpcConditionType),
   chain: createUnionSchema(SUPPORTED_CHAIN_IDS),
-  method: z.enum(['eth_getBalance', 'balanceOf']),
+  method: z.enum(['eth_getBalance']),
   parameters: z.union([
     z.array(EthAddressOrUserAddressSchema).nonempty(),
     // Using tuple here because ordering matters
--- a/packages/taco/src/conditions/base/rpc.ts
+++ b/packages/taco/src/conditions/base/rpc.ts
@@ -5,7 +5,7 @@ import createUnionSchema from '../zod';
 
 import {
   EthAddressOrUserAddressSchema,
-  ParamOrContextParamSchema,
+  paramOrContextParamSchema,
   returnValueTestSchema,
 } from './shared';
 
@@ -18,7 +18,7 @@ export const rpcConditionSchema = z.object({
   parameters: z.union([
     z.array(EthAddressOrUserAddressSchema).nonempty(),
     // Using tuple here because ordering matters
-    z.tuple([EthAddressOrUserAddressSchema, ParamOrContextParamSchema]),
+    z.tuple([EthAddressOrUserAddressSchema, paramOrContextParamSchema]),
   ]),
   returnValueTest: returnValueTestSchema,
 });
--- a/packages/taco/src/conditions/base/shared.ts
+++ b/packages/taco/src/conditions/base/shared.ts
@@ -12,10 +12,7 @@ export const contextParamSchema = z.string().regex(CONTEXT_PARAM_REGEXP);
 // If a string starts with `:`, it's a ContextParam
 export const plainStringSchema = z.string().refine(
   (str) => {
-    if (str.startsWith(CONTEXT_PARAM_PREFIX)) {
-      return str.match(CONTEXT_PARAM_REGEXP);
-    }
-    return true;
+    return !str.startsWith(CONTEXT_PARAM_PREFIX);
   },
   {
     message: 'Context parameters must start with ":"',
--- a/packages/taco/src/conditions/base/shared.ts
+++ b/packages/taco/src/conditions/base/shared.ts
@@ -7,10 +7,10 @@ import {
 } from '../const';
 import { CONTEXT_PARAM_PREFIX } from '../context/context';
 
-export const ContextParamSchema = z.string().regex(CONTEXT_PARAM_REGEXP);
+export const contextParamSchema = z.string().regex(CONTEXT_PARAM_REGEXP);
 // We want to discriminate between ContextParams and plain strings
 // If a string starts with `:`, it's a ContextParam
-export const PlainStringSchema = z.string().refine(
+export const plainStringSchema = z.string().refine(
   (str) => {
     if (str.startsWith(CONTEXT_PARAM_PREFIX)) {
       return str.match(CONTEXT_PARAM_REGEXP);
@@ -22,26 +22,26 @@ export const PlainStringSchema = z.string().refine(
   },
 );
 
-export const ParamPrimitiveTypeSchema = z.union([
-  PlainStringSchema,
+export const paramPrimitiveTypeSchema = z.union([
+  plainStringSchema,
   z.boolean(),
   z.number(),
 ]);
-export const ParamSchema = z.union([
-  ParamPrimitiveTypeSchema,
-  z.array(ParamPrimitiveTypeSchema),
+export const paramSchema = z.union([
+  paramPrimitiveTypeSchema,
+  z.array(paramPrimitiveTypeSchema),
 ]);
 
-export const ParamOrContextParamSchema: z.ZodSchema = z.union([
-  ParamSchema,
-  ContextParamSchema,
-  z.lazy(() => z.array(ParamOrContextParamSchema)),
+export const paramOrContextParamSchema: z.ZodSchema = z.union([
+  paramSchema,
+  contextParamSchema,
+  z.lazy(() => z.array(paramOrContextParamSchema)),
 ]);
 
 export const returnValueTestSchema = z.object({
   index: z.number().int().nonnegative().optional(),
   comparator: z.enum(['==', '>', '<', '>=', '<=', '!=']),
-  value: ParamOrContextParamSchema,
+  value: paramOrContextParamSchema,
 });
 
 export type ReturnValueTestProps = z.infer<typeof returnValueTestSchema>;
--- a/packages/taco/src/conditions/base/shared.ts
+++ b/packages/taco/src/conditions/base/shared.ts
@@ -19,11 +19,10 @@ export const plainStringSchema = z.string().refine(
   },
 );
 
-export const paramSchema: z.ZodSchema = z.union([
+const paramSchema = z.union([
   plainStringSchema,
   z.boolean(),
   z.number(),
-  z.lazy(() => z.array(paramSchema)),
 ]);
 
 export const paramOrContextParamSchema: z.ZodSchema = z.union([
--- a/packages/taco/src/conditions/base/shared.ts
+++ b/packages/taco/src/conditions/base/shared.ts
@@ -32,9 +32,10 @@ export const ParamSchema = z.union([
   z.array(ParamPrimitiveTypeSchema),
 ]);
 
-export const ParamOrContextParamSchema = z.union([
+export const ParamOrContextParamSchema: z.ZodSchema = z.union([
   ParamSchema,
   ContextParamSchema,
+  z.lazy(() => z.array(ParamOrContextParamSchema)),
 ]);
 
 export const returnValueTestSchema = z.object({
--- a/packages/taco/src/conditions/base/shared.ts
+++ b/packages/taco/src/conditions/base/shared.ts
@@ -3,15 +3,35 @@ import { z } from 'zod';
 import {
   CONTEXT_PARAM_REGEXP,
   ETH_ADDRESS_REGEXP,
-  USER_ADDRESS_PARAM
+  USER_ADDRESS_PARAM,
 } from '../const';
+import { CONTEXT_PARAM_PREFIX } from '../context/context';
 
 export const ContextParamSchema = z.string().regex(CONTEXT_PARAM_REGEXP);
-// Using unknown here because we don't know what the type of the parameter is
-// It could be a string, number, boolean, an array, etc.
-// We'll leave it up to the user to validate the type of the parameter.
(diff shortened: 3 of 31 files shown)


packages/taco/src/dkg.ts
```diff
@@ -161,9 +164,7 @@ export class DkgClient {
   ): Promise<DkgRitual> {
     const ritual = await DkgClient.getRitual(provider, domain, ritualId);
     if (ritual.state !== DkgRitualState.FINALIZED) {
-      throw new Error(
-        `Ritual ${ritualId} is not finalized. State: ${ritual.state}`,
-      );
+      throw new Error(ERR_RITUAL_FAILED(ritualId, ritual));
```
derekpierre, line 167: If the ritual state is not finalized, it could have failed, but it also could have not yet completed...or does something else before this call ensure that it is completed?
```suggestion
      throw new Error(ERR_RITUAL_NOT_FINALIZED(ritualId, ritual));
```

packages/taco/test/conditions/condition-expr.test.ts
```diff
@@ -47,7 +46,9 @@ describe('condition set', () => {
     method: testFunctionAbi.name,
     parameters: [USER_ADDRESS_PARAM, customParamKey],
     returnValueTest: {
-      ...testReturnValueTest,
+      index: 0,
```
derekpierre, line 49: `testFunctionAbi` has an output of `uint256` so `index` does not apply.
```suggestion
```

packages/taco/test/conditions/context.test.ts
```diff
@@ -55,10 +53,11 @@ describe('context', () => {
     const customParams: Record<string, CustomContextParam> = {};
     customParams[customParamKey] = 1234;
 
-    const contractConditionObj = {
+    const contractConditionObj: ContractConditionProps = {
       ...testContractConditionObj,
       returnValueTest: {
-        ...testReturnValueTest,
+        index: 0,
```
derekpierre, line 59: `testContractConditionObj` is an ERC20 `balanceOf` contract call that returns a `uint256`, `index` is not applicable.

```suggestion
```

packages/taco/test/conditions/context.test.ts
```diff
@@ -117,10 +119,11 @@ describe('context', () => {
     });
 
     it('detects if a signer is required', () => {
-      const conditionObj = {
+      const conditionObj: ContractConditionProps = {
         ...testContractConditionObj,
         returnValueTest: {
-          ...testReturnValueTest,
+          index: 0,
```
derekpierre, line 125: `testContractConditionObj` is an ERC20 `balanceOf` contract call that returns a `uint256`,`index` is not applicable.

```suggestion
```

Also is this the best test - it is technically an invalid condition since `USER_ADDRESS_PARAM` is not valid as a return value type since the contract returns a `uint256`. You could instead do an NFT `ownerOf` condition.

packages/taco/test/conditions/context.test.ts
```diff
@@ -145,10 +148,11 @@ describe('context', () => {
     });
 
     it('rejects on a missing signer', () => {
-      const conditionObj = {
+      const conditionObj: ContractConditionProps = {
         ...testContractConditionObj,
         returnValueTest: {
-          ...testReturnValueTest,
+          index: 0,
```
derekpierre, line 154: See prior comments about `index` not being applicable.

Also applies to other conditions after this.

packages/taco/test/test-utils.ts
```diff
@@ -235,7 +240,11 @@ export const testContractConditionObj: ContractConditionProps = {
   standardContractType: 'ERC20',
   method: 'balanceOf',
   parameters: ['0x1e988ba4692e52Bc50b375bcC8585b95c48AaD77'],
-  returnValueTest: testReturnValueTest,
+  returnValueTest: {
+    index: 0,
```
derekpierre, line 244: `index` is not applicable.

```suggestion
```

packages/taco/src/conditions/base/rpc.ts
```diff
@@ -12,11 +17,14 @@ export const rpcConditionSchema = z.object({
...
+    z.array(EthAddressOrUserAddressSchema).nonempty(),
     // Using tuple here because ordering matters
-    z.tuple([EthAddressOrUserAddressSchema, z.union([z.string(), z.number()])]),
+    z.tuple([EthAddressOrUserAddressSchema, ParamOrContextParamSchema]),
   ]),
-  returnValueTest: returnValueTestSchema,
+  returnValueTest: z.object({
+    comparator: ComparatorSchema,
+    value: ContextParamOrNumberSchema,
+  }),
```
derekpierre, line 27: This is fine for now, but just pointing out that if we add more allowed rpc methods down the road, an index value may be applicable, and would need to be added as allowable for the return value test. I believe there are some RPC calls that return lists.

piotr-roslaniec:
I dropped the changes related to return value tests

packages/taco/src/conditions/const.ts
```diff
@@ -4,6 +4,8 @@ export const USER_ADDRESS_PARAM = ':userAddress';
 
 export const ETH_ADDRESS_REGEXP = new RegExp('^0x[a-fA-F0-9]{40}$');
 
+export const CONTEXT_PARAM_REGEXP = new RegExp('^:[a-zA-Z0-9]+$');
```
cygnusv, line 7: This regex would allow stuff like `:1234` as context variables, which I guess we don't want.

My suggestion:
```suggestion
export const CONTEXT_PARAM_REGEXP = new RegExp('^:[a-zA-Z]\\w*$');
```
This restricts the first character to letters, and for the rest it uses the word character class `\w` (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_class_escape#w)

Some simple tests:
```ts
const regex = new RegExp('^:[a-zA-Z]\\w*$');

console.log(regex.test(':piotr'));
console.log(regex.test(':hey_piotr'));
console.log(regex.test(':42'));

// Output
// > true
// > true
// > false
```

packages/taco/src/conditions/context/context.ts
```diff
@@ -112,7 +112,7 @@ export class ConditionContext {
   }
 
   private isContextParameter(param: unknown): boolean {
-    return typeof param === 'string' && param.startsWith(CONTEXT_PARAM_PREFIX);
+    return String(param).startsWith(CONTEXT_PARAM_PREFIX);
```
cygnusv, line 115: How about matching with the regexp instead?

packages/taco/test/test-utils.ts
```diff
@@ -206,15 +206,15 @@ export const mockMakeSessionKey = (secret: SessionStaticSecret) => {
...
   index: 0,
   comparator: '>',
-  value: '100',
+  value: 100,
 };
 
 export const testTimeConditionObj: TimeConditionProps = {
   conditionType: TimeConditionType,
   returnValueTest: {
     index: 0,
```
cygnusv, line 215:
```suggestion
```

packages/taco/test/test-utils.ts
```diff
@@ -206,15 +206,15 @@ export const mockMakeSessionKey = (secret: SessionStaticSecret) => {
 export const testReturnValueTest: ReturnValueTestProps = {
   index: 0,
```
cygnusv, line 207:
```suggestion
```

cygnusv (changes requested):

packages/taco/src/conditions/const.ts
piotr-roslaniec: Why wouldn't we want `:1234` as a condition parameter? Or `:1abc`?

packages/taco/src/conditions/context/context.ts
piotr-roslaniec: Well, I think I'm going to remove that regexp all together, for consistency with `nucypher`. This is how a similar method is implemented in `nucypher`:
```
def is_context_variable(variable) -> bool:
    return isinstance(variable, str) and variable.startswith(_CONTEXT_PREFIX)
```

piotr-roslaniec:
[Related Gitbook change request ](https://app.gitbook.com/o/R2meumXNNad4y1B10iL7/s/WosjlL4zUGUMlcMfuSAp/~/changes/305/app-development/threshold-access-control-tac)

packages/taco/src/conditions/context/context.ts
cygnusv: > Why wouldn't we want :1234 as a condition parameter? Or :1abc? These parameters are currently accepted by nucypher (...)
> Well, I think I'm going to remove that regexp altogether, for consistency with nucypher

I think you should bring the regex back. I just wrote a test that shows that nucypher accepts all kinds of crazy context variable names:
```py
    a_context_variable_name = ":::::asdf//asdf         asdf"
    another_context_variable_name = ":12 \" ¿? \n "
```
For the sake of UX, we should keep consistency with variable names in regular coding languages (i.e., only letters, numbers and underscores, usually not starting with a number). I'm opening a PR in nucypher shortly to make it consistent with this.

piotr-roslaniec: I want to maintain consistency between `nucypher-ts` and `nucypher` to avoid weird issues down the road (the former should be downstream of the latter). But if we could address this in both places I think that would be great.

packages/taco/src/conditions/base/shared.ts
```diff
@@ -1,11 +1,19 @@
...
+
+export const ContextParamSchema = z.string().startsWith(CONTEXT_PARAM_PREFIX);
+export const ParamSchema = z.union([z.number(), z.string(), z.boolean()]);
+export const ParamOrContextParamSchema = z.union([
+  ParamSchema,
+  ContextParamSchema,
+]);
 
 export const returnValueTestSchema = z.object({
   index: z.number().optional(),
```
derekpierre, line 14: To tighten this up (i.e. ensure index >= 0), can this be done:
```suggestion
  index: z.number().nonnegative().optional(),
```

packages/taco/src/conditions/base/shared.ts
```diff
@@ -1,11 +1,19 @@
 import { z } from 'zod';
 
 import { ETH_ADDRESS_REGEXP, USER_ADDRESS_PARAM } from '../const';
+import { CONTEXT_PARAM_PREFIX } from '../context/context';
+
+export const ContextParamSchema = z.string().startsWith(CONTEXT_PARAM_PREFIX);
```
derekpierre, line 6: If using the regex, can we do:
```suggestion
export const ContextParamSchema = z.string().regex(...);
```

packages/taco/src/conditions/base/shared.ts
```diff
@@ -1,11 +1,19 @@
...
+export const ParamOrContextParamSchema = z.union([
+  ParamSchema,
+  ContextParamSchema,
+]);
 
 export const returnValueTestSchema = z.object({
   index: z.number().optional(),
   comparator: z.enum(['==', '>', '<', '>=', '<=', '!=']),
-  value: z.unknown(),
+  value: ParamOrContextParamSchema,
```
derekpierre, line 16: I don't think that `ParamSchema` defined as:
```typescript
export const ParamSchema = z.union([z.number(), z.string(), z.boolean()]);
```
covers everything here. I only gave `bool` as a single example in my comment. But truthfully, couldn't the value here be an array (of a variety of types). Basically, the example of a contract returning a struct (which is returned as tuple/list). See examples in `nucypher`.

While it is good to be as prescriptive as possible, we have to be careful about being too narrow to prevent some use cases which are still in the realm of solidity specification. I think this was the original reason why `z.unknown()` was used for a value.  In `nucypher` we use `fields.Raw` for `value` the same reason.

packages/taco/src/conditions/base/rpc.ts
```diff
@@ -12,9 +16,9 @@ export const rpcConditionSchema = z.object({
   chain: createUnionSchema(SUPPORTED_CHAIN_IDS),
   method: z.enum(['eth_getBalance', 'balanceOf']),
```
derekpierre, line 17: I don't believe `balanceOf` is an RPC call; it's an ERC20 contract call (ContractCondition should be used for that). Only `eth_getBalance` should be allowed here for now. On the `nucypher` side, only `eth_getBalance` is allowed - https://github.com/nucypher/nucypher/blob/v7.0.0/nucypher/policy/conditions/evm.py#L127

packages/taco/src/conditions/base/contract.ts
```diff
@@ -80,7 +81,7 @@ export const contractConditionSchema = rpcConditionSchema
     standardContractType: z.enum(['ERC20', 'ERC721']).optional(),
     method: z.string(),
     functionAbi: functionAbiSchema.optional(),
-    parameters: z.array(z.unknown()),
+    parameters: z.array(ParamOrContextParamSchema),
```
derekpierre, line 84: Same comment here about `ParamSchema` not covering other valid options for parameters. See comment in `shared.ts`.

Some solidity functions can take lists (of various types) as parameters as well.

packages/taco/src/conditions/context/context.ts
derekpierre: > I'm opening a PR in nucypher shortly to make it consistent with this.

@cygnusv once you open the PR do you mind if I piggyback a couple of small robustness improvements to that PR?

derekpierre: @cygnusv you are welcome to add any commits to https://github.com/nucypher/nucypher/pull/3330. Up to you.

cygnusv: Opened https://github.com/nucypher/nucypher/pull/3331. @piotr-roslaniec see that I used a more restricted regex since I suspect that `\w` behavior may different between TS and Python. In any case, the difference lies in accepting unicode characters, which I don't think it's very necessary anyway, but I'm open to suggestions (as long as we can keep complete consistency between TS and Python)

piotr-roslaniec: I will update this PR in accordance with the draft here: https://github.com/nucypher/nucypher/pull/3331

packages/taco/src/conditions/context/context.ts
```diff
@@ -112,7 +112,7 @@ export class ConditionContext {
   }
 
   private isContextParameter(param: unknown): boolean {
-    return typeof param === 'string' && param.startsWith(CONTEXT_PARAM_PREFIX);
+    return !!String(param).match(CONTEXT_PARAM_REGEXP);
```
derekpierre, line 115: I was thinking about this a bit more.

So currently the zod set up for parameters/values is:
```
export const ContextParamSchema = z.string().regex(CONTEXT_PARAM_REGEXP);
export const ParamSchema = z.unknown();
export const ParamOrContextParamSchema = z.union([
  ParamSchema,
  ContextParamSchema,
]);
```

Since this is a union that contains z.unknown() I suspect that we don't get any real validation of context parameters at the zod level of validation.

So it begs the question, whether in this method, you should do the same thing that @cygnusv does in his pending PR - https://github.com/nucypher/nucypher/pull/3331/files#diff-f364280e61c1cdd8b5991c2c34c40e0b85cd87f172f15bd7c3454dd30919fad6R79

Which is:
- check if string, and starts with "colon" (call this check (a))
    - if (a) passes and regex validates then return true
    - if (a) passes but then validating with the regex fails, then throw an error - this is basically the case of an incorrect context variable  
- if (a) fails then return false - non-context variable parameter value.

cc @cygnusv .

packages/taco/test/conditions/context.test.ts
derekpierre: Resolved other comments but this comment about the invalidity of this test still stands - the condition is invalid. 

Alternatively, you could set it up that the parameters used for `testContractConditionObj` use `:userAddress` context variable instead of a specific value (which is does currently). But the returnValueTest `value` to compare is for an int, not a user address.

packages/taco/src/conditions/context/context.ts
piotr-roslaniec: I confirmed this is an issue, i.e. `zod` is not correctly discriminating between a plain string with a `:` and a context parameter. When a string contains both `:` and an illegal context parameter character (like `&`), `zod` interprets it as a string.

I've added a more thorough validation based on your suggestions. I also made the related parameter schema less-permissive because it contributed to this issue - `z.unknown()` type matches any type and value.

packages/taco/test/conditions/context.test.ts
```diff
@@ -242,3 +243,36 @@ describe('context', () => {
...
+
+  it('accepts a boolean', () => {
+    expect(ParamOrContextParamSchema.safeParse(true).success).toBe(true);
+  });
+
+  it('accepts an array', () => {
+    expect(
+      ParamOrContextParamSchema.safeParse([1, 'hello', true]).success,
+    ).toBe(true);
+  });
```
derekpierre, line 268: Nice job on the tests. If we are going to be this prescriptive we need to be really comprehensive with our testing, and add as many cases as we can think of in terms of:
- parameters to function calls (think of passing structs etc.)
- returns from function calls since this schema is also used for the returnValueTest `value`.

Can you also add tests to try the following:

```typescript
  it('accepts nested arrays', () => {
    expect(
      ParamOrContextParamSchema.safeParse([1, "hello", [123, 456, "hello", true]]).success,
    ).toBe(true);
  });
  ```
  and this:
  
  ```typescript
  it('accepts nested arrays', () => {
    expect(
      ParamOrContextParamSchema.safeParse([1, "hello", [123, ":hello"]]).success,
    ).toBe(true);
  });
  ```

Do they pass? If they don't pass, and we go down this route - I think you may need to do a repetition of ParamSchema on itself, `lazy` I think zod called it - similar to what was done for CompoundCondition.

Comment (piotr-roslaniec) [reply packages/taco/test/conditions/context.test.ts]: - Floats are already covered by `z.number()`. One can use `z.number().int()` to further discriminate number types, etc. 
- Bytes as hex and string addresses should work just fine with `z.string()`.
- I added some tests to cover nested structures. On that note, I want to write more tests before release

File: packages/taco/src/conditions/base/shared.ts [hunk 19]
```diff
@@ -1,11 +1,47 @@
...
+export const plainStringSchema = z.string().refine(
+  (str) => {
+    if (str.startsWith(CONTEXT_PARAM_PREFIX)) {
+      return str.match(CONTEXT_PARAM_REGEXP);
+    }
+    return true;
+  },
+  {
+    message: 'Context parameters must start with ":"',
+  },
```
cygnusv, line 22: Does this mean that a string that starts with `:` and matches the context regex is considered a valid plain string? That doesn't seem right, but it's possible I don't understand the goal of this schema.

packages/taco/test/conditions/context.test.ts
derekpierre: > - Floats are already covered by z.number(). One can use z.number().int() to further discriminate number types, etc.
> - Bytes as hex and string addresses should work just fine with z.string().

Indeed. I just meant that it would be good to have explicit tests for those anyway to prevent future regressions and clarify expectations.

packages/taco/src/conditions/base/shared.ts
piotr-roslaniec: You understood correctly - This schema will accept a plain string or a condition context param, so it should have been called `plainStringOrContextParamSchema`.

This was unintended. At the time of writing, I thought I needed a negative statement to validate two different types of string, so I added that edge case. But it turns out it's much simpler, I'm going to publish an update shortly.

packages/taco/src/conditions/base/shared.ts
```diff
@@ -1,11 +1,44 @@
...
+export const paramSchema = z.union([
+  paramPrimitiveTypeSchema,
+  z.array(paramPrimitiveTypeSchema),
+]);
+
+export const paramOrContextParamSchema: z.ZodSchema = z.union([
+  paramSchema,
+  contextParamSchema,
+  z.lazy(() => z.array(paramOrContextParamSchema)),
+]);
```
derekpierre, line 36: Are we being redundant here, with `paramPrimitivetypesSchema`, `paramSchema` and `paramOrContextParamSchema`?

It's just unclear to me that we need to separate paramSchema from paramOrContextParamSchema - they feel like one in the same, and won't be used separately. Can we achieve the same thing that boils down to the following (pseudocode)? 

```
strings_that_dont_start_with_colon = z.string().refine(
(str) => {
    return !str.startsWith(CONTEXT_PARAM_PREFIX);
  },
  {
    message: "String must not be a context parameter i.e. not start with ':'",
   })

paramPrimitiveTypesSchema = union[
     strings_that_dont_start_with_colon,
     strings_that_are_context_params,
     boolean,
     number
]

paramSchema = union[
     paramPrimitiveTypesSchema
     lazy -> array(paramPrimitiveTypesSchema)
]
```

NOTE: The names used above (`strings_that_..`) is not a recommendation for name change 😅 , just simplifying the picture for myself and potentially others. The original names are fine.

packages/taco/src/conditions/base/shared.ts
```diff
@@ -1,11 +1,44 @@
...
+
+export const contextParamSchema = z.string().regex(CONTEXT_PARAM_REGEXP);
+// We want to discriminate between ContextParams and plain strings
+// If a string starts with `:`, it's a ContextParam
+export const plainStringSchema = z.string().refine(
+  (str) => {
+    return !str.startsWith(CONTEXT_PARAM_PREFIX);
+  },
+  {
+    message: 'Context parameters must start with ":"',
```
derekpierre, line 18: Should this message convey something like:

```suggestion
    message: "String must not be a context parameter i.e. not start with ':'",
```

packages/taco/src/conditions/base/shared.ts
piotr-roslaniec: Ah, yes, that's one-too-many schemas

packages/taco/src/conditions/base/rpc.ts
piotr-roslaniec: I'm going to document that with a comment, but for now I think it won't add to the ergonomics of the schema.

packages/taco/test/conditions/context.test.ts
piotr-roslaniec: So there are comparators that don't make sense for certain parameter types? I would think we can '>' compare strings since strings have an (alphanumeric) ordering. But we can also constrain the type of comparator on the parameter type.

packages/taco/src/conditions/base/shared.ts
```diff
@@ -1,11 +1,41 @@
...
+export const plainStringSchema = z.string().refine(
+  (str) => {
+    return !str.startsWith(CONTEXT_PARAM_PREFIX);
+  },
+  {
+    message: 'String must not be a context parameter i.e. not start with ":"',
+  },
+);
+
+export const paramSchema: z.ZodSchema = z.union([
```
derekpierre, line 22: Is `paramSchema` ever used externally - wondering if it doesn't need to be exported.

packages/taco/test/conditions/context.test.ts
derekpierre: It's not the parameter types, but there are return values that don't make sense based on what the contract function is expected to return. The `testContractConditionObj`'s method is ERC20.balanceOf which is supposed to return a `uint256`. However, the returnValueTest is being modify to use an `address` as the comparator value. So the condition is attempting to compare a `unit256` to an `address` which is confusing.

packages/taco/src/conditions/base/rpc.ts
```diff
@@ -3,20 +3,24 @@ import { z } from 'zod';
...
+  method: z.enum(['eth_getBalance']),
   parameters: z.union([
-    z.array(EthAddressOrUserAddressSchema).length(1),
+    z.array(EthAddressOrUserAddressSchema).nonempty(),
     // Using tuple here because ordering matters
-    z.tuple([EthAddressOrUserAddressSchema, z.union([z.string(), z.number()])]),
+    z.tuple([EthAddressOrUserAddressSchema, paramOrContextParamSchema]),
   ]),
-  returnValueTest: returnValueTestSchema,
+  returnValueTest: returnValueTestSchema, // Update to allow multiple return values after expanding supported methods
```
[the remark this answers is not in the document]
piotr-roslaniec: Hmmm no, it will only be updated once we update the `method` field in the RPC schema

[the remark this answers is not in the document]
derekpierre: I thought the `returnValueTestSchema` is from here - https://github.com/piotr-roslaniec/nucypher-ts/blob/fix-lynx/packages/taco/src/conditions/base/shared.ts#L37? In which case tuples of values etc. are allowed (same as parameters). Or am I mistaken?

packages/taco/test/conditions/context.test.ts
```diff
@@ -122,10 +123,11 @@ describe('context', () => {
...
       const conditionObj = {
         ...testContractConditionObj,
+        parameters: [USER_ADDRESS_PARAM],
         returnValueTest: {
-          ...testReturnValueTest,
-          value: USER_ADDRESS_PARAM,
-        },
+          comparator: '==',
+          value: 100,
+        } as ReturnValueTestProps,
```
derekpierre, line 130: Should this test could be broken into 2?
1) detects if signer is required because context var only in parameters
2) detects if signer is required because content vari only in returnValueTest

packages/taco/src/conditions/base/rpc.ts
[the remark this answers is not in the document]
piotr-roslaniec: Ah no, you're right. It is covered, but I was hoping we could constrain it instead of relaxing it (as we did in the latest change) when we have more information about the available methods.

theref (approved):

packages/taco/test/conditions/context.test.ts
```diff
@@ -216,24 +238,123 @@ describe('context', () => {
...
+
+  it('accepts an integer', () => {
+    expect(paramOrContextParamSchema.safeParse(123).success).toBe(true);
+  });
+
+  it('accepts an floating number', () => {
+    expect(paramOrContextParamSchema.safeParse(123.4).success).toBe(true);
+  });
+
+  it('accepts a hex string', () => {
```
derekpierre, line 283:
```suggestion
  it('accepts a string', () => {
```

packages/taco/test/conditions/context.test.ts
```diff
@@ -116,13 +120,31 @@ describe('context', () => {
...
+      expect(() => conditionExpr.buildContext(provider, {})).toThrow(
+        `Signer required to satisfy ${USER_ADDRESS_PARAM} context variable in condition`,
+      );
+    });
+
+    it('detects when signer is required by return value test', () => {
+      const conditionObj = {
+        ...testContractConditionObj,
+        returnValueTest: {
+          comparator: '==',
```
derekpierre, line 145: Can we do something like the following to have the user address comparator value make sense, since `testContractCOnditionObj` is a ERC20.balanceOf condition that returns an int? 

```suggestion
      const conditionObj = {
        ...testContractConditionObj,
        standardContractType: 'ERC721',
        method: 'ownerOf',
        parameters: [3591],
        returnValueTest: {
          comparator: '==',
```

derekpierre (commented):
Couple minor comments - almost there 👍

packages/taco/src/conditions/base/rpc.ts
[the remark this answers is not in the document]
derekpierre: You can, I guess, if you have rpcConditionSchema define it's own returnValueSchema...?

derekpierre (approved):
🎸 

Thanks for completing such a heavy PR 👏 .

PR #386, opened 2023-11-07
Remove redundant exports from `taco` package

piotr-roslaniec:
**Type of PR:**
- Other

**Required reviews:**
- 3

**What this does:**
- Removes redundant (duplicated) exports from the `taco` package
- Moves `CompoundCondition` into `condition.base` (already covered in GitBook)
- Reorganizes conditions into `rpc`, `contract`, and `time` modules

**Why it's needed:**
- Causes duplicated symbols in generated docs


packages/taco/src/conditions/index.ts
```diff
@@ -1,9 +1,6 @@
 import * as base from './base';
 import * as predefined from './predefined';
 
-export * from './base';
-export * from './predefined';
```
piotr-roslaniec, line 5: The substance of this PR

manumonti (approved):
LGTM! ✌️

demos/taco-demo/src/ConditionBuilder.tsx
```diff
@@ -8,7 +8,7 @@ interface Props {
   enabled: boolean;
 }
 
-const rpcCondition = new conditions.RpcCondition({
+const rpcCondition = new conditions.base.RpcCondition({
```
derekpierre, line 11: Do we have to use `base` for the exported package? I like how it looked before i.e. `conditions.RpcCondition`.

piotr-roslaniec: We could skip `base` as a submodule and place all exports in `conditions`. We can keep `predefined` as a submodule: `conditions.predefined.ERC721Ownership`, etc.

I guess not having `base` is handy in some cases, but I'm not sure if it's not too messy, I'll look into it.

packages/taco/test/conditions/compound-condition.test.ts
```diff
@@ -185,9 +196,9 @@ describe('validation', () => {
   ])('accepts shorthand for "%s" operator', (operator, operands, expected) => {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const compoundConditionShorthands: Record<string, any> = {
-      'and': CompoundCondition.and,
-      'or': CompoundCondition.or,
-      'not': CompoundCondition.not,
+      and: CompoundCondition.and,
+      or: CompoundCondition.or,
+      not: CompoundCondition.not,
```
cygnusv, line 201: How does removing the quotes work here?

demos/taco-demo/src/ConditionBuilder.tsx
cygnusv: Agree with @derekpierre, but interested to hear why do you think it can be messy.

piotr-roslaniec: If we remove the `conditions.base` namespace, it's going to roughly triple the size of the `conditions` namespace.

Perhaps it's not a big deal, but it's a detour from what I imagined the organization of these concepts to look like. Any thoughts? Do you think it's manageable?

image

packages/taco/test/conditions/compound-condition.test.ts
cygnusv: Huh...I still don't get it. How would it work if your key were the string `"foo: bar"`?
(In any case, it's not important for this PR, just wanting to understand more. TS/JS can be weird.)

piotr-roslaniec: These are equivalent
```
const a = { my_key: 1 };
const b = { "my_key": 1 };
// a == b
```

This however, will not compile:
```
const a = { foo: bar: 1 }; // whoops, throws compilation error
const b = { "foo: bar": 1 }; // works, amazingly
```

We can use `"string notation"` to use actual strings as object keys. This works similarly how you can declare a dict in Python.

The second notation is the object notation. The keys are actually object properties so they have to follow the syntax rules for objects.

demos/taco-demo/src/ConditionBuilder.tsx
derekpierre: It is a bit all-encompassing. Perhaps instead of lumping many items into one submodule (`conditions` or `base`), would it make sense / be possible to have more submodules, something like:

- conditions (module)
    - conditions.time (sub-module)
        - TimeCondition
        - TimeConditionProps
        - TimeConditionMethod
        - TimeConditionType
    - conditions.rpc (sub-module)
        ...
    - conditions.contract (sub-module)
        ...
    - conditions.compound (sub-module)
        ...
    - conditions.predefined (sub-module)
        ...
    - condition.context (sub-module)
        - ConditionContext
        - CustomContextParam
    - Condition
    - ConditionExpression
    - ConditionExpressionJSON
    - ConditionProps


Perhaps `context` does not need its own sub-module - your judgement would be better with the organization than mine would be.

piotr-roslaniec: It looks neat, but I don't know if it's useful from the perspective of the API user. `base` conditions and `predefined` conditions (we could also rename them to `templates` or something) are useful because they indicate how to use these classes (basic building blocks vs concrete blocks. So maybe something like `conditions.base.rpc` would work if we want to tidy up `base` a bit, but I don't think replacing `conditions.base` with `conditions.rpc` is that much better.

At any rate, I think these changes are cosmetic at best.

derekpierre: Indeed this discussion is a cosmetic change, but also a broader question of API clarity based on the initial concern of a module having too many items, which I agree with.

 I think "useful" can boil down to many factors that include:
1. A concern about having too many items in `conditions`, but it looks like there are many items in `conditions.base` anyway
2. Does separating the items make the API cleaner and easier to understand, and what does that separation look like.

> base conditions and predefined conditions (we could also rename them to templates or something) are useful because they indicate how to use these classes (basic building blocks vs concrete blocks

Perhaps to your point, `conditions.predefined` in the layout I used may be less clear and perhaps it should be more descriptive, so something like `conditions.predefined.nft`/`conditions.predefined.erc721`, `conditions.predefined.token`/`conditions.predefined.erc20`, and that would be clearer to users? wdyt?

> So maybe something like conditions.base.rpc would work if we want to tidy up base a bit, but I don't think replacing conditions.base with conditions.rpc is that much better.

Let's go down this route, so would something like the following be more appealing?

- conditions
    - base 
        - time
            - TimeCondition
            - TimeConditionProps
            - TimeConditionMethod
            - TimeConditionType
        - rpc
           ...
        - contract
          ...
    - predefined
       - nft
          ...
       - token   
          ...
    - compound
       ...
    - context
        - ConditionContext
        - CustomContextParam
    - Condition
    - ConditionExpression
    - ConditionExpressionJSON
    - ConditionProps

piotr-roslaniec: It looks fine, but I have an issue with refactoring some code right now. There is a circular dependency ([see here](https://github.com/nucypher/taco-web/pull/386/files#diff-799f1bc9e953af5f70d8c658a3d78b8cd507da71930a62f6f506fc928eca667aL22-L23) and I'm not sure how to resolve it yet. So if I were to apply these changes it would be sort of inconsistent. I've pushed these changes, we can revert them if anything.

derekpierre: I see. I think the problem stems from the use of `ConditionFactory` within `Condition` since `ConditionFactory` needs to know about all of the various condition types i.e. import them.

Here is something I played around with to rectify that which is based over this PR's branch. I'm not claiming that it is complete, but the tests pass (`pnpm test`) - https://github.com/derekpierre/taco-web/commit/ff414b89b932cff7af3e10cc5341f4cb14c8d9a3.

Perhaps that can be helpful to you and built upon?

piotr-roslaniec: Thanks @derekpierre, I included these changes in the newest commit

theref (approved):

PR #396, opened 2023-11-13
Add test cases for conditions

piotr-roslaniec:
**Type of PR:**
-  Other

**Required reviews:**
- 1

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**
- Based on top of #379 
- Ideas for testing cases welcome

### DIFF
--- a/packages/taco/test/conditions/base/rpc.test.ts
+++ b/packages/taco/test/conditions/base/rpc.test.ts
@@ -94,5 +94,41 @@ describe('validation', () => {
         },
       });
     });
+
+    it('rejects empty parameters', () => {
+      const rpcObj = {
+        ...testRpcConditionObj,
+        parameters: [],
+      };
+
+      const result = RpcCondition.validate(rpcConditionSchema, rpcObj);
+      expect(result.error).toBeDefined();
+      expect(result.data).toBeUndefined();
+      expect(result.error?.format()).toMatchObject({
+        parameters: {
+          _errors: ['Array must contain at least 1 element(s)'],
+        },
+      });
+    });
+
+    it('rejects non-array parameters', () => {
+      const badRpcObj = {
+        ...testRpcConditionObj,
+        parameters: 'not an array',
+      };
+
+      const result = RpcCondition.validate(rpcConditionSchema, badRpcObj);
+      expect(result.error).toBeDefined();
+      expect(result.data).toBeUndefined();
+      expect(result.error?.format()).toMatchObject({
+        parameters: {
+          _errors: [
+            // Expecting two errors here because of the nested array-tuple in the parameters schema
+            'Expected array, received string',
+            'Expected array, received string',
+          ],
+        },
+      });
+    });
   });
 });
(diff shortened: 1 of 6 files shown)


packages/taco/test/conditions/base/rpc.test.ts
```diff
@@ -94,5 +94,41 @@ describe('validation', () => {
         },
       });
     });
+
+    it('rejects empty parameters', () => {
```
derekpierre, line 98: This is fine for now, but something to note/update as we expand to allow other rpc methods.

packages/taco/test/conditions/compound-condition.test.ts
```diff
@@ -216,4 +216,36 @@ describe('validation', () => {
...
+
+  it('rejects non-array operands for "and" and "or" operators', () => {
+    const badObj = {
+      operator: 'and',
+      operands: testContractConditionObj,
+    };
+    expect(() => new CompoundCondition(badObj)).toThrow();
+  });
+
+  it('rejects array operands for "not" operator', () => {
```
derekpierre, line 244: To clarify, the following still holds, correct?
- `not`, `and`, and `or` all accept array operands, just that `not` requires the array to be size 1, and `and`/`or` expect array size to be > 1.
- They all should fail if "operands" is not an array.

packages/taco/test/conditions/context.test.ts
```diff
@@ -262,15 +262,21 @@ describe('param or context param schema', () => {
   });
 
   it('accepts a hex string', () => {
-    expect(paramOrContextParamSchema.safeParse("deadbeef").success).toBe(true);
+    expect(paramOrContextParamSchema.safeParse('deadbeef').success).toBe(true);
```
derekpierre, line 265: Careful here - any hex string as a representation of bytes should be prefixed by `0x`. Strings that are just strings don't need the `0x` prefix.

packages/taco/test/conditions/compound-condition.test.ts
piotr-roslaniec: I've added some tests to cover these cases

derekpierre: 👍

packages/taco/test/conditions/compound-condition.test.ts
```diff
@@ -205,4 +216,67 @@ describe('validation', () => {
...
+
+  it('rejects invalid operator', () => {
+    const badObj = {
+      operator: 'invalid-operator',
+      operands: [testContractConditionObj, testTimeConditionObj],
+    };
+    expect(() => new CompoundCondition(badObj)).toThrow();
+  });
+
+  it.each(['or', 'and'])(
```
derekpierre, line 228: Should `not` also be in here - it should reject empty operands as well?
```suggestion
  it.each(['or', 'and', 'not'])(
```

packages/taco/test/conditions/compound-condition.test.ts
```diff
@@ -205,4 +216,67 @@ describe('validation', () => {
...
+    (operator) => {
+      const badObj = {
+        operator,
+        operands: [],
+      };
+      expect(() => new CompoundCondition(badObj)).toThrow();
+    },
+  );
+
+  it.each(['or', 'and'])(
```
derekpierre, line 239: Should `not` be in here as well - it should reject non-array operands?
```suggestion
  it.each(['or', 'and', 'not'])(
```

manumonti (approved):
LGTM!

PR #398, opened 2023-11-15
Update DKG states based on last minute changes to Coordinator contract.

derekpierre:
**Type of PR:**

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

Updates `taco-web` to incorporate latest changes made via https://github.com/nucypher/nucypher-contracts/pull/187.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -42,7 +42,7 @@
   "dependencies": {
     "@ethersproject/abi": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
-    "@nucypher/nucypher-contracts": "0.10.1",
+    "@nucypher/nucypher-contracts": "0.12.0",
     "@nucypher/nucypher-core": "0.13.0-alpha.1",
     "axios": "^1.6.0",
     "deep-equal": "^2.2.1",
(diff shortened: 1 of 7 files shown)


cygnusv (approved):

piotr-roslaniec (approved):

jMyles (approved):
Simple and effective.

My only pointer here is that the various Ritual states represent a much broader view of the network than makes sense to encapsulate purely in this repo, or even purely as part of TACo.  These states apply across our codebases, and also to signing cohorts like the one used by tBTC.

Also: I wonder, why are some states prepended with "DKG_" ? For example, is there a proper state of invalidity (other than, obviously, expired) other than something we'd regard as invalidity of the DKG?  Why was "INVALID" insufficient?

derekpierre:
> Also: I wonder, why are some states prepended with "DKG_" ? For example, is there a proper state of invalidity (other than, obviously, expired) other than something we'd regard as invalidity of the DKG? Why was "INVALID" insufficient?

These were mimicked from the contract changes - https://github.com/nucypher/nucypher-contracts/pull/187.

I believe the thinking here was that some states are part of the DKG Ritual process itself (transcripts, aggregations), but once completed successfully (operative word being successful), then using `DKG_` had less meaning because it was no longer really a DKG, but rather had moved passed the DKG process. (perhaps @cygnusv had other thoughts)

PR #402, opened 2023-11-17
Update peer & core dependencies config

### LINKED ISSUE #400: Do we need @nucypher/nucypher-core dependency on taco and pre packages?
It's already a dependency in `shared`, which is a dependency on both `taco` and `pre`.

### LINKED ISSUE #401: Missing peer dependency `@nucypher/core`
Getting this warning when installing `"@nucypher/taco": "0.1.0-rc.6"`. I think potentially an issue with how the packages are configured in the monorepo

```bash
packages/api/core
└─┬ @nucypher/taco 0.1.0-rc.6
  └── ✕ missing peer @nucypher/shared@"*"
Peer dependencies that should be installed:
  @nucypher/shared@"*" 
  ```

### PR DESCRIPTION
**Pull request by piotr-roslaniec**
**Type of PR:**
- Other

**Required reviews:**
- 1

**What this does:**

**Issues fixed/closed:**
- Fixes #401

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/demos/taco-demo/package.json
+++ b/demos/taco-demo/package.json
@@ -12,7 +12,7 @@
     "type-check": "tsc --noEmit"
   },
   "dependencies": {
-    "@nucypher/taco": "0.1.0-rc.6",
+    "@nucypher/taco": "^0.1.0-rc.6",
     "@nucypher/shared": "0.1.0-rc.4",
     "@usedapp/core": "^1.2.13",
     "buffer": "^6.0.3",
--- a/demos/taco-demo/package.json
+++ b/demos/taco-demo/package.json
@@ -13,7 +13,7 @@
   },
   "dependencies": {
     "@nucypher/taco": "^0.1.0-rc.6",
-    "@nucypher/shared": "0.1.0-rc.4",
+    "@nucypher/shared": "^0.1.0-rc.4",
     "@usedapp/core": "^1.2.13",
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: packages/taco/test/conditions/compound-condition.test.ts [hunk 1]
```diff
@@ -185,9 +196,9 @@ describe('validation', () => {
   ])('accepts shorthand for "%s" operator', (operator, operands, expected) => {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const compoundConditionShorthands: Record<string, any> = {
-      'and': CompoundCondition.and,
-      'or': CompoundCondition.or,
-      'not': CompoundCondition.not,
+      and: CompoundCondition.and,
+      or: CompoundCondition.or,
+      not: CompoundCondition.not,
```
cygnusv, line 201: When was this block changed to using quotes? I remember there was a recent change, in which I commented, with the same diff (i.e., removing the quotes). Is it possible that there's conflicting linting configurations somewhere?

piotr-roslaniec: I think that's the case, but I wasn't able to reproduce that. I will keep an eye on that. The proper way to format is to use `pnpm lint:fix` and that should be consistent with the CI check, `pnpm ci:lint`.

theref (approved):

PR #416, opened 2023-11-23
Force dependency resolution for `@nucypher/nucypher-core`

### LINKED ISSUE #400: Do we need @nucypher/nucypher-core dependency on taco and pre packages?

piotr-roslaniec:
**Type of PR:**
- Other

**Required reviews:**
- 1

**Issues fixed/closed:**
- Fixes #400


package.json
```diff
@@ -81,7 +81,8 @@
       "node-forge@<1.0.0": ">=1.0.0",
       "node-forge@<1.3.0": ">=1.3.0",
       "glob-parent@<5.1.2": ">=5.1.2",
-      "nth-check@<2.0.1": ">=2.0.1"
+      "nth-check@<2.0.1": ">=2.0.1",
+      "@nucypher/nucypher-core": "0.13.0-alpha.1"
```
derekpierre, line 85: How does this statement work? Is one version used for dev and the newest version on npm used otherwise (versions defined as "*")? OR does this version dictate the version used across the packages? or ... ?

jMyles: I'm also a little confused by this spec.

In the documentation for `overrides`, it says, "You may not set an override for a package that you directly depend on unless both the dependency and the override itself share the exact same spec. "

So how does this work?

https://docs.npmjs.com/cli/v10/configuring-npm/package-json#overrides

derekpierre: Cool. So basically this is the single definition of the `nucypher-core` version being used across all packages.

derekpierre (approved):
🎸

PR #417, opened 2023-11-23
refactor: taco nodejs example

alimahdiyar:
**Type of PR:**

- Documentation

**Required reviews:**

- 1

**What this does:**
Adds a more clear and understandable example code for Taco by separating the encrypt and decrypt functions

**Why it's needed:**
It was not clear in the example how to create an encrypted message, send it somewhere else, and decrypt it later. The decryption code parts were using objects from encryption. So I separated encrypt and decrypt functions to show the users how they can be separately used.


theref (approved):

piotr-roslaniec:
Thanks @alimahdiyar

PR #462, opened 2024-01-11
Update `tapir` ritual

piotr-roslaniec:
**Type of PR:**
- Other

**Required reviews:**
- 1

**What this does:**
- Updates `tapir` ritual to use id `0`
- Updates contract registry using new `@nucypher/nucypher-contracts:0.16.0` version
- There seems to be some issue with [Infura project settings](https://github.com/nucypher/taco-web/actions/runs/7492767616/job/20397015512?pr=462#step:7:22). It works fine on the `lynx` job, but fails on the `tapir` job. I updated `tapir` job to use a public RPC instead.

### DIFF
--- a/.github/workflows/tapir.yml
+++ b/.github/workflows/tapir.yml
@@ -10,10 +10,10 @@ on:
 
 # TODO: Use variables when GH supports it for forks. See https://github.com/orgs/community/discussions/44322
 env:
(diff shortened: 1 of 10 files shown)


KPrasch (approved):
🤠

derekpierre (approved):
🎸

PR #463, opened 2024-01-11
Release `taco@0.2.x`

piotr-roslaniec:
**Type of PR:**
- Other

**Required reviews:**
- 1

**Why it's needed:**
- One of the underlying dependencies, `@nucypher/nucypher-contracts` was updated
- Necessary to keep `taco-web` working for `tapir` users

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
       "node-forge@<1.3.0": ">=1.3.0",
       "glob-parent@<5.1.2": ">=5.1.2",
       "nth-check@<2.0.1": ">=2.0.1",
-      "@nucypher/nucypher-core": "0.13.0-alpha.1"
+      "@nucypher/nucypher-core": "^0.14.1"
     }
   }
 }
--- a/packages/taco/package.json
+++ b/packages/taco/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@nucypher/taco",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "keywords": [
     "taco",
     "threshold",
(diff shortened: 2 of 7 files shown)


manumonti (approved):
LGTM 👍

package.json
```diff
@@ -82,7 +82,7 @@
       "node-forge@<1.3.0": ">=1.3.0",
       "glob-parent@<5.1.2": ">=5.1.2",
       "nth-check@<2.0.1": ">=2.0.1",
-      "@nucypher/nucypher-core": "0.13.0-alpha.1"
+      "@nucypher/nucypher-core": "^0.14.1"
```
derekpierre, line 85: Judging by https://github.com/nucypher/nucypher-core/pull/89 there are no concerns about `0.14.1` not being compatible with nodes running `0.13.0` correct? The breaking changes were limited to the wasm bindings, which the nodes don't care about.

packages/taco/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nucypher/taco",
-  "version": "0.1.0",
+  "version": "0.2.2",
```
derekpierre, line 3: Why 0.2.2 and not 0.2.0?

package.json
piotr-roslaniec: Yes, there changes are strictly for `nucypher-core-wasm` and there is no impact on Python users.

packages/taco/package.json
piotr-roslaniec: I messed up and published a bad build. I yanked previous versions, but `npmjs.com` doesn't allow republishing so I had to bump the version.

PR #492, opened 2024-02-21
Update `@nucypher/nucypher-contracts`

### LINKED ISSUE #485: Don't get participants from Coordinator with transcripts
Use updated `getParticipants(ritual_id, startIndex, maxResults, transcripts)` contract call to obtain participants for a ritual. Transcripts (the flag to obtain transcripts or not) should be False for just decryption, although in the future if doing local verification you may want it to be True. Note that if transcripts is ever True, eg. for local verification, you will **NEED** to do pagination. If transcripts are False, then you done "need" pagination. 

This functionality is available on `lynx` but not yet `mainnet` - it will require a mainnet contract upgrade of the Coordinator.

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:**
- 1

**Issues fixed/closed:**
- Fixes #485

### DIFF
--- a/packages/taco/src/taco.ts
+++ b/packages/taco/src/taco.ts
@@ -159,6 +159,7 @@ export const decrypt = async (
     porterUri,
     messageKit,
     ritualId,
+    ritual.sharesNum,
     ritual.threshold,
     signer,
     customParameters,
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -42,7 +42,7 @@
   "dependencies": {
     "@ethersproject/abi": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
-    "@nucypher/nucypher-contracts": "^0.16.0",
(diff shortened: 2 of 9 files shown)


piotr-roslaniec:
Will be fixed by releasing this PR: https://github.com/nucypher/nucypher-contracts/pull/242

KPrasch (approved):

packages/taco/src/taco.ts
```diff
@@ -159,6 +159,7 @@ export const decrypt = async (
     porterUri,
     messageKit,
     ritualId,
+    ritual.sharesNum,
     ritual.threshold,
```
derekpierre, line 163: Any reason we don't want to simply pass the `ritual` object as a single parameter instead of these 3 individual data values which it already contains?

piotr-roslaniec: No specific reason, I might consider it

PR #493, opened 2024-02-21
Add  a `mainnet` test job to ci

piotr-roslaniec:
**Type of PR:**
-  Other

**Required reviews:**
- 1

**What this does:**
- Adds another cron job for testing TACo on `mainnet`, similar to existing `tapir` and `lynx` jobs

**Why it's needed:**
- We don't monitor compatibility of TACo and `mainnet` 

**Notes for reviewers:**
- The new `mainnet` job is failing because I've created this PR from my fork, where access to repository secrets is not possible

### DIFF
--- a/.github/workflows/mainnet.yml
+++ b/.github/workflows/mainnet.yml
@@ -1,0 +1,56 @@
+name: '🔎 Mainnet Example'
+
+on:
+  schedule:
+    - cron: "0 * * * *"  # Every hour
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+# TODO: Use variables when GH supports it for forks. See https://github.com/orgs/community/discussions/44322
+env:
+  RPC_PROVIDER_URL: "https://mainnet.infura.io/v3/3747007a284045d483c342fb39889a30"
+  ENCRYPTOR_PRIVATE_KEY: ${{ secrets.MAINNET_ENCRYPTOR_PRIVATE_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/mainnet.yml
```diff
@@ -0,0 +1,56 @@
...
+  schedule:
+    - cron: "0 * * * *"  # Every hour
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+# TODO: Use variables when GH supports it for forks. See https://github.com/orgs/community/discussions/44322
+env:
+  RPC_PROVIDER_URL: "https://mainnet.infura.io/v3/3747007a284045d483c342fb39889a30"
```
manumonti, line 13: Maybe this was discussed in the past, but can't we use secrets for this RPC endpoint also? What if someone takes this API key and uses it until the free requests are run out?

manumonti (approved):
LGTM! 👌 I just left a thought.

.github/workflows/mainnet.yml
piotr-roslaniec: I agree this is problematic, but so far we have gotten away with this. If move this variable to `secrets`, it's going to break these checks for fork-PR workflow we are currently using.

We can always switch to public RPCs if this becomes an issue.

derekpierre: Just as a side note: this endpoint restricts the contract that calls that can be made to.

PR #508, opened 2024-04-09
Deprecate Mumbai on `tapir`

### LINKED ISSUE #506: Deprecate Polygon Mumbai
- Currently used in examples and demos with the `tapir` domain

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:**
-  1

**Issues fixed/closed:**
- Fixes #506

**Notes for reviewers:**
- Mainnet job is failing for an unrelated reason (expired ritual)

### DIFF
--- a/demos/taco-demo/src/App.tsx
+++ b/demos/taco-demo/src/App.tsx
@@ -7,7 +7,7 @@ import {
   initialize,
   ThresholdMessageKit,
 } from '@nucypher/taco';
-import { Mumbai, useEthers } from '@usedapp/core';
+import { useEthers } from '@usedapp/core';
 import { ethers } from 'ethers';
 import React, { useEffect, useState } from 'react';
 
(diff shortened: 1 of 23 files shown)


manumonti (approved):

demos/taco-demo/src/App.tsx
```diff
@@ -32,7 +32,7 @@ export default function App() {
 
   useEffect(() => {
     initialize();
-    switchNetwork(Mumbai.chainId);
+    switchNetwork(80002);
```
derekpierre, line 35: I guess there may not be support for an 'Amoy' object just yet?

piotr-roslaniec: Yes, we need to bump a bunch of deps that may expose this object but it's going to cause a mess. Hence going to address it at another time.

derekpierre: 👍

PR #510, opened 2024-04-13
Test npm package availability on CI

piotr-roslaniec:
**Type of PR:**
- Other

**Required reviews:**
- 1

**What this does:**
- Checks whether `@nucypher/taco` has a correct build available under expected npm tags

**Why it's needed:**
- To catch bad builds and tagging issues

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -1,0 +1,28 @@
+name: 'Test npmjs.com package'
+
+on:
+  schedule:
+    - cron: "0 * * * *"  # Every hour
+  workflow_dispatch:
+
+jobs:
+  npm:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [ '18.x' ]
+        version: [ 'latest', 'mainnet', 'devnet', 'testnet' ]
+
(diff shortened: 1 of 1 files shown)


manumonti (approved):
The code looks good to me, but I have a question:

Why to check every hour? I'm assuming that the main reason for packages not being available is some problem during the build and publish process. So, checking the package after it is published should be enough.

What do you think?

piotr-roslaniec:
We are publishing packages manually (Codechecks will work great after we release `1.0.0`: https://github.com/nucypher/taco-web/pull/460). So having to run this job is just one more thing we need to remember to do.

NPM tags can also change without publishing a new version. For example, a change on `lynx` followed by a change on `tapir` will require us to release a new version for `lynx` and then tag it again for `tapir`.

It doesn't make a huge difference to me whether this job runs every hour or every 24 hours. I think the sooner we catch this the better. If we do this every 24 hours then we spend 24 less compute, but I don't think compute spent is an issue.

manumonti:
Got it, thanks! it makes more sense to me now.

derekpierre (approved):
🎸

PR #517, opened 2024-05-02
Remove instruction to add wallet to allow list

theref:
**Type of PR:**
- Documentation

We've had a couple of messages at hackathons where people ask to have their wallet added to allow list. This is no longer required because ritual id 0 is fully open

### DIFF
--- a/demos/taco-demo/src/App.tsx
+++ b/demos/taco-demo/src/App.tsx
@@ -123,8 +123,7 @@ export default function App() {
       </p>
       <p>
         Connect with us on our{' '}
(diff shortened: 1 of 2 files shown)


piotr-roslaniec:
I think we should replace this with:
1) A link to TACo [docs](https://docs.threshold.network/app-development/threshold-access-control-tac/integration-guide/get-started-with-tac#testnet-configuration), with a short sentence explaining that the access is open on testnet
2) A link to our Discord, with explanation that the access for production environment is gated and requires an allowlist configuration and other steps

piotr-roslaniec (approved):

cygnusv (approved):

PR #529, opened 2024-06-12
Allow reusing SIWE messages in EIP4361-based condition authentication

### LINKED ISSUE #528: Add single sign-on capabilities to EIP4361 support
Related to #526 .
Needed as part of #527 .

Instead of prompting for an additional signature for the decryptor account, the original sign-in used for the app should be allowed to be reused.

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:**
- 2

**What this does:**
- Allows to reuse existing SIWE messages by providing them as a context parameter at decryption time

**Issues fixed/closed:**
- Fixes #528

**Why it's needed:**

**Notes for reviewers:**


piotr-roslaniec:
Sending review requests for an early review of this draft

packages/taco/test/conditions/context.test.ts
```diff
@@ -392,23 +430,64 @@ describe('authentication provider', () => {
...
+
+    // Now, we expect that the auth signature will be available in the context variables
+    // BUT we want it to be available under the USER_ADDRESS_PARAM_EIP4361 key
+    const authSignature = contextVars[
+      USER_ADDRESS_PARAM_EIP4361
+    ] as AuthSignature;
+    expect(authSignature).toBeDefined();
+    await testEIP4361AuthSignature(authSignature);
+
+    // TODO: Should we enforce that? It does not affect the server-side verification of the signature
```
manumonti, line 489: Sorry, it is not clear to me: enforce what?

manumonti (approved):
This looks good to me, not sure how much more work is needed, but it seems pretty complete.

piotr-roslaniec:
Asking for review despite some rough edges to confirm the design

packages/taco/src/conditions/context/context.ts
```diff
@@ -133,43 +116,67 @@ export class ConditionContext {
...
+    // We expect to find it in the custom parameters
+    if (requestedParameters.has(USER_ADDRESS_PARAM_EXTERNAL_EIP4361)) {
+      if (!this.customParameters[USER_ADDRESS_PARAM_EXTERNAL_EIP4361]) {
+        throw new Error(
+          ERR_MISSING_CONTEXT_PARAMS([USER_ADDRESS_PARAM_EXTERNAL_EIP4361]),
+        );
+      }
+      // Notice that we are using the same value for both EIP4361 and external EIP4361
+      parameters[USER_ADDRESS_PARAM_EIP4361] =
+        this.customParameters[USER_ADDRESS_PARAM_EXTERNAL_EIP4361];
```
derekpierre, line 170: What's the reason for this? On L148-L151 we check to ensure that both entries are not in concurrently in the requested parameters.

derekpierre (commented):
Based on the changes in #531 , feels like this PR would look a lot different post-#531. And if merged pre-#531 it would create quite a bit of merge conflicts for #531. Would basing over #531 make that simpler.

We also don't address how to help the user "easily" generates the `AuthSignature` .
It would be nice to have to a function/API (seemingly an `AuthProvider`) that helps with generating the `AuthSignature` format we want for external SIWE? Or is that planned for a follow-up? wdyt?

packages/taco/src/conditions/context/context.ts
derekpierre: See https://github.com/nucypher/nucypher/pull/3513.

piotr-roslaniec: Sure, we can address this on the server-side. This was a workaround for (at the time) not having this support.

piotr-roslaniec:
@derekpierre I think we can rebase this over #531.

I added a public method to the EIP4361 auth provider to expose message-creating logic to end-users.

piotr-roslaniec:
Rebased over https://github.com/nucypher/taco-web/pull/531

packages/taco-auth/src/providers/eip4361.ts
```diff
@@ -35,10 +35,17 @@ export class EIP4361AuthProvider {
...
     const address = await this.signer.getAddress();
+    const chainId = (await this.provider.getNetwork()).chainId;
     const { domain, uri } = this.getParametersOrDefault();
+    const message = EIP4361AuthProvider.prepareSiweMessage(address, chainId, domain, uri);
+    const signature = await this.signer.signMessage(message);
+    const scheme = 'EIP4361';
+    return { signature, address, scheme, typedData: message };
+  }
+
+  public static prepareSiweMessage(address: string, chainId: number, domain: string, uri: string) {
```
derekpierre, line 46: Is this static method related to the TODO on L61?

I might be missing the use case for this method? Just a helper for the user to create a SIWE message to then use for the context variable? If so, it seems pretty restrictive since there are more values that SIWE allows apps to customize.

In terms of our default EIP4361 auth provider behaviour, there is a question about what values to use for `domain` and `uri`, and whether the user can customize that, so wondering if this is related?

packages/taco/src/conditions/const.ts
```diff
@@ -34,4 +34,6 @@ export const RESERVED_CONTEXT_PARAMS = [
   USER_ADDRESS_PARAM_DEFAULT,
   USER_ADDRESS_PARAM_EIP712,
   USER_ADDRESS_PARAM_EIP4361,
+  // USER_ADDRESS_PARAM_EXTERNAL_EIP4361 is not reserved and can be used as a custom context parameter
+  // USER_ADDRESS_PARAM_EXTERNAL_EIP4361
```
derekpierre, line 38: Why not delete these lines instead of commenting them out?

derekpierre (commented):
One thought for this PR, was that the `AuthSignature` is the thing that is TACo-specific as opposed to the SIWE message. It could be helpful to have something (utility function or class or otherwise) that generates the `AuthSignature` type so that users don't need to know the specifics of it eg. `generateSIWEAuthSignature(address: string, siweMessage: string, signature: Uint8Array)`. With something like that, the user can use whatever they want for generating the siweMessage and `taco-auth` can take those as parameters and simply produce an `AuthSignature` type for them. 

It's not something overly flashy, but is helpful in abstracting away the creation of the `AuthSignature` a bit.

If they are using single sign-on then presumably they already have a SIWE library that they are using and we would be helping them convert that into a format that we can use i.e. the `AuthSignature`.

Thoughts?

packages/taco/src/conditions/const.ts
piotr-roslaniec: I thought it was more explicit (explicitly disabled), is it confusing?

piotr-roslaniec:
> Thoughts?

I think `AuthSignature` will become a `class` rather than an `interface`, and we will add more utilities and conversion functions as we go. Adding more factory methods, especially for SIWE, makes sense. I will look into it.

packages/taco-auth/src/providers/eip4361.ts
piotr-roslaniec: > Just a helper for the user to create a SIWE message to then use for the context variable?

Yes, it's just a helper function to expose a SIWE-message-creating logic to end-users. I think I will also refactor-out the types related to SIWE so that it's more configurable and gives users more freedom (as you've noted).

> so wondering if this is related?

Yes, it's related, but I planned on fixing the defaults later. I will attempt to squeeze it into this PR, but if it doesn't make sense I will follow up in the next one.

derekpierre: > Yes, it's just a helper function to expose a SIWE-message-creating logic to end-users. I think I will also refactor-out the types related to SIWE so that it's more configurable and gives users more freedom (as you've noted).

In that case, I'm a bit apprehensive about the benefits of us trying to provide extensive SIWE-message-creating logic. Then we just end up either making something more restrictive (less parameters than the SIWE API) or writing a layer on top of the SIWE library that uses the same API parameterization of the library anyway. Users should just use the `siwe` library or some other one for generating the SIWE message if they want such a high level of customization of messages.

At the moment, we have two worlds:
1. Apps that expect TACo to do the SIWE info for them because they don't use SIWE at all (`:userAddress`)
2. Apps that currently use SIWE and expect to provide TACo with their existing SIWE info so that it can be reused (`:userAddressExternalEIP4361`)

In world 1) apps don't currently use SIWE at all, so TACo will create/generate the AuthSignature on their behalf. At most, the app may need to specify the `domain` / `uri` if `taco-auth` can't detect those values from the `window` object / use some default value to get the right value based on the SIWE spec (i.e. prevent the signing from being flagged as "unsafe"). But these apps probably don't need more customization than that, or else they would probably be in world 2) anyway.

In world 2) the app uses their own tooling for the SIWE message/signature and all they want to do is give that information to TACo (Ceramic example). No customization of SIWE message parameters needed here because the message and signature are already generated by the app separately.

piotr-roslaniec: I see your point. I had a slightly different idea of how to go about this, but I see merit in this solution. 

I dropped the commit with previous changes and added an ability to set defaults for the EIP4361 provider. I will expose it as a part of https://github.com/nucypher/taco-web/issues/533

derekpierre: > I had a slightly different idea of how to go about this, but I see merit in this solution.

Happy to listen to your ideas.

manumonti (approved):

packages/taco/src/conditions/const.ts
derekpierre: > Maybe an assertion is better.

Interesting. Where would you put the assertion?

Alternatively, maybe keep the comment on L37 and remove L38?

packages/taco/src/conditions/context/context.ts
```diff
@@ -119,12 +129,20 @@ export class ConditionContext {
...
+        rvt.forEach((value) => {
+          if (ConditionContext.isContextParameter(value)) {
+            requestedParameters.add(value);
+          }
+        });
+      } else if (ConditionContext.isContextParameter(rvt)) {
         requestedParameters.add(rvt);
+      } else {
+        // Not a context parameter, we can skip
       }
```
derekpierre, line 146: Is the empty block for a comment something common in typescript? Perhaps just the comment?
```suggestion
      }
      // else not a context parameter, we can skip
```

packages/taco/src/conditions/context/context.ts
```diff
@@ -86,16 +88,24 @@ export class ConditionContext {
...
-      });
+    for (const param of requestedParameters) {
+      // If it's not a user address parameter, we can skip
+      if (!USER_ADDRESS_PARAMS.includes(param)) {
+        continue;
+      }
+
+      // If it's a user address parameter, we need to check if we have an auth provider
+      const authMethod = AUTH_METHOD_FOR_PARAM[param];
+      if (!authMethod && !this.customParameters[param]) {
```
derekpierre, line 99: For this case, should we also check that the `customParameters[param]` value is an `AuthSignature`? So not just that it is present in the custom parameters, but its type is correct. Or is that already done somewhere else?

piotr-roslaniec: I don't think we validate it against the structure of `AuthSignature`. We may want to use `zod` for that. Let's see.

packages/taco/src/conditions/context/context.ts
piotr-roslaniec: It's no TS-related, just something I sometimes do to be more explicit.

packages/taco/src/conditions/const.ts
piotr-roslaniec: An assertion could be a statement in this file that gets executed when the file is imported. That's a bit ham-fisted, I need to think about this. In the meantime, I will remove L38.

derekpierre: Agreed - the assertion seems too heavy.

packages/taco/src/conditions/context/context.ts
piotr-roslaniec: Documented in https://github.com/nucypher/taco-web/issues/535, will address in a follow-up PR

piotr-roslaniec:
I'm going to go ahead and merge this, but we can continue discussing here or in #537 if anything

PR #531, opened 2024-06-13
Authentication providers

piotr-roslaniec:
**Type of PR:**
- Feature
- Other

**Required reviews:**
-  2

**What this does:**
- Introduces authentication providers to be used internally 
- Exposes auth providers to internal APIs - No changs in user-facing TACo API
- Exposes method to discover required context parameters from condition context

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/packages/taco-auth/src/providers/eip4361.ts
+++ b/packages/taco-auth/src/providers/eip4361.ts
@@ -2,14 +2,15 @@ import { ethers } from 'ethers';
 import { generateNonce, SiweMessage } from 'siwe';
 
 import { LocalStorage } from '../storage';
-import { AuthSignature } from '../types';
+import { AuthSignature, EIP4361_AUTH_METHOD } from '../types';
 
-export type FormattedEIP4361 = string;
+export type EIP4361TypedData = string;
 
 export class EIP4361AuthProvider {
   private readonly storage: LocalStorage;
 
   constructor(
+    // TODO: We only need the provider to fetch the chainId, consider removing it
     private readonly provider: ethers.providers.Provider,
     private readonly signer: ethers.Signer,
   ) {
@@ -18,7 +19,7 @@ export class EIP4361AuthProvider {
 
   public async getOrCreateAuthSignature(): Promise<AuthSignature> {
     const address = await this.signer.getAddress();
-    const storageKey = `eth-signin-message-${address}`;
+    const storageKey = `eth-${EIP4361_AUTH_METHOD}-message-${address}`;
 
     // If we have a message in localStorage, return it
     const maybeMessage = this.storage.getItem(storageKey);
@@ -27,12 +28,12 @@ export class EIP4361AuthProvider {
     }
 
     // If at this point we didn't return, we need to create a new message
-    const typedSignature = await this.createSiweMessage();
-    this.storage.setItem(storageKey, JSON.stringify(typedSignature));
-    return typedSignature;
+    const authMessage = await this.createSIWEAuthMessage();
+    this.storage.setItem(storageKey, JSON.stringify(authMessage));
+    return authMessage;
   }
 
-  private async createSiweMessage(): Promise<AuthSignature> {
+  private async createSIWEAuthMessage(): Promise<AuthSignature> {
     const address = await this.signer.getAddress();
     const { domain, uri } = this.getParametersOrDefault();
     const version = '1';
@@ -53,6 +54,7 @@ export class EIP4361AuthProvider {
     return { signature, address, scheme, typedData: message };
   }
 
+  // TODO: Create a facility to set these parameters or expose them to the user
   private getParametersOrDefault() {
     // If we are in a browser environment, we can get the domain and uri from the window object
     if (typeof window !== 'undefined') {
--- a/packages/taco-auth/src/providers/eip4361.ts
+++ b/packages/taco-auth/src/providers/eip4361.ts
@@ -1,12 +1,12 @@
 import { ethers } from 'ethers';
 import { generateNonce, SiweMessage } from 'siwe';
 
-import { LocalStorage } from './storage';
-import { AuthSignature } from './types';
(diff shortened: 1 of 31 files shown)


packages/taco/src/taco.ts
```diff
@@ -153,6 +154,8 @@ export const decrypt = async (
     messageKit.acp.publicKey,
   );
   const ritual = await DkgClient.getActiveRitual(provider, domain, ritualId);
+  // TODO: Temporary helper method to keep the external taco.ts decrypt function simple
```
derekpierre, line 157: What's the long-term plan/vision for the TODO?

derekpierre:
> Exposes method to discover required context parameters from condition context

More of a general question and something I may have missed - how is the above done by apps?

packages/taco-auth/src/providers/eip4361.ts
```diff
@@ -1,24 +1,25 @@
...
-export type FormattedEIP4361 = string;
+export type EIP4361TypedData = string;
 
-export class EIP4361SignatureProvider {
+export class EIP4361AuthProvider {
   private readonly storage: LocalStorage;
 
   constructor(
+    // TODO: We only need the provider to fetch the chainId, consider removing it
     private readonly provider: ethers.providers.Provider,
```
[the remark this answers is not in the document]
piotr-roslaniec: Yes, that's correct. Signer is only to sign things, provider is to maintain communication with an external RPC / RPC provider

packages/taco/src/taco.ts
piotr-roslaniec: To expose authentication providers in the user-facing TACo API once we are ready to release. This is going to be a breaking change and it makes sense to bundle it with other auth-related changes, and maybe context parameters refactor (more user-facing API).

packages/taco/src/conditions/context/context.ts
```diff
@@ -213,37 +149,30 @@ export class ConditionContext {
...
-  }
-
-  public async toWASMContext(): Promise<Context> {
+  public async toCoreContext(): Promise<CoreContext> {
     const asJson = await this.toJson();
-    return new Context(asJson);
+    return new CoreContext(asJson);
   }
 
+  public toContextParameters = async (): Promise<Record<string, ContextParam>> => {
```
piotr-roslaniec, line 161: > More of a general question and something I may have missed - how is the above done by apps?
@derekpierre They would have to call this context method. We can make it more user-friendly.

derekpierre: This function is only available *after* `ConditionContext` creation and it returns the populated values. 

I'm wondering how I, as the app developer, know what context variables are needed for the condition in the ciphertext I'm about to decrypt - so that I know what context variable values I need to provide to the `decrypt(...)` call. (implicit v explicit).

`ConditionContext` has the `requiredParmeters` property on an instance, but to create an instance of `ConditionContext`, a customParameters dictionary must be provided (constructor, or fromConditions).

piotr-roslaniec: Decryptor is given `ThresholdMessageKit` which is then made into `ConditionContext` internally. So, not sure what is the best way to expose this to the user-facing API: `ConditionContext.requiredParameters(messageKit)`?

I will look into it.

packages/taco/src/conditions/condition-expr.ts
```diff
@@ -56,29 +56,27 @@ export class ConditionExpression {
...
       this.condition,
       customParameters,
-      signer,
+      authProviders,
     );
   }
 
-  public contextRequiresSigner(): boolean {
-    return this.condition.requiresSigner();
+  public contextRequiresAuthentication(): boolean {
```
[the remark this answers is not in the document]
piotr-roslaniec: I think at this moment we're using it in tests only. I will remove it.

packages/taco/src/conditions/context/context.ts
```diff
@@ -213,37 +151,35 @@ export class ConditionContext {
...
-      provider,
-      ConditionExpression.fromWASMConditions(conditions).condition,
+      ConditionExpression.fromCoreConditions(conditions).condition,
       customParameters,
-      signer,
+      authProviders,
     );
   }
+
+  public static requestedParameters(messageKit: ThresholdMessageKit): Set<string> {
```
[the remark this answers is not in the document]
piotr-roslaniec: I don't think it's that confusing, but `context.ts` is proving difficult to navigate so I will adjust the naming

derekpierre (approved):
🎸

packages/taco-auth/src/providers/eip4361.ts
```diff
@@ -53,6 +54,7 @@ export class EIP4361SignatureProvider {
     return { signature, address, scheme, typedData: message };
   }
 
+  // TODO: Create a facility to set these parameters or expose them to the user
```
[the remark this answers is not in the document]
piotr-roslaniec: I will do that on the back of exposing authentication providers from `taco`.

Created: https://github.com/nucypher/taco-web/issues/533

packages/taco-auth/src/types.ts
```diff
@@ -1,9 +1,33 @@
...
+  [EIP712_AUTH_METHOD]?: EIP712AuthProvider;
+  [EIP4361_AUTH_METHOD]?: EIP4361AuthProvider;
+  // Fallback to satisfy type checking
+  [key: string]: AuthProvider | undefined;
+};
 
 export interface AuthSignature {
   signature: string;
   address: string;
   scheme: 'EIP712' | 'EIP4361';
```
derekpierre, line 18: Is it helpful for these values to use the `*_AUTH_METHOD` constants below?

piotr-roslaniec: I'm afraid it's not so easy with TypeScripts, types and constants do not mix.

PR #532, opened 2024-06-21
Deprecate EIP712 authentication provider

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:**
- 1

**What this does:**
- Marks `EIP712AuthProvider` as deprecated

**Issues fixed/closed:**
- Refers to https://github.com/nucypher/tdec/issues/178

**Notes for reviewers:**
- Should we outright remove EIP712 support instead of deprecating it? If not, what is a good timeline to allow `taco-web` users to adjust?
- Created https://github.com/nucypher/taco-web/issues/536

### DIFF
--- a/packages/taco-auth/src/providers/eip712.ts
+++ b/packages/taco-auth/src/providers/eip712.ts
@@ -56,6 +56,9 @@ const EIP712Domain = [
   },
 ];
 
+/**
+ * @deprecated Use EIP4361AuthProvider instead.
+ */
 export class EIP712AuthProvider implements AuthProvider {
   private readonly storage: LocalStorage;
 
@@ -64,6 +67,10 @@ export class EIP712AuthProvider implements AuthProvider {
     private readonly provider: ethers.providers.Provider,
     private readonly signer: ethers.Signer,
   ) {
+    console.warn(
+      'DeprecationWarning: The EIP712AuthProvider authentication provider is deprecated. ' +
+      'Please use EIP4361AuthProvider instead. Refer to the documentation for more details.'
+    );
     this.storage = new LocalStorage();
   }
 
--- a/packages/taco-auth/src/types.ts
+++ b/packages/taco-auth/src/types.ts
@@ -19,6 +19,9 @@ export interface AuthSignature {
   typedData: EIP712TypedData | EIP4361TypedData;
 }
 
+/**
+ * @deprecated Use EIP4361_AUTH_METHOD instead.
+ */
 export const EIP712_AUTH_METHOD = 'EIP712';
 export const EIP4361_AUTH_METHOD = 'EIP4361';
 


packages/taco-auth/src/types.ts
```diff
@@ -19,10 +19,17 @@ export interface AuthSignature {
...
+/**
+ * @deprecated Use EIP4361_AUTH_METHOD instead.
+ */
 export const EIP712_AUTH_METHOD = 'EIP712';
 export const EIP4361_AUTH_METHOD = 'EIP4361';
 
 export const USER_ADDRESS_PARAM_DEFAULT = ':userAddress';
+
+/**
+ * @deprecated Use USER_ADDRESS_PARAM_EIP4361 instead.
```
derekpierre, line 31: These context variables were never publicly released - just the epic, right? If so, then a removal can be done instead of deprecation.

packages/taco-auth/src/providers/eip712.ts
```diff
@@ -64,6 +67,10 @@ export class EIP712AuthProvider implements AuthProvider {
     private readonly provider: ethers.providers.Provider,
     private readonly signer: ethers.Signer,
   ) {
+    console.warn(
+      'DeprecationWarning: The default authentication provider, EIP712AuthProvider, is deprecated. ' +
+      'Please use EIP4361AuthProvider instead. Refer to the documentation for more details.'
+    );
```
derekpierre, line 73: Since the use EIP712 was implicit until now, is this necessary?

`EIP712AuthProvider` was not actually used/specified by anyone in prior releases. Can we simply remove it?

derekpierre (commented):
> Should we outright remove EIP712 support instead of deprecating it? If not, what is a good timeline to allow taco-web users to adjust?

Since it was implicit before I think it can just be removed. The associated new context variables can also be removed because they were never publicly released. 

I think (?) the concern you are raising is a question for `nucypher` code instead of `taco-web` because everything was so implicit in the past - see the notes for reviewers in https://github.com/nucypher/nucypher/pull/3515.

Let me know if I'm missing something.

packages/taco-auth/src/types.ts
piotr-roslaniec: They were never exposed to the `taco` API, but the values they represent were in use, hence they should be deprecated.

packages/taco-auth/src/providers/eip712.ts
piotr-roslaniec: IMHO it doesn't matter if behavior is implicit or explicit, only whether the changes in the API are breaking, and this is a breaking change

packages/taco-auth/src/types.ts
piotr-roslaniec: No, I think you are right. I think that was my mistake. I will fix it.

packages/taco-auth/src/providers/eip712.ts
derekpierre: I'm not sure I follow. I believe this file/module and the class was only added as part of the unreleased epic (the epic we are merging this PR into), so why keep the file/module and the class that we no longer intend to use as part of that epic and was never released...

By "implicit" I mean that no user has ever specified or created a EIP712AuthProvider object, because it was never released. So there is no real "breaking" for this class/module since has never been released.  The broader API (`decrypt`) will indeed have "breaking" changes but that is irrespective of this particular auth provider, or the scheme used for validating wallets. 

It's possible that a warning could be used somewhere, but I don't think this is the spot, since it seems like this file/module should just be deleted.

piotr-roslaniec: We've released an API with a documented behavior (using EIP712 by default) and are now replacing that behavior with another (EIP4361). I don't know specifically how this change can affect our users (is it going to "break" anything) but changing the implicit (or explicit) behavior of an API is a breaking change (according to semver).

Deprecating EIP712 in T+1 major version will allow us to communicate this breaking change to users who are currently using EIP712. Again, I don't know whether it will actually break something, but it makes sense to announce it and only remove it in T+2 after this upcoming change is communicated.

derekpierre: I see. Ok, so the goal of keeping the file is more to keep this (EIP712)  as an option (albeit a pending deprecating one), in case some adopter decides they want to use EIP712 instead of EIP4361 for `:userAddress`. That's the reason for keeping this module for now.

If we are affording that option, then the server-side will need to be adjusted to accept both EIP712 and EIP4361 for `:userAddress` and https://github.com/nucypher/nucypher/pull/3515 will need to be adjusted.

piotr-roslaniec: So the server side verification will be broken for EIP712 in the next release? Not deprecated, but outright removed?

In that case, these changes have no sense and we should remove EIP712 from the client too. I will address it in this PR.

derekpierre: > So the server side verification will be broken for EIP712 in the next release? Not deprecated, but outright removed?

That's still a discussion for #3515, and not currently set in stone. Still need to make an official decision there.

piotr-roslaniec:
Accidentally closed during a rebase

cygnusv (approved):

packages/taco-auth/src/types.ts
derekpierre: ^ Is the goal to remove these additional context variables in a separate PR then?

piotr-roslaniec: Yes, there's a follow-up issue for that: https://github.com/nucypher/taco-web/issues/536

derekpierre: But these don't need to wait T+2 releases to be removed ... ?

Issue #533, opened 2024-06-24
Expose authentication providers from `@nucypher/taco`


piotr-roslaniec:
- Today, users providers `signer: ethers.Signer` instances 
- With this PR, users should be able to configure authentication providers for `taco` methods
- Provide users with facilities to configure providers where needed (see [this comment](https://github.com/nucypher/taco-web/pull/531/files/3eb50c0195f4f2dd528b409e10ebc15bdf6ea1e2#r1651005660))

piotr-roslaniec:
Closed by #534

PR #534, opened 2024-06-25
Expose auth providers from `taco`

piotr-roslaniec:
**Type of PR:**
-  Feature

**Required reviews:**
- 2

**What this does:**
- `decrypt` method in `taco` now takes `authProvider: EIP4361` instead of `signer: ethers.Signer`
- Exposes `EIP4361Provider` from `taco` rather than `taco-auth` to avoid fragmenting TACo API across multiple packages. To be reassessed when/if we publish `taco-auth`
- Users can now modify SIWE parameters in `EIP4361Provider` 

**Issues fixed/closed:**
- Fixes https://github.com/nucypher/taco-web/issues/533

**Why it's needed:**

**Notes for reviewers:**
- Fails on Lynx because of incompatibility with v7.3.0


manumonti (approved):
LGTM! 🙌

I just left a little thought I had.

packages/taco-auth/src/providers/eip4361.ts
```diff
@@ -18,14 +18,33 @@ const ERR_MISSING_SIWE_PARAMETERS = 'Missing default SIWE parameters';
...
+    if (typeof window !== 'undefined') {
+      // If we are in a browser environment, we can get the domain and uri from the window object
+      const maybeOrigin = window?.location?.origin;
+      return {
+        domain: maybeOrigin.split('//')[1].split('.')[0],
+        uri: maybeOrigin,
+      };
+    }
+    // If not, we have no choice but to throw an error
+    throw new Error(ERR_MISSING_SIWE_PARAMETERS);
```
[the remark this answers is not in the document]
piotr-roslaniec: If the provider is not specified, there would be a different error thrown somewhere (during the construction of condition context). So we are already taking care of that case in a different branch of code. 

This particular error is thrown in the provider constructor, so we would enter a different branch of code here (we are trying to make the provider). Does that make sense?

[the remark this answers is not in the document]
manumonti: Ah, yes. Gotcha.

packages/taco/src/taco.ts
```diff
@@ -143,6 +143,25 @@ export const decrypt = async (
...
+    provider,
+    domain,
+    messageKit,
+    authProviders,
+    porterUri,
+    customParameters,
+  );
+};
+
+export const decryptWithAuthProviders = async (
```
derekpierre, line 158: If we are accepting that we are making backwards incompatible changes for the next release (major version bump), do we need a new method, as opposed to just changing the old method?

piotr-roslaniec: It wasn't in the cards when I made this PR. I will address necessary changes in the follow-up PR or close this PR without merging.

packages/taco/src/taco.ts
```diff
@@ -143,6 +143,25 @@ export const decrypt = async (
...
+  provider: ethers.providers.Provider,
+  domain: Domain,
+  messageKit: ThresholdMessageKit,
+  authProviders?: AuthProviders,
```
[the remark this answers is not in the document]
piotr-roslaniec: If we don't foresee new providers near-term, it could be just `authProvider: EIP4361Provider`

piotr-roslaniec:
Fails on Lynx because of incompatibility with v7.3.0

packages/taco-auth/src/providers/eip4361.ts
```diff
@@ -18,14 +18,33 @@ const ERR_MISSING_SIWE_PARAMETERS = 'Missing default SIWE parameters';
...
+  }
+
+  private getDefaultParameters() {
+    if (typeof window !== 'undefined') {
+      // If we are in a browser environment, we can get the domain and uri from the window object
+      const maybeOrigin = window?.location?.origin;
+      return {
+        domain: maybeOrigin.split('//')[1].split('.')[0],
+        uri: maybeOrigin,
+      };
```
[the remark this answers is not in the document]
piotr-roslaniec: Yes, `document` and `window` are globally available objects provided by the browser environment.

I updated this snipper.

packages/taco-auth/src/providers/eip4361.ts
```diff
@@ -46,7 +64,7 @@ export class EIP4361AuthProvider {
 
   private async createSIWEAuthMessage(): Promise<AuthSignature> {
     const address = await this.signer.getAddress();
-    const { domain, uri } = this.getParametersOrDefault();
+    const { domain, uri } = this.providerParams;
     const version = '1';
     const nonce = generateNonce();
     const chainId = (await this.provider.getNetwork()).chainId;
```
derekpierre, line 70: Can we do `signer.getChainId( )` here and get rid of `provider` from the constructor?

packages/taco/src/taco.ts
```diff
@@ -140,8 +140,8 @@ export const decrypt = async (
   provider: ethers.providers.Provider,
   domain: Domain,
   messageKit: ThresholdMessageKit,
+  authProvider: EIP4361AuthProvider,
```
derekpierre, line 143: Should this be optional or is it always provided? It's not needed if the condition doesn't actually use the user address context variable.

packages/taco-auth/src/providers/eip4361.ts
piotr-roslaniec: Not every `signer` has a connected `provider`. In `ethers`, it's possible to create a signer using a raw private key without ever connecting to an RPC. 

I think we may want to explore requiring that every `signer` has a `provider` already connected and to do that in the whole codebase. But I think this is for another PR since we would be changing quite a lot of APIs.

packages/taco/src/taco.ts
piotr-roslaniec: Right, that makes sense

manumonti (approved):
LGTM! 👌

manumonti (approved):

packages/taco-auth/src/providers/eip4361.ts
derekpierre: Ah ok.

derekpierre (approved):
🎸

Issue #535, opened 2024-06-28
Validate use-provided authentication claims


piotr-roslaniec:
- Validate user-provided `AuthSignature`, both from `localStorage` and context variables, using `zod`

derekpierre:
Closed via #535

PR #527, opened 2024-06-05
[EPIC] Auth

derekpierre:
**Type of PR:**

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

Customizable Authentication.

Comprised of PRs:
- [x] #518 
- [x] #526 
- [x] #529 
- [ ] #530 - Not a blocker
- [x] #531
- [x] #532
- [x] #534   
- [x] #537 
- [x] #546 
- [x] #547

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
> What should reviewers focus on?
> Is there a particular commit/function/section of your PR that requires more attention from reviewers?

PR #537, opened 2024-06-28
Validate user-provided `AuthSignature`

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:**
- 2

**What this does:**
- Implements `AuthSignature` validation with `zod`
- Moves some code around in `taco-auth` to prevent circular-dependency issues that break `zod`

**Issues fixed/closed:**
- Fixes https://github.com/nucypher/taco-web/issues/535

**Why it's needed:**

**Notes for reviewers:**
- ...

### DIFF
--- a/packages/shared/src/schemas.ts
+++ b/packages/shared/src/schemas.ts
@@ -1,0 +1,4 @@
+import { z } from 'zod';
+
+export const ETH_ADDRESS_REGEXP = new RegExp('^0x[a-fA-F0-9]{40}$');
+export const EthAddressSchema = z.string().regex(ETH_ADDRESS_REGEXP);
--- a/packages/shared/src/schemas.ts
+++ b/packages/shared/src/schemas.ts
@@ -1,4 +1,16 @@
+import { ethers } from 'ethers';
 import { z } from 'zod';
 
 export const ETH_ADDRESS_REGEXP = new RegExp('^0x[a-fA-F0-9]{40}$');
-export const EthAddressSchema = z.string().regex(ETH_ADDRESS_REGEXP);
+
+const isAddress = (address: string) => {
+  try {
+    return ethers.utils.getAddress(address);
+  } catch {
+    return false;
+  }
+};
+
+export const EthAddressSchema = z.string()
+  .regex(ETH_ADDRESS_REGEXP)
+  .refine(isAddress, { message: 'Invalid Ethereum address' });
--- a/packages/taco-auth/src/providers/eip4361.ts
+++ b/packages/taco-auth/src/providers/eip4361.ts
@@ -1,10 +1,13 @@
 import { ethers } from 'ethers';
 import { generateNonce, SiweMessage } from 'siwe';
+import { z } from 'zod';
 
+import { EIP4361_AUTH_METHOD } from '../auth-provider';
+import { AuthSignature } from '../auth-sig';
 import { LocalStorage } from '../storage';
-import { AuthSignature, EIP4361_AUTH_METHOD } from '../types';
 
-export type EIP4361TypedData = string;
+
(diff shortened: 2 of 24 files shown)


packages/taco-auth/src/storage.ts
```diff
@@ -36,11 +38,24 @@ export class LocalStorage {
         : new BrowserStorage();
   }
 
-  getItem(key: string): string | null {
-    return this.storage.getItem(key);
+  public getAuthSignature(key: string): AuthSignature | null {
```
theref, line 41: :rocket:

theref (approved):
Looks great

packages/test-utils/src/utils.ts
```diff
@@ -53,8 +53,8 @@ export const fromBytes = (bytes: Uint8Array): string =>
 
 export const fakePorterUri = 'https://_this_should_crash.com/';
 
-const makeFakeProvider = (timestamp: number, blockNumber: number) => {
-  const block = { timestamp };
+const makeFakeProvider = (timestamp: number, blockNumber: number, blockHash: string) => {
```
manumonti, line 56: Not relevant for this PR, but shouldn't the code formatter split this line in several ones?


```suggestion
const makeFakeProvider = (
  timestamp: number,
  blockNumber: number,
  blockHash: string,
) => {
```

packages/shared/src/schemas.ts
```diff
@@ -0,0 +1,4 @@
+import { z } from 'zod';
+
+export const ETH_ADDRESS_REGEXP = new RegExp('^0x[a-fA-F0-9]{40}$');
+export const EthAddressSchema = z.string().regex(ETH_ADDRESS_REGEXP);
```
cygnusv, line 4: Should we go a step further and validate the address with ethers and a zod `.refine`?

https://docs.ethers.org/v6/api/address/#getAddress

packages/taco-auth/src/storage.ts
```diff
@@ -36,11 +38,24 @@ export class LocalStorage {
...
+    this.storage.setItem(key, asJson);
+  }
+
+  public static fromJson(json: string): AuthSignature {
+    return authSignatureSchema.parse(JSON.parse(json));
   }
 
-  setItem(key: string, value: string): void {
-    this.storage.setItem(key, value);
+  public static toJson(signature: AuthSignature): string {
```
cygnusv, line 58: Not a blocker but it's a bit weird that json<=>authsignature conversion methods are part of the storage interface

manumonti (approved):
LGTM! 👌

packages/taco-auth/src/providers/eip4361.ts
```diff
@@ -1,10 +1,13 @@
...
+import { z } from 'zod';
 
+import { EIP4361_AUTH_METHOD } from '../auth-provider';
+import { AuthSignature } from '../auth-sig';
 import { LocalStorage } from '../storage';
-import { AuthSignature, EIP4361_AUTH_METHOD } from '../types';
 
-export type EIP4361TypedData = string;
+
+export const EIP4361TypedDataSchema = z.string();
```
derekpierre, line 10: Should we use zod `refine` here as well? You can try calling the `SiweMessage` constructor as part of the refine function to ensure that the string can be parsed properly.

piotr-roslaniec: Addressed here: [`f9e2030` (#527)](https://github.com/nucypher/taco-web/pull/527/commits/f9e203023327dc343e9aaf26fe2a362d5c8f2683)

derekpierre: 🎸

PR #542, opened 2024-07-02
Replace deprecated Porter URIs

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:**
- 1

**Notes for reviewers:**
- Not sure about lynx: 
  - https://porter-lynx.nucypher.community/ works 
  - https://porter-lynx.nucypher.io/ doesn't work

### DIFF
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -13,9 +13,8 @@ import { Base64EncodedBytes, ChecksumAddress, HexEncodedBytes } from './types';
 import { fromBase64, fromHexString, toBase64, toHexString } from './utils';
 
 const porterUri: Record<string, string> = {
-  mainnet: 'https://porter.nucypher.community',
-  tapir: 'https://porter-tapir.nucypher.community',
-  oryx: 'https://porter-oryx.nucypher.community',
+  mainnet: 'https://porter.nucypher.io',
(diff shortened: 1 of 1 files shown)


packages/shared/src/porter.ts
```diff
@@ -13,9 +13,8 @@ import { Base64EncodedBytes, ChecksumAddress, HexEncodedBytes } from './types';
 import { fromBase64, fromHexString, toBase64, toHexString } from './utils';
 
 const porterUri: Record<string, string> = {
-  mainnet: 'https://porter.nucypher.community',
-  tapir: 'https://porter-tapir.nucypher.community',
-  oryx: 'https://porter-oryx.nucypher.community',
+  mainnet: 'https://porter.nucypher.io',
+  tapir: 'https://porter-tapir.nucypher.io',
   lynx: 'https://porter-lynx.nucypher.community',
```
derekpierre, line 18: If we are switching, should we just switch all? i.e. also switch lynx - cc @KPrasch

piotr-roslaniec: Yes, but currently https://porter-lynx.nucypher.io/ doesn't work

derekpierre: Yep. I just mean that we shouldn't merge this before making that happen.

derekpierre:
Will be fixed via #555 .

PR #543, opened 2024-07-04
Update `mainnet` example

piotr-roslaniec:
**Type of PR:**
- Other

**Required reviews:**
- 1

**What this does:**
- Fixes bad RPC URL causing `mainnet` job to fail
- Adds a check for `isAuthorized` method in nodejs example

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**
- Tested locally
- The example action will only run on `main`, so we won't see it in GH Actions until we merge

### DIFF
--- a/examples/taco/nodejs/src/index.ts
+++ b/examples/taco/nodejs/src/index.ts
@@ -73,7 +73,8 @@ const encryptToBytes = async (messageString: string) => {
     encryptorSigner,
   );
 
-  // Calling it after the encryption because we need material from messageKit
+  // Note: Not actually needed but used by CI for checking contract compatibility.
+  // Calling it after the encryption because we need material from messageKit.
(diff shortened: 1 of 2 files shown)


examples/taco/nodejs/src/index.ts
```diff
@@ -72,6 +73,17 @@ const encryptToBytes = async (messageString: string) => {
     encryptorSigner,
   );
 
+  // Calling it after the encryption because we need material from messageKit
```
derekpierre, line 76: Since it is an example perhaps add something about this check not really being needed, so that users don't assume it needs to be added to their code.

```suggestion
  // Note: Not actually needed but used by CI for checking contract compatibility.
  // Calling it after the encryption because we need material from messageKit.
```

derekpierre (approved):
🎸 . Approved with one suggestion/comment.

PR #550, opened 2024-07-17
Add JsonApiCondition

theref:
Fixes https://github.com/nucypher/sprints/issues/23

### DIFF
--- a/packages/taco/src/conditions/base/json-api.ts
+++ b/packages/taco/src/conditions/base/json-api.ts
@@ -1,22 +1,32 @@
+import { JSONPath } from 'jsonpath-plus';
 import { z } from 'zod';
 
 import { Condition } from '../condition';
 import {
-  EthAddressOrUserAddressSchema,
   OmitConditionType,
   paramOrContextParamSchema,
   returnValueTestSchema,
 } from '../shared';
 
-export const JsonApiConditionType = 'json';
+export const JsonApiConditionType = 'json-api';
 
 export const JsonApiConditionSchema = z.object({
   conditionType: z.literal(JsonApiConditionType).default(JsonApiConditionType),
-  parameters: z.union([
-    z.array(EthAddressOrUserAddressSchema).nonempty(),
-    // Using tuple here because ordering matters
-    z.tuple([EthAddressOrUserAddressSchema, paramOrContextParamSchema]),
-  ]),
+  endpoint: z.string().url(),
+  parameters: z.array(paramOrContextParamSchema),
+  query: z.string().refine(
+    (path) => {
+      try {
+        JSONPath.toPathArray(path);
+        return true;
+      } catch (error) {
+        return false;
+      }
+    },
+    {
+      message: "Invalid JSON path",
+    }
+  ),
   returnValueTest: returnValueTestSchema, // Update to allow multiple return values after expanding supported methods
 });
 
--- a/packages/taco/src/conditions/base/json-api.ts
+++ b/packages/taco/src/conditions/base/json-api.ts
@@ -1,0 +1,32 @@
+import { z } from 'zod';
+
+import { Condition } from '../condition';
+import {
+  EthAddressOrUserAddressSchema,
+  OmitConditionType,
+  paramOrContextParamSchema,
+  returnValueTestSchema,
+} from '../shared';
+
+export const JsonApiConditionType = 'json';
+
+export const JsonApiConditionSchema = z.object({
+  conditionType: z.literal(JsonApiConditionType).default(JsonApiConditionType),
+  parameters: z.union([
+    z.array(EthAddressOrUserAddressSchema).nonempty(),
+    // Using tuple here because ordering matters
+    z.tuple([EthAddressOrUserAddressSchema, paramOrContextParamSchema]),
+  ]),
+  returnValueTest: returnValueTestSchema, // Update to allow multiple return values after expanding supported methods
+});
+
+export type JsonApiConditionProps = z.infer<typeof JsonApiConditionSchema>;
+
+export class JsonApiCondition extends Condition {
+  constructor(value: OmitConditionType<JsonApiConditionProps>) {
+    super(JsonApiConditionSchema, {
+      conditionType: JsonApiConditionType,
+      ...value,
+    });
+  }
+}
--- a/packages/taco/src/conditions/base/json-api.ts
+++ b/packages/taco/src/conditions/base/json-api.ts
@@ -11,8 +11,8 @@ export const JsonApiConditionType = 'json-api';
 export const JsonApiConditionSchema = z.object({
   conditionType: z.literal(JsonApiConditionType).default(JsonApiConditionType),
   endpoint: z.string().url(),
-  parameters: z.record(z.string(), z.unknown()),
-  query: z.string(),
+  parameters: z.record(z.string(), z.unknown()).optional(),
+  query: z.string().optional(),
   returnValueTest: returnValueTestSchema, // Update to allow multiple return values after expanding supported methods
 });
 
--- a/packages/taco/src/conditions/base/json-api.ts
+++ b/packages/taco/src/conditions/base/json-api.ts
@@ -1,10 +1,8 @@
-import { JSONPath } from 'jsonpath-plus';
 import { z } from 'zod';
 
 import { Condition } from '../condition';
 import {
   OmitConditionType,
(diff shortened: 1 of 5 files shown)


derekpierre:
Just a side note - I don't think this should go into `main`; at least not yet. The next release of `taco-web` (SIWE etc.) should not have this code.

`JsonApiCondition` needs to be released on the server-side first and won't be released until `7.5.0`. Perhaps instead you merge this PR into an epic/dev branch for now? Perhaps there's an existing alternative strategy that @piotr-roslaniec already uses...?

theref:
The original issue https://github.com/nucypher/sprints/issues/23 specifically says "Get taco-web preemptively compatible..." so I assumed there was some reason for having this merged before `7.5.0` is release

derekpierre:
> The original issue [nucypher/sprints#23](https://github.com/nucypher/sprints/issues/23) specifically says "Get taco-web preemptively compatible..." so I assumed there was some reason for having this merged before `7.5.0` is release

Yep. The goal is for everyone to start getting more familiar with `taco-web`.  This task pushes that forward. 

The "pre-emptive" part is what I'm referring to. The code will be used, just not in the next release of `taco-web`. The next release of `taco-web` pretty much boils down to Sign-in With Ethereum support (https://github.com/nucypher/taco-web/pull/527 and some other fixes). If we merge this PR into `main` then it will be co-mingled with SIWE support which is not what we want. These are the steps for `nucypher` 7.4.0 release and subsequent taco-web release - https://github.com/nucypher/sprints/issues/5#issuecomment-2209378181. 

Then a similar set of steps for the 7.5.0 `nucypher` release (w/ `JsonApiCondition` support) and subsequent taco-web release but instead of SIWE it'll be the `JsonApiCondition` feature.

packages/taco/test/test-utils.ts
```diff
@@ -222,6 +226,17 @@ export const testTimeConditionObj: TimeConditionProps = {
   chain: TEST_CHAIN_ID,
 };
 
+export const testJsonApiConditionObj = {
+  conditionType: JsonApiConditionType,
+  endpoint: 'https://api.coingecko.com/api/v3/simple/price',
```
piotr-roslaniec, line 231: Does a request to this endpoint actually goes off during tests?

theref: no, that endpoint would only be hit at decryption (condition evaluation) time

piotr-roslaniec: We should avoid calling external resources during tests. We can mock this request during evaluation and replace it with `https://_this_would_totally_fail.com` to make sure it crashes if we forget to use the mock etc

theref: it's never called during a test, it's just an example url to test the schema

theref: fixed in https://github.com/nucypher/taco-web/pull/550/commits/466fc939f4e89e13fd74a65b20fc6181910a64bc

packages/taco/src/conditions/base/json-api.ts
```diff
@@ -0,0 +1,28 @@
...
+  returnValueTestSchema,
+} from '../shared';
+
+export const JsonApiConditionType = 'json-api';
+
+export const JsonApiConditionSchema = z.object({
+  conditionType: z.literal(JsonApiConditionType).default(JsonApiConditionType),
+  endpoint: z.string().url(),
+  parameters: z.record(z.string(), z.unknown()).optional(),
+  query: z.string().optional(),
```
KPrasch, line 15: I'm not sure how to implement this, but this field is expected to be a valid JSONPath query.  It may be helpful to consumers to have a taco-web level validation of these queries to prevent invalid conditions from by created.  Perhaps this can be done with a jsonpath library or a rudimentary regex.

derekpierre: Json path strings have a specific format/syntax. It would be good to ensure that the format/syntax is correct at construction time. Perhaps https://www.npmjs.com/package/jsonpath / https://www.npmjs.com/package/@types/jsonpath can be of use for validation. Here's an example we did for SIWE messages - https://github.com/nucypher/taco-web/pull/527/files#diff-979bf898c06acdb90eb18e31377b0d4423990d3299e14a7596b3efca73272b49R20.

KPrasch: > as i mentioned on the call

Ah, perhaps I missed that comment, my bad.

>  The path only becomes valid/invalid when there is some json associated with it

This is a good point for syntactically valid expressions, however...

> any string is a valid json path

This is not true since it's possible to create syntactically invalid json path expressions.  Nonetheless, this is not a blocking issue but does represent a small improvement in developer experience.  Here are some reasonable (imo) examples:

`$.store.book[?(@.price < ]`
 The filter expression is incomplete and not properly closed.

`$[store][book]`
Incorrect usage of square brackets. Proper syntax should be `$.store.book` or `$.store['book']`.

`$.store.book[*`
Asterisk wildcard is not properly closed with a bracket. It should be `$.store.book[*]`.

theref: this is incorrect, without any json to compare with `$.store.book[?(@.price < ]` is a valid json path, because the key in the dictionary could be the string `"book[?(@.price < ]"`

I have gone through this with some of the json path libraries already

`$[store][book]` again, the key could be the string "[store][book]"

KPrasch: In order to parse the string literal `book[?(@.price < ]` as a dictionary key that part of the expression must be enclosed in qutation marks?  Something like `$.store['book[?(@.price < ]']`

theref: ok, i'll add it back in and some tests

theref: yeah i understand this - my point is that if the field `Fields('book[?(@.price < ]')` exists then it's a valid json path

theref: maybe when I had this in my testing examples where wrong, but I've added it back in https://github.com/nucypher/taco-web/pull/550/commits/4546b2f6503da046ec275748c79e0e42d1673ff1
Currently it fails, due to the path being valid without any json to compare with

KPrasch: > yeah i understand this - my point is that if the field `Fields('book[?(@.price < ]')` exists then it's a valid json path

It's only syntactically valid if the expression has the correct enclosing quotes (`'$.store[\'book[?(@.price < ]\']'`).  Also consider unclosed square brackets for example (`'$.store.book[*'` is syntactically invalid but can be made valid by adding the appropriate quote and square bracket `'$.store[\'book[*\']'` is valid).   

My point is: it is possible to create syntactically invalid jsonpath expressions.  There's nothing we can do to validate that the expression will be valid against a JSON endpoint's response but we can easily detect syntax errors.

KPrasch (approved):
Overall looks good 👍🏻 I'm glad to see that this was a straight-forward implementation following the changes to nucypher/nucypher.  I left several comments about detecting JSONPath syntax errors but it's not a blocking concern. I opened up https://github.com/nucypher/taco-web/issues/559 to track it.  It's reasonable to address this in a follow-up PR imo.  

Also it looks like there are several CI failures.  It will be good to get the green check mark before merging.

Nice work!

packages/taco/src/conditions/base/json-api.ts
```diff
@@ -0,0 +1,28 @@
...
+  OmitConditionType,
+  returnValueTestSchema,
+} from '../shared';
+
+export const JsonApiConditionType = 'json-api';
+
+export const JsonApiConditionSchema = z.object({
+  conditionType: z.literal(JsonApiConditionType).default(JsonApiConditionType),
+  endpoint: z.string().url(),
+  parameters: z.record(z.string(), z.unknown()).optional(),
```
piotr-roslaniec, line 14: Does `parameters` take any primitive type as a parameter, or just `string` and `object`? Looking at the tests etc., I'm under the impression it's `object` only.
```suggestion
  parameters: z.record(z.object()).optional(),
```

theref: I'm not sure I follow. The example lingo from the issue is:
```
    "parameters": {
        "ids": "ethereum",
        "vs_currencies": "usd",
    },
```
so i guess it could be `z.record(z.string(), z.object()).optional(),`

PR #553, opened 2024-07-18
Multiple Porter URIs

### LINKED ISSUE #320: Fallback chain of Porters
Configure the Taco API to use a fallback chain of Porters. This chain is merely a list of URLs that can be sourced from a combination of (i) hardcoded, (ii) retrieved from a repo, and (iii) user configured values.

vzotova:
**Type of PR:**

- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
Refs #320

**Why it's needed:**

**Notes for reviewers:**
Based on #527

### DIFF
--- a/demos/taco-demo/src/App.tsx
+++ b/demos/taco-demo/src/App.tsx
@@ -3,7 +3,6 @@ import {
   decrypt,
   domains,
   encrypt,
-  getPorterUri,
   initialize,
   ThresholdMessageKit,
   toHexString,
@@ -89,7 +88,7 @@ export default function App() {
       provider,
       domain,
       encryptedMessage,
-      getPorterUri(domain),
+      undefined,
       provider.getSigner(),
     );
 
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -35,10 +35,14 @@ export const getPorterUri = (domain: Domain): string => {
   return getPorterUris(domain)[0];
 };
 
-export const getPorterUris = (domain: Domain, porterUri?: string): string[] => {
+export const getPorterUris = (domain: Domain, porterUri?: string | string[]): string[] => {
   const porterUris: string[] = [];
   if (porterUri) {
-    porterUris.push(porterUri);
+    if (porterUri instanceof Array) {
+      porterUris.push(...porterUri);
+    } else {
+      porterUris.push(porterUri);
+    }
   }
   const uri = defaultPorterUri[domain];
   if (!uri) {
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -6,7 +6,11 @@ import {
   RetrievalKit,
   TreasureMap,
 } from '@nucypher/nucypher-core';
-import axios, { AxiosResponse } from 'axios';
+import axios, {
+  AxiosRequestConfig,
+  AxiosResponse,
+  HttpStatusCode,
+} from 'axios';
 import qs from 'qs';
 
 import { Base64EncodedBytes, ChecksumAddress, HexEncodedBytes } from './types';
@@ -120,10 +124,47 @@ export type TacoDecryptResult = {
 };
 
 export class PorterClient {
-  readonly porterUrl: URL;
+  readonly porterUrls: URL[];
 
-  constructor(porterUri: string) {
-    this.porterUrl = new URL(porterUri);
+  constructor(porterUris: string | [string, ...[string]]) {
+    if (porterUris instanceof Array) {
+      this.porterUrls = porterUris.map((uri) => new URL(uri));
+    } else {
+      this.porterUrls = [new URL(porterUris)];
+    }
+  }
+
+  protected async tryAndGet<T, D>(
+    path: string,
+    config?: AxiosRequestConfig<D>,
+  ): Promise<AxiosResponse<T>> {
+    let resp!: AxiosResponse<T>;
+    for (const porterUrl of this.porterUrls) {
+      resp = await axios.get(new URL(path, porterUrl).toString(), config);
+      if (resp.status === HttpStatusCode.Ok) {
+        return resp;
+      }
+    }
+    return resp;
+  }
+
+  protected async tryAndPost<T, D>(
+    path: string,
+    data?: D,
+    config?: AxiosRequestConfig<D>,
+  ): Promise<AxiosResponse<T>> {
+    let resp!: AxiosResponse<T>;
+    for (const porterUrl of this.porterUrls) {
+      resp = await axios.post(
+        new URL(path, porterUrl).toString(),
+        data,
+        config,
+      );
+      if (resp.status === HttpStatusCode.Ok) {
+        return resp;
+      }
+    }
+    return resp;
   }
 
   public async getUrsulas(
@@ -136,8 +177,8 @@ export class PorterClient {
       exclude_ursulas: excludeUrsulas,
       include_ursulas: includeUrsulas,
     };
-    const resp: AxiosResponse<GetUrsulasResult> = await axios.get(
-      new URL('/get_ursulas', this.porterUrl).toString(),
+    const resp: AxiosResponse<GetUrsulasResult> = await this.tryAndGet(
+      '/get_ursulas',
       {
         params,
         paramsSerializer: (params) => {
@@ -170,10 +211,8 @@ export class PorterClient {
       bob_verifying_key: toHexString(bobVerifyingKey.toCompressedBytes()),
       context: conditionContextJSON,
     };
-    const resp: AxiosResponse<PostRetrieveCFragsResponse> = await axios.post(
-      new URL('/retrieve_cfrags', this.porterUrl).toString(),
-      data,
-    );
+    const resp: AxiosResponse<PostRetrieveCFragsResponse> =
+      await this.tryAndPost('/retrieve_cfrags', data);
 
     return resp.data.result.retrieval_results.map(({ cfrags, errors }) => {
       const parsed = Object.keys(cfrags).map((address) => [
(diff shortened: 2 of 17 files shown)


packages/shared/src/porter.ts
```diff
@@ -120,10 +133,47 @@ export type TacoDecryptResult = {
...
+    if (porterUris instanceof Array) {
+      this.porterUrls = porterUris.map((uri) => new URL(uri));
+    } else {
+      this.porterUrls = [new URL(porterUris)];
+    }
+  }
 
-  constructor(porterUri: string) {
-    this.porterUrl = new URL(porterUri);
+  protected async tryAndGet<T, D>(
```
piotr-roslaniec, line 146: You can make this even more generic by using `axios.request` and passing in `GET` and `POST` methods

vzotova: thanks
✔️

packages/shared/src/porter.ts
```diff
@@ -28,11 +32,24 @@ export const domains: Record<string, Domain> = {
 };
 
 export const getPorterUri = (domain: Domain): string => {
-  const uri = porterUri[domain];
+  return getPorterUris(domain)[0];
+};
+
+export const getPorterUris = (domain: Domain, porterUri?: string | string[]): string[] => {
```
piotr-roslaniec, line 38: What do you think about updating this function signature like this? In-place initialization of the default value, we can also demand that the users pass in an array no matter what.
```suggestion
export const getPorterUris = (domain: Domain, porterUris: string[] = []): string[] => {
```

piotr-roslaniec: It should also simplify other methods slightly

packages/shared/src/porter.ts
```diff
@@ -120,10 +137,35 @@ export type TacoDecryptResult = {
...
+        continue;
+      }
+      if (resp.status === HttpStatusCode.Ok) {
+        return resp;
+      }
+    }
+    if (lastError !== undefined) {
+      throw lastError;
+    }
+    throw new Error("Porter returns bad response");
```
piotr-roslaniec, line 168: So this line only runs if Porter returned an HTTP code different than `200 OK`? But no exception was thrown.

I think it would be helpful to somehow propagate the response from Porter in this exception, maybe like so:
```suggestion
    throw new Error(`Porter returns bad response: ${resp.status} - ${resp.message}`);
```
This wouldn't work on this particular line, but maybe we can squeeze it into an `else` statement after line 163.

Just throwing ideas, WDYT?

packages/shared/test/porter.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    expect(getUrsulasSpy).toBeCalledTimes(fakePorterUris.length);
+    expect(getUrsulasSpy).toHaveBeenNthCalledWith(1, expect.objectContaining(
+      { ...params, baseURL: fakePorterUris[0] })
+    );
+    expect(getUrsulasSpy).toHaveBeenNthCalledWith(2, expect.objectContaining(
+      { ...params, baseURL: fakePorterUris[1] })
+    );
+    expect(getUrsulasSpy).toHaveBeenNthCalledWith(3, expect.objectContaining(
+      { ...params, baseURL: fakePorterUris[2] }));
+  });
```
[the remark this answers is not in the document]
vzotova: ✔️

packages/shared/src/porter.ts
vzotova: ✔️

packages/shared/src/porter.ts
vzotova: Actually it will work because resp at this line won't be empty and it will be last bad response
✔️

piotr-roslaniec (approved):
Very fine

Welcome on board

demos/taco-demo/src/App.tsx
```diff
@@ -89,7 +88,7 @@ export default function App() {
       provider,
       domain,
       encryptedMessage,
-      getPorterUri(domain),
+      undefined,
```
manumonti, line 91: We can get rid of this line, right?
```suggestion

```

manumonti (approved):
Great work here! 👏

I have some small suggestions, any of them is a blocker.

demos/taco-demo/src/App.tsx
vzotova: vs code shows error (and pnpm ci:lint) when I remove undefined, looks like it's stil using old interface with `porterUri?: string`. I'll keep looking

packages/shared/test/porter.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+  });
+};
+
+describe('PorterClient', () => {
+  beforeAll(async () => {
+    await initialize();
+  });
+
+  it('Get Ursulas', async () => {
```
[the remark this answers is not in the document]
vzotova: ✔️

packages/shared/test/porter.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  initialize,
+  GetUrsulasResult,
+  PorterClient,
+  toHexString,
+  Ursula,
+} from '../src';
+import { fakeUrsulas } from '@nucypher/test-utils';
+import { beforeAll, describe, expect, SpyInstance, it, vi } from 'vitest';
+import axios, { HttpStatusCode } from 'axios';
+
```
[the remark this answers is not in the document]
vzotova: ✔️

derekpierre:
@vzotova Can we hold off on merging this into `epic-auth` until after the sprint retrospective call today?

vzotova:
> @vzotova Can we hold off on merging this into `epic-auth` until after the sprint retrospective call today?

Of course

demos/taco-demo/src/App.tsx
piotr-roslaniec: We can only remove it if it's the last parameter in a function call - It will be implicitly set as `undefined`. Otherwise, we have to set it as `undefined` to preserve parameter ordering for further parameters.

manumonti: > We can only remove it if it's the last parameter in a function call - It will be implicitly set as `undefined`. Otherwise, we have to set it as `undefined` to preserve parameter ordering for further parameters.

Oh, got it.

derekpierre:
> > @vzotova Can we hold off on merging this into `epic-auth` until after the sprint retrospective call today?
> 
> Of course

Let's put this PR into a separate EPIC branch.

packages/shared/src/porter.ts
```diff
@@ -6,20 +6,24 @@ import {
...
 import qs from 'qs';
 
 import { Base64EncodedBytes, ChecksumAddress, HexEncodedBytes } from './types';
 import { fromBase64, fromHexString, toBase64, toHexString } from './utils';
 
-const porterUri: Record<string, string> = {
+const defaultPorterUri: Record<string, string> = {
   mainnet: 'https://porter.nucypher.community',
   tapir: 'https://porter-tapir.nucypher.community',
   oryx: 'https://porter-oryx.nucypher.community',
```
KPrasch, line 22: Oryx is fully deprecated so this line can be removed.

vzotova: ✔️

packages/shared/src/porter.ts
```diff
@@ -6,20 +6,23 @@ import {
...
 
 import { Base64EncodedBytes, ChecksumAddress, HexEncodedBytes } from './types';
 import { fromBase64, fromHexString, toBase64, toHexString } from './utils';
 
-const porterUri: Record<string, string> = {
+const defaultPorterUri: Record<string, string> = {
   mainnet: 'https://porter.nucypher.community',
   tapir: 'https://porter-tapir.nucypher.community',
-  oryx: 'https://porter-oryx.nucypher.community',
   lynx: 'https://porter-lynx.nucypher.community',
```
KPrasch, line 22: We're trying to deprecate the `.community` domain in favor of the `.io` domain.

```suggestion
  mainnet: 'https://porter.nucypher.io',
  tapir: 'https://porter-tapir.nucypher.io',
  lynx: 'https://porter-lynx.nucypher.io',
```

KPrasch (commented):
Does this PR intend to integrate with the porter URIs from the nucypher/nucypher-porter repo?

<https://raw.githubusercontent.com/nucypher/nucypher-porter/development/porter_instances.json>

KPrasch (approved):

theref (approved):

packages/shared/src/porter.ts
vzotova: addressed in #555

packages/shared/src/porter.ts
```diff
@@ -120,10 +132,39 @@ export type TacoDecryptResult = {
...
+      }
+      if (resp.status === HttpStatusCode.Ok) {
+        return resp;
+      }
+    }
+    if (lastError !== undefined) {
+      throw lastError;
+    }
+    throw new Error(
+      'Porter returns bad response: ${resp.status} - ${resp.data}',
```
derekpierre, line 166:
```suggestion
      'Porter returned bad response: ${resp.status} - ${resp.data}',
```

packages/taco/src/tdec.ts
```diff
@@ -61,7 +61,7 @@ export const encryptMessage = async (
 export const retrieveAndDecrypt = async (
   provider: ethers.providers.Provider,
   domain: Domain,
-  porterUri: string,
+  porterUris: string[],
```
derekpierre, line 64: I don't believe the functions in this module are exposed externally as part of the `taco-web` library. Would it be simpler to create the `PorterClient` object in the `taco.ts` module and pass that object in instead of the string array, or does that cause other issues?

vzotova: I'd need help with that, I'm not that familiar with taco-web to answer your questions.

packages/shared/src/porter.ts
vzotova: addressed in 22ff9ac67647b0fc957980f11b72fcbf167b1c2a

derekpierre (approved):
🎸

packages/taco/src/tdec.ts
derekpierre: Will be addressed vi #560 - https://github.com/nucypher/taco-web/pull/560#discussion_r1711843350.

PR #547, opened 2024-07-11
Reuse siwe authentication

manumonti:
**Type of PR:**

- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**
Add support to reuse SIWE messages on taco-web from other applications.

**Issues fixed/closed:**
> - Fixes #...

Related to https://github.com/nucypher/sprints/issues/22

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
Based on #546 .

PR #554, opened 2024-07-19
[WIP] API Rework

derekpierre:
**Type of PR:**

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Based over #547 

- Rework of decryption API to use `ConditionContext` instead of splitting up params for authProvider and customParameters.
- Allow `ConditionContext` to be populated by developers with authProviders and customParameters before being used for decryption
- Reduce parameters used in calls for decryption -> retrieve flow. Instead only obtain the ritual from the ritual id when the values are actually needed.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


packages/taco/src/conditions/context/context.ts
```diff
@@ -168,6 +173,38 @@ export class ConditionContext {
...
+      }
+    } else if (contextParam === USER_ADDRESS_PARAM_EXTERNAL_EIP4361) {
+      if (authProvider instanceof SingleSignOnEIP4361AuthProvider) {
+        this.authProviders[contextParam] = authProvider;
+        return;
+      }
+    }
+    throw new Error(
+      ERR_INVALID_AUTH_PROVIDER_TYPE(contextParam, typeof authProvider),
+    );
```
piotr-roslaniec, line 205: We can avoid nesting on line 192 if we throw a bit earlier, just like on line 186

packages/taco/src/taco.ts
```diff
@@ -202,6 +185,7 @@ export const isAuthorized = async (
     messageKit,
   );
 
+// TODO is this still valid and actually needed? should we remove this?
```
piotr-roslaniec, line 188: I think it's worthwhile to revisit it post release of `@nucypher/contracts` - IIRC, there were some changes to `EncryptorAllowlist` contract, etc.

So we may want to revisit how (if) to expose this logic in `taco-web`.

piotr-roslaniec (approved):
LGTM, please tag me for a review whenever

packages/taco/src/conditions/context/context.ts
derekpierre: ✔️

PR #555, opened 2024-07-25
GitHub Porter URIs

### LINKED ISSUE #320: Fallback chain of Porters

vzotova:
**Type of PR:**

- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
Fixes #320 

**Why it's needed:**

**Notes for reviewers:**
Based on #553

### DIFF
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -40,9 +40,8 @@ export const getPorterUri = async (domain: Domain): Promise<string> => {
 
 export const getPorterUris = async (
   domain: Domain,
-  porterUris: string[] = [],
 ): Promise<string[]> => {
-  const fullList = [...porterUris];
+  const fullList = [];
   const uri = defaultPorterUri[domain];
   if (!uri) {
     throw new Error(`No default Porter URI found for domain: ${domain}`);
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -187,7 +187,7 @@ export class PorterClient {
       throw lastError;
     }
     throw new Error(
-      'Porter returns bad response: ${resp.status} - ${resp.data}',
+      `Porter returns bad response: ${resp.status} - ${resp.data}`,
     );
   }
 
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -23,7 +23,7 @@ const defaultPorterUri: Record<string, string> = {
 };
 
 const porterUriSource: string =
-  'https://raw.githubusercontent.com/nucypher/nucypher-porter/development/porter_instances.json';
+  'https://raw.githubusercontent.com/nucypher/nucypher-porter/main/porter_instances.json';
 
 export type Domain = keyof typeof defaultPorterUri;
 export type PorterURISourceResponse = Record<string, string[]>;
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -26,6 +26,7 @@ const porterUriSource: string =
   'https://raw.githubusercontent.com/nucypher/nucypher-porter/development/porter_instances.json';
 
 export type Domain = keyof typeof defaultPorterUri;
+export type PorterURISourceResponse = Record<string, string[]>;
 
 export const domains: Record<string, Domain> = {
   DEVNET: 'lynx',
@@ -40,7 +41,7 @@ export const getPorterUri = async (domain: Domain): Promise<string> => {
 export const getPorterUris = async (
   domain: Domain,
   porterUris: string[] = [],
-): Promise<string[] => {
+): Promise<string[]> => {
   const fullList = [...porterUris];
   const uri = defaultPorterUri[domain];
   if (!uri) {
@@ -52,16 +53,19 @@ export const getPorterUris = async (
   return fullList;
 };
 
-export const getPorterUrisFromSource = async (domain: Domain): Promise<string[]> => {
-  const source = porterUriSource[domain];
+export const getPorterUrisFromSource = async (
+  domain: Domain,
+): Promise<string[]> => {
+  const source = porterUriSource;
   if (!source) {
     return [];
   }
   try {
-    const resp = await axios.get(source, {
+    const resp = await axios.get(porterUriSource, {
       responseType: 'blob',
     });
-    return resp.data.split(/\r?\n/).filter(Boolean);
+    const uris: PorterURISourceResponse = JSON.parse(resp.data);
+    return uris[domain];
   } catch (e) {
     return [];
   }
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -17,13 +17,13 @@ import { Base64EncodedBytes, ChecksumAddress, HexEncodedBytes } from './types';
 import { fromBase64, fromHexString, toBase64, toHexString } from './utils';
 
 const defaultPorterUri: Record<string, string> = {
-  mainnet: 'https://porter.nucypher.community',
-  tapir: 'https://porter-tapir.nucypher.community',
-  lynx: 'https://porter-lynx.nucypher.community',
+  mainnet: 'https://porter.nucypher.io',
+  tapir: 'https://porter-tapir.nucypher.io',
+  lynx: 'https://porter-lynx.nucypher.io',
 };
 
 const porterUriSource: string =
-  'https://raw.githubusercontent.com/nucypher/nucypher-porter/main/porter_instances.json';
+  'https://raw.githubusercontent.com/nucypher/nucypher-porter/development/porter_instances.json';
 
 export type Domain = keyof typeof defaultPorterUri;
 
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -175,15 +175,15 @@ export class PorterClient {
       const localConfig = { ...config, baseURL: porterUrl.toString() };
       try {
         resp = await axios.request(localConfig);
+        if (resp.status === HttpStatusCode.Ok) {
+          return resp;
+        }
       } catch (e) {
         lastError = e;
         continue;
       }
-      if (resp.status === HttpStatusCode.Ok) {
-        return resp;
-      }
     }
-    if (lastError !== undefined) {
+    if (lastError) {
       throw lastError;
     }
     throw new Error(
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -22,6 +22,12 @@ const defaultPorterUri: Record<string, string> = {
   lynx: 'https://porter-lynx.nucypher.community',
 };
 
+const porterUriSource: Record<string, string> = {
+  mainnet: 'https://raw.githubusercontent.com/nucypher/nucypher-porter/main/pytest.ini',
+  tapir: 'https://raw.githubusercontent.com/nucypher/nucypher-porter/main/pytest.ini',
+  lynx: 'https://raw.githubusercontent.com/nucypher/nucypher-porter/main/pytest.ini',
+};
+
 export type Domain = keyof typeof defaultPorterUri;
 
 export const domains: Record<string, Domain> = {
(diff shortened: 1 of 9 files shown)


packages/shared/src/porter.ts
```diff
@@ -23,6 +23,12 @@ const defaultPorterUri: Record<string, string> = {
   lynx: 'https://porter-lynx.nucypher.community',
 };
 
+const porterUriSource: Record<string, string> = {
+  mainnet: 'https://raw.githubusercontent.com/nucypher/nucypher-porter/main/pytest.ini',
+  tapir: 'https://raw.githubusercontent.com/nucypher/nucypher-porter/main/pytest.ini',
+  lynx: 'https://raw.githubusercontent.com/nucypher/nucypher-porter/main/pytest.ini',
```
piotr-roslaniec, line 29: WDYT about using one JSON file keyed by a network?

packages/shared/src/porter.ts
```diff
@@ -31,20 +37,37 @@ export const domains: Record<string, Domain> = {
...
+export const getPorterUrisFromSource = async (domain: Domain): Promise<string[]> => {
+  const source = porterUriSource[domain];
+  if (!source) {
+    return [];
+  }
+  try {
+    const resp = await axios.get(source, {
+      responseType: 'blob',
+    });
+    return resp.data.split(/\r?\n/).filter(Boolean);
```
piotr-roslaniec, line 65: IMHO using a JSON file here is easier to access and deserialize (and we can even add types for the resulting data structure).

packages/shared/src/porter.ts
vzotova: ✔️

packages/shared/src/porter.ts
vzotova: ✔️

vzotova:
@piotr-roslaniec @cygnusv @theref To finish that PR I'd like to ask you where JSON should be stored? It can be empty for the beginning

piotr-roslaniec:
IMHO https://github.com/nucypher/nucypher-porter repo is fine for this purpose

packages/shared/src/porter.ts
```diff
@@ -6,33 +6,71 @@ import {
...
 import qs from 'qs';
 
 import { Base64EncodedBytes, ChecksumAddress, HexEncodedBytes } from './types';
 import { fromBase64, fromHexString, toBase64, toHexString } from './utils';
 
-const porterUri: Record<string, string> = {
+const defaultPorterUri: Record<string, string> = {
   mainnet: 'https://porter.nucypher.community',
   tapir: 'https://porter-tapir.nucypher.community',
   oryx: 'https://porter-oryx.nucypher.community',
```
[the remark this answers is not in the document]
piotr-roslaniec: Related: https://github.com/nucypher/taco-web/pull/542

packages/shared/src/porter.ts
```diff
@@ -6,33 +6,69 @@ import {
...
   MAINNET: 'mainnet',
 };
 
-export const getPorterUri = (domain: Domain): string => {
-  const uri = porterUri[domain];
+export const getPorterUri = async (domain: Domain): Promise<string> => {
+  return (await getPorterUris(domain))[0];
+};
+
+export const getPorterUris = async (
```
piotr-roslaniec, line 41: This function is guaranteed to never return an empty array, right? WDYT about adding an `if` statement and throwing an error before we return, just to make sure?

packages/shared/src/porter.ts
```diff
@@ -6,33 +6,69 @@ import {
...
-  lynx: 'https://porter-lynx.nucypher.community',
+const defaultPorterUri: Record<string, string> = {
+  mainnet: 'https://porter.nucypher.io',
+  tapir: 'https://porter-tapir.nucypher.io',
+  lynx: 'https://porter-lynx.nucypher.io',
 };
 
-export type Domain = keyof typeof porterUri;
+const porterUriSource: string =
+  'https://raw.githubusercontent.com/nucypher/nucypher-porter/development/porter_instances.json';
```
piotr-roslaniec, line 26: `development` is the default branch, should we point to `main` instead?

packages/shared/src/porter.ts
```diff
@@ -120,10 +156,39 @@ export type TacoDecryptResult = {
...
+      }
+      if (resp.status === HttpStatusCode.Ok) {
+        return resp;
+      }
+    }
+    if (lastError !== undefined) {
+      throw lastError;
+    }
+    throw new Error(
+      'Porter returns bad response: ${resp.status} - ${resp.data}',
```
piotr-roslaniec, line 190: Do we ever test for throwing this error?

```suggestion
      `Porter returns bad response: ${resp.status} - ${resp.data}`,
```

vzotova: Yes, there is test in porter.test `returns error in case all porters fail`

packages/shared/src/porter.ts
```diff
@@ -120,10 +156,39 @@ export type TacoDecryptResult = {
...
+      this.porterUrls = porterUris.map((uri) => new URL(uri));
+    } else {
+      this.porterUrls = [new URL(porterUris)];
+    }
+  }
+
+  protected async tryAndCall<T, D>(
+    config: AxiosRequestConfig<D>,
+  ): Promise<AxiosResponse<T>> {
+    let resp!: AxiosResponse<T>;
```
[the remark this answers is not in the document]
vzotova: to mark it's 100% assigned, vs code was showing error without it

packages/shared/src/porter.ts
vzotova: @KPrasch merged list into the development branch

packages/shared/src/porter.ts
vzotova: I think it will be overkill, there is already error in case there is no default URI so it feels too much

packages/shared/src/porter.ts
piotr-roslaniec: Ok, so we didn't catch this issue (template string misuse) because `returns error in case all porters fail` doesn't test for the error message in the `Error` we've thrown

packages/shared/src/porter.ts
```diff
@@ -120,10 +156,39 @@ export type TacoDecryptResult = {
...
+        resp = await axios.request(localConfig);
+      } catch (e) {
+        lastError = e;
+        continue;
+      }
+      if (resp.status === HttpStatusCode.Ok) {
+        return resp;
+      }
+    }
+    if (lastError !== undefined) {
```
piotr-roslaniec, line 186:
```suggestion
    if (lastError) {
```

piotr-roslaniec:
@vzotova Before we merge this, it may be helpful to run the Node.js example using these new Porter URLs

packages/shared/src/porter.ts
vzotova: ✔️

packages/shared/src/porter.ts
```diff
@@ -120,10 +156,39 @@ export type TacoDecryptResult = {
...
+      const localConfig = { ...config, baseURL: porterUrl.toString() };
+      try {
+        resp = await axios.request(localConfig);
+      } catch (e) {
+        lastError = e;
+        continue;
+      }
+      if (resp.status === HttpStatusCode.Ok) {
+        return resp;
+      }
```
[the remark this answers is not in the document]
vzotova: ✔️

packages/shared/src/porter.ts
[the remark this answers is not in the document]
piotr-roslaniec: It's a minor smell so I guess we let it slide. But at any rate, `let` statements where we assign value later should not be required to `!` if the branching is done properly

packages/shared/src/porter.ts
piotr-roslaniec: @KPrasch I think it'd be less confusing if we used `main`. Just my 2 cents.

packages/taco/src/taco.ts
```diff
@@ -148,7 +148,7 @@ export const decrypt = async (
   porterUris: string[] = [],
   customParameters?: Record<string, CustomContextParam>,
 ): Promise<Uint8Array> => {
-  const porterUrisFull: string[] = getPorterUris(domain, porterUris);
+  const porterUrisFull: string[] = await getPorterUris(domain, porterUris);
```
derekpierre, line 151: Wondering what the spirit of the API is if the caller already provides its list (1 or more) of porter URIs they want to use. 

In `getPorterUris` it seems we always append to the list, but perhaps we shouldn't if the user already provides a value. Obviously, since we are appending to a list, we still giving preference to their URIs first, but is it possible that the caller only wants to use specific Porters (for some reason 🤷 )?  This is not an option at the moment i.e. there is no ability for the user to actually override of whatever defaults we deem are good - only the ability to bias to their value(s) but not circumvent the use of our defaults if necessary. Don't get me wrong, in most cases I see devs not providing their own URIs but rather relying on our chain of URIs, but from an API perspective, it feels a little limiting.

So what if :
1. caller does not provide any porter Uris -> we get the chain of uris that includes defaults, and whatever is in the github repos
2. caller provides 1 or more porter Uris - we don't append and only use the URIs provided by the caller.

The `getPorterUris` call is available to the user as part of the `nucypher/shared` API so it can be used directly before calling `decrypt` if they desire. Not blindly appending may also save us from potential duplicate porter URIs if the user already provides existing default URIs directly.

Thoughts?

packages/shared/src/porter.ts
```diff
@@ -17,36 +17,60 @@ import { Base64EncodedBytes, ChecksumAddress, HexEncodedBytes } from './types';
...
-  mainnet: 'https://porter.nucypher.community',
-  tapir: 'https://porter-tapir.nucypher.community',
-  lynx: 'https://porter-lynx.nucypher.community',
+  mainnet: 'https://porter.nucypher.io',
+  tapir: 'https://porter-tapir.nucypher.io',
+  lynx: 'https://porter-lynx.nucypher.io',
 };
 
+const porterUriSource: string =
+  'https://raw.githubusercontent.com/nucypher/nucypher-porter/development/porter_instances.json';
```
derekpierre, line 26: Currently, this file contains the same values as the defaults...should the file in the repos be an empty dict for the time being, and we add to it over time, if there are other non-default porter URIs that can be used? i.e. the functionality is there if non-default porter URIs become available to use.

packages/taco/src/taco.ts
vzotova: from API point of view - I'd prefer append always to minimize amount of code for user. Other thought - it's possible to use `PorterClient` directly, in that case user can specify anything.

packages/shared/src/porter.ts
vzotova: I thought maybe file in repo will be empty, but with second thought now - maybe just check array for duplicates?

packages/shared/src/porter.ts
derekpierre: Now that https://github.com/nucypher/nucypher-porter/pull/72 is merged, this is now available on the `main` branch.

vzotova: ✔️

packages/taco/src/taco.ts
cygnusv: I think the API behaviour I'd expect is that either the library uses the user-provided input, or it uses the defaults. While appending to the list of defaults is a possible option, I don't think it's very natural from a dev experience point of view.

vzotova: sure, changed api
✔️

KPrasch (approved):

packages/shared/src/porter.ts
derekpierre: Any reason you don't want the repos file to be empty?

vzotova: No reasons, I'm totally fine with that

derekpierre: Cool. Let's update the file in `nucypher-porter` to be empty then. If there are other non-default porter URIs we can use, we can add them at that time. 

I assume the code in this PR can handle an empty dict.

derekpierre (approved):
🎸

packages/shared/src/porter.ts
derekpierre: Followed-up in https://github.com/nucypher/nucypher-porter/pull/76

packages/taco/README.md
```diff
@@ -60,7 +60,7 @@ const decryptedMessage = await decrypt(
   web3Provider,
   domains.TESTNET,
   messageKit,
-  getPorterUri(domains.TESTNET),
+  await getPorterUri(domains.TESTNET),
```
derekpierre, line 63: Would this work - `getPorterUri` only returns a `string` and not `string[]`. Or does typescript automatically handle that gracefully?

vzotova: would it be ok to remove that line? basically it will show how to use default URIs

derekpierre: 👍 - sounds reasonable to me.

PR #560, opened 2024-07-30
Decrypt API Rework

derekpierre:
**Type of PR:**

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
- Rework of decryption API to use `ConditionContext` instead of splitting up params for authProvider and customParameters.
- Allow `ConditionContext` to be populated by developers with authProviders and customParameters before being used for decryption
- Reduce parameters used in calls for decryption -> retrieve flow. Instead only obtain the ritual from the ritual id when the values are actually needed.

The updated use of the code will look like the following:

1. Create condition context from message kit:
```typescript
const conditionContext = conditions.context.ConditionContext.fromMessageKit(messageKit);
```

2. Add any required auth provider(s):
    * Add auth provider for `:userAddress`, if necessary:
    ```typescript
    const authProvider = new EIP4361AuthProvider(web3Provider, web3Provider.getSigner());
    conditionContext.addAuthProvider(USER_ADDRESS_PARAM_DEFAULT, authProvider);
    ```

    * Add auth provider for `:userAddressExternalEIP4361`, if necessary:
    ```typescript
    const singleSignOnEIP4361AuthProvider = await SingleSignOnEIP4361AuthProvider.fromExistingSiweInfo(messageStr, 
    signature);
    conditionContext.addAuthProvider(USER_ADDRESS_PARAM_EXTERNAL_EIP4361, authProvider);
    ```

3. Add any required custom context variables, if necessary:
```typescript
const conditionContext = conditions.context.ConditionContext.fromMessageKit(messageKit);
conditionContext.addCustomContextParameterValues({
    ":myCustomValue1": value_1,
    ":myCustomVallue2": value_2,
    ...
});
```

4. Call decrypt
```typescript
const decryptedMessage = await decrypt(web3Provider, domain, messageKit, conditionContext);
```

---

Additionally, developers can also better query what parameters are needed for a condition in a message kit. This helps when apps have various permutations of conditions that they use, which include different parameters:
```typescript
const conditionContext = conditions.context.ConditionContext.fromMessageKit(messageKit);
requestedContextParameters = conditionContext.requestedContextParameters  // set of context variables required for underlying condition
// Devs can check for specific context parameters or loop over the set and populate context variables appropriately.
```

**Issues fixed/closed:**
> - Fixes #...

- Closes https://github.com/nucypher/sprints/issues/44

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/examples/taco/nodejs/src/index.ts
+++ b/examples/taco/nodejs/src/index.ts
@@ -11,7 +11,6 @@ import {
   fromBytes,
   initialize,
   isAuthorized,
-  ThresholdMessageKit,
   toBytes,
   toHexString,
 } from '@nucypher/taco';
--- a/examples/taco/nodejs/src/index.ts
+++ b/examples/taco/nodejs/src/index.ts
@@ -117,12 +117,7 @@ const decryptFromBytes = async (encryptedBytes: Uint8Array) => {
   const conditionContext =
     conditions.context.ConditionContext.fromMessageKit(messageKit);
   conditionContext.addAuthProvider(USER_ADDRESS_PARAM_DEFAULT, authProvider);
-  return decrypt(
-    provider,
-    domain,
-    messageKit,
-    conditionContext,
-  );
+  return decrypt(provider, domain, messageKit, conditionContext);
 };
 
 const runExample = async () => {
--- a/examples/taco/nodejs/src/index.ts
+++ b/examples/taco/nodejs/src/index.ts
@@ -1,11 +1,12 @@
 import { format } from 'node:util';
 
 import {
+  EIP4361AuthProvider,
+  ThresholdMessageKit,
+  USER_ADDRESS_PARAM_DEFAULT,
   conditions,
   decrypt,
   domains,
-  EIP4361AuthProvider,
-  USER_ADDRESS_PARAM_DEFAULT,
   encrypt,
   fromBytes,
   initialize,
--- a/examples/taco/nodejs/src/index.ts
+++ b/examples/taco/nodejs/src/index.ts
@@ -109,14 +109,21 @@ const decryptFromBytes = async (encryptedBytes: Uint8Array) => {
     domain: 'localhost',
     uri: 'http://localhost:3000',
   };
-  const authProvider = new EIP4361AuthProvider(
-    provider,
-    consumerSigner,
-    siweParams,
-  );
   const conditionContext =
     conditions.context.ConditionContext.fromMessageKit(messageKit);
-  conditionContext.addAuthProvider(USER_ADDRESS_PARAM_DEFAULT, authProvider);
+
+  // illustrative optional example of checking what context parameters are required
+  // unnecessary if you already know what the condition contains
+  if (
+    conditionContext.requestedContextParameters.has(USER_ADDRESS_PARAM_DEFAULT)
+  ) {
+    const authProvider = new EIP4361AuthProvider(
+      provider,
+      consumerSigner,
+      siweParams,
+    );
+    conditionContext.addAuthProvider(USER_ADDRESS_PARAM_DEFAULT, authProvider);
+  }
   return decrypt(provider, domain, messageKit, conditionContext);
 };
 
--- a/examples/taco/nodejs/src/index.ts
+++ b/examples/taco/nodejs/src/index.ts
@@ -5,6 +5,7 @@ import {
   decrypt,
   domains,
   EIP4361AuthProvider,
+  USER_ADDRESS_PARAM_DEFAULT,
   encrypt,
   fromBytes,
   initialize,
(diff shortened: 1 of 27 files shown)


packages/taco/src/taco.ts
```diff
@@ -144,9 +137,8 @@ export const decrypt = async (
   provider: ethers.providers.Provider,
   domain: Domain,
   messageKit: ThresholdMessageKit,
-  authProvider?: EIP4361AuthProvider,
+  context?: ConditionContext,
   porterUri?: string,
```
derekpierre, line 141: Note to self - implement suggestion here - https://github.com/nucypher/taco-web/pull/553#discussion_r1711397108 - before merging.

theref (approved):
> ```
> requestedContextParameters = conditionContext.requestedContextParameters  // list of context variables required for underlying condition
> // Devs can loop over the list and populate context variables appropriately.
> ```

Could we get an example of iterating over this?

Otherwise it looks great!!!

derekpierre:
> > ```
> > requestedContextParameters = conditionContext.requestedContextParameters  // list of context variables required for underlying condition
> > // Devs can loop over the list and populate context variables appropriately.
> > ```
> 
> Could we get an example of iterating over this?

Great idea - added an illustrative optional example to the nodejs example.

examples/taco/nodejs/src/index.ts
```diff
@@ -109,14 +109,21 @@ const decryptFromBytes = async (encryptedBytes: Uint8Array) => {
...
-    provider,
-    consumerSigner,
-    siweParams,
-  );
   const conditionContext =
     conditions.context.ConditionContext.fromMessageKit(messageKit);
-  conditionContext.addAuthProvider(USER_ADDRESS_PARAM_DEFAULT, authProvider);
+
+  // illustrative optional example of checking what context parameters are required
+  // unnecessary if you already know what the condition contains
```
theref, line 116: beautiful! 🚀

examples/taco/nodejs/src/index.ts
```diff
@@ -108,17 +109,22 @@ const decryptFromBytes = async (encryptedBytes: Uint8Array) => {
...
+    conditionContext.requestedContextParameters.has(USER_ADDRESS_PARAM_DEFAULT)
+  ) {
+    const authProvider = new EIP4361AuthProvider(
+      provider,
+      consumerSigner,
+      siweParams,
+    );
+    conditionContext.addAuthProvider(USER_ADDRESS_PARAM_DEFAULT, authProvider);
+  }
+  return decrypt(provider, domain, messageKit, conditionContext);
```
arjunhassard, line 127: So this avoids having to hardcode context parameters – i.e. the app doesn't need to know the conditions in advance in order to match them to the right authentication method? That's pretty amazing! 

One thing I'm wondering is – what resource is being queried here? Do we need to create a public resource that maps conditions to context parameters? Or is this a list that developers will create for their own domain?

arjunhassard (approved):
Once we offer lots of authentication methods and broader conditionality, this structural improvement will be invaluable to some use cases!

examples/taco/nodejs/src/index.ts
derekpierre: > So this avoids having to hardcode context parameters – i.e. the app doesn't need to know the conditions in advance in order to match them to the right authentication method? That's pretty amazing!

Indeed. This is the case where an application doesn't just use the same condition every time, but perhaps some permutations of a set of conditions, each with varying context parameters. In this way the dev can use logic to get the set of expected parameters and populate them based on their own logic.

They could have dynamic logic like the following:
```
if (requestedContextParmeters.has(":tokenId")) {
   // user subjected to condition 1 which uses ":tokenId" custom context parameter
   const value = ...
   conditionContext.addCustomContextParameters({":tokenId": value});
}
if (requestedContextParmeters.has(":otherParam")) {
   // user subjected to condition 2 which uses ":otherParam" custom context parameter
   const otherValue = ...
   conditionContext.addCustomContextParameters({":otherParam": otherValue});
}
...
```

> One thing I'm wondering is – what resource is being queried here? 

Assuming I correctly understand the question, the `ConditionContext` looks at the underlying condition in the `ThresholdMessageKit` (encrypted data), and then returns all the context parameters in the underlying condition (`requestedContextParameters` property). It is then up to the app to populate the values appropriately. *How* they do that is up to them - we are not being prescriptive.

> Do we need to create a public resource that maps conditions to context parameters? Or is this a list that developers will create for their own domain?

Interesting. To me, it's the latter. Devs can decide how they want to get/generate the values to use for conditions.  
IMHO, on the decryption side, we should endeavour for apps not to have to deal with the underlying condition object(string) themselves. They can if they want to, of course, but they shouldn't really have to. It's part of the reason I made the `ConditionContext` object take a `ThresholdMessageKit` and not the condition object contained within the `ThresholdMessageKit`.

With something like a public mapping resource (condition string/type -> context parameters...?), it may be easy for simple conditions eg. `BalanceOf`, or `OwnsNFT`, but it's tough cover all cases with a mapping because our conditions are extensible, which is intentional for breadth of functionality. Devs can use whatever custom context parameters they want, and they can also use `:userAddressExternalEIP4361` (single sign on SIWE) OR `:userAddress` (taco SIWE) if they like...it would be tough to cover all cases.  Even with the mapping, they'll still need to know/determine what value to use for the context parameters - so they are kind of back to something like the 'if' statements above.

vzotova (approved):

manumonti (approved):
Great work here! 🙌

cygnusv (approved):

Issue #104, opened 2022-11-07
ENS name ownership condition


theref:

arjunhassard:
- [ ] Use existing RPC conditions to build off-the-shelf condition set for decryptor to control ENS name(s) 
- [ ] Write up walkthrough tutorial and add to [Advanced Usage Guides](https://docs.threshold.network/app-development/threshold-access-control-tac/advanced-usage) or [Conditions](https://docs.threshold.network/app-development/threshold-access-control-tac/conditions)
- [ ] Edit tutorial (@arjunhassard)

derekpierre:
Is the idea for this issue only to address name resolution on the client-side (i.e. at encryption time, just use the resolved wallet address in place of the ens name), or is there an associated server-side change that is expected.

If the idea is only to do name resolution at the client side (i.e. at encryption time), then one limitation is that it would prevent any updating of the ens domain to point to a different wallet address after the fact. 

Obviously, changes on the server side require updates across the network, but aside from that, the actual code change is probably not that difficult - https://web3py.readthedocs.io/en/stable/ens_overview.html#ethereum-name-service-ens.

piotr-roslaniec:
I think there are two different functionalities, that we may or may not want to bundle together:
- ENS-to-address resolution on the client side. "When I input `bob.eth`, I actually mean `0xb0b...`"
- ENS address-ownership condition. "Whoever owns `bob.eth` should be able to decrypt".

manumonti:
> ENS address-ownership condition. "Whoever owns bob.eth should be able to decrypt".

AFAIK this is the goal of this issue.

I have created a new issue for the other idea: https://github.com/nucypher/taco-web/issues/512

manumonti:
After a chat about this with @piotr-roslaniec, we came to some conclusions:

### There are different use cases:
As Piotr realized, there are at least two different use cases that will require ENS:

- The "*only if your address resolves to this ENS name*" use case: we want to encrypt some information so the decryptor will be able to decrypt only if their address resolves to a specific ENS name.

- The "*lottery winner*" use case: we want to encrypt some information so only the lottery winner can decrypt it. The lottery is celebrated every day and, as a result, the winner's address is automatically registered as "`lotterywinner.eth`" during this day. So, the condition is "only the address which the domain `lotterywinner.eth` resolves to will be able to decrypt the information". As we can see, this address will be different every day.


### We cannot use the current TACo conditions, but there are some alternatives
[...]

PR #564, opened 2024-08-08
Added condition ENS ownership

### LINKED ISSUE #512: ENS address name resolution
Is the idea for this issue only to address name resolution on the client-side (i.e. at encryption time, just use the resolved wallet address in place of the ens name), or is there an associated server-side change that is expected.

If the idea is only to do name resolution at the client side (i.e. at encryption time), then one limitation is that it would prevent any updating of the ens domain to point to a different wallet address after the fact. 

Obviously, changes on the server side require updates across the network, but aside from that, the actual code change is probably not that difficult - https://web3py.readthedocs.io/en/stable/ens_overview.html#ethereum-name-service-ens.

_Originally posted by @derekpierre in https://github.com/nucypher/taco-web/issues/104#issuecomment-2066898049_

andresceballosm:
**Type of PR:**

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**
- This proposal suggests adding a condition for ENS on the client side. The client will send a structure to the node as shown below. This structure will enable the validation of whether the signer in the decryption process is the owner of the ENS with the resolved address (validate that signer === resolvedAddress).
```
  {
  "schema": {},
  "value": {
    "conditionType": "ensOwnership",
    "domain": "tapir",
    "ensName": "andrestest.eth",
    "resolvedAddress": "0xf45ED59ea03AaA29EAB6DB0Ea019E5299b335Ea2"
  }
}
```

**Issues fixed/closed:**

> - Fixes #512 

**Why it's needed:**

**Notes for reviewers:**
- ENS resolve only works in mainnet and Sepolia
- to call the condition
```
     const ensCondition = new conditions.ensOwnership.EnsAddressOwnershipCondition({
        ensName: 'andrestest.eth',
        signer,
        domain
      });
```


theref:
This is a great start, thank you! I've got 2 comments :)

1. With the current design, you're converting the ens name to an address using `viem`. This happens at condition creation time. The problem here is that ens ownership can change, and the condition won't stay in line with the ownership.

2. This implementation would also require us to build an equivalent `ensOwnership` condition in `nucypher/nucypher`. Would it be possible to have `ensOwnership` be a wrapper of `ContractCondition`?

The condition is still defined exactly as you specified:
```
     const ensCondition = new conditions.ensOwnership.EnsAddressOwnershipCondition({
        ensName: 'andrestest.eth',
        signer,
        domain
      });
```
but behind the scenes it would create a contract condition that reads from the ens registry. This would fix both of my comments. @KPrasch thoughts?

piotr-roslaniec:
Please see this issue comment https://github.com/nucypher/taco-web/issues/104#issuecomment-2079254236 for the context on ENS resolution. TLDR, it's a multi-step process that requires some development in `nucypher/nucypher` and we can't support it out-of-the-box using `ContractConditions` at the moment. Maybe with sequential conditions?

PR #567, opened 2024-08-19
Prevent perpetual use of  expired auth signature; remove superfluous API functions

### LINKED ISSUE #324: Expose `registerEncrypters` from `taco` API

derekpierre:
**Type of PR:**

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**

Based over #560 (somewhat of a follow-up PR).

- `EIP4361AuthProvider` caches the auth signature it first generates perpetually (unless storage is explicitly cleared eg. `localStorage.clear()` in apps). However, this auth signature can become expired (> 2h old), and so the provider should detect that on cache retrieval and regenerate the auth signature if it becomes expired.
- Add test for `SingleSignOnEIP4361AuthProvider`
- Clean up tests so that signer is obtained from provider, instead of creating separate `fakeProvider()` and `fakeSigner()`; this way they are better linked for tests. This only applies to tests - in real life it is possible to have a signer with no provider, but this was not the case for tests prior to this change anyway.
- Remove no longer applicable/superfluous API calls from being exposed:
   - `isAuthorized(...)` - calls into the `Coordinator` contract to call this function, but we are moving away from this where calls should be made directly to the `AccessController`. However the accessController only checks for `isEncryptionAuthorized` and doesn't provide the ability to check if an encrypted address is authorized - which is what is implemented here. Therefore, it only applies to the GlobalAllowList contracts, but it's unclear if we need that ability in `taco-web`
   - `registerEncrypters(...)` - Was never publicly exposed and it is unclear if we need this ability in `taco-web`. #324 alluded to it, but not sure if we still care. Let me know if we do (cc @arjunhassard ).
   
   *If we decide down the road we need them, we can add them back.*

**Issues fixed/closed:**

> - Fixes #...

Closes https://github.com/nucypher/taco-web/issues/324

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/examples/taco/nodejs/src/index.ts
+++ b/examples/taco/nodejs/src/index.ts
@@ -10,7 +10,6 @@ import {
   encrypt,
   fromBytes,
   initialize,
-  isAuthorized,
   toBytes,
   toHexString,
 } from '@nucypher/taco';
@@ -80,18 +79,6 @@ const encryptToBytes = async (messageString: string) => {
     ritualId,
     encryptorSigner,
   );
-
-  // Note: Not actually needed but used by CI for checking contract compatibility.
-  // Calling it after the encryption because we need material from messageKit.
-  const isEncryptorAuthenticated = await isAuthorized(
-    provider,
-    domain,
-    messageKit,
-    ritualId,
-  );
-  if (!isEncryptorAuthenticated) {
-    throw new Error('Not authorized');
-  }
 
   return messageKit.toBytes();
 };
--- a/packages/taco-auth/src/providers/eip4361/eip4361.ts
+++ b/packages/taco-auth/src/providers/eip4361/eip4361.ts
@@ -52,13 +52,32 @@ export class EIP4361AuthProvider {
     // If we have a signature in localStorage, return it
     const maybeSignature = this.storage.getAuthSignature(storageKey);
     if (maybeSignature) {
-      return maybeSignature;
+      // check whether older than node freshness requirement
+      if (this.isMessageExpired(maybeSignature.typedData)) {
+        // clear signature so that it will be recreated and stored
+        this.storage.clear(storageKey);
+      } else {
+        return maybeSignature;
+      }
     }
 
     // If at this point we didn't return, we need to create a new message
     const authMessage = await this.createSIWEAuthMessage();
     this.storage.setAuthSignature(storageKey, authMessage);
     return authMessage;
+  }
+
+  private isMessageExpired(message: string): boolean {
+    const siweMessage = new SiweMessage(message);
+    if (!siweMessage.issuedAt) {
+      // don't expect to ever happen; but just in case
+      return false;
+    }
+
+    const twoHourWindow = new Date(siweMessage.issuedAt);
+    twoHourWindow.setHours(twoHourWindow.getHours() + 2);
+    const now = new Date();
+    return twoHourWindow < now;
   }
 
   private async createSIWEAuthMessage(): Promise<AuthSignature> {
--- a/packages/pre/test/acceptance/alice-grants.test.ts
+++ b/packages/pre/test/acceptance/alice-grants.test.ts
@@ -10,7 +10,6 @@ import {
   bytesEqual,
   fakePorterUri,
   fakeProvider,
-  fakeSigner,
   fakeUrsulas,
   fromBytes,
   mockGetUrsulas,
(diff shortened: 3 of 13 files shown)


examples/taco/nodejs/src/index.ts
```diff
@@ -81,18 +80,6 @@ const encryptToBytes = async (messageString: string) => {
     encryptorSigner,
   );
```
manumonti, line 82: Are we still "checking contract compatibility" even when removing this from the Testnet examples that are run every hour?

https://github.com/nucypher/taco-web/actions/workflows/lynx.yml

manumonti (approved):
LGTM! I've just left a concern I have.

examples/taco/nodejs/src/index.ts
derekpierre: At the moment, I don't think `isAuthorized` is useful as part of the exposed `taco` API, hence the removal here. If others feel that it is useful I can put it back.

It's unclear to me how comprehensive this check is as it pertains to "checking compatibility" based on one function.  Do you have any additional insight there, perhaps I'm missing something?  Perhaps testing the `DkgCoordinatorAgent` with `lynx` contracts could be a better solution to CI checking compatibility...maybe - not sure what that entails or if it is feasible?

Additionally, `Coordinator.isEncryptionAuthorized()` which is called by `isAuthorized()`, is actually deprecated in the latest Coordinator contract, and will be removed down the road. Instead, the underlying acccessController for the ritual Id should be called directly for that check.

manumonti: No, I don't have a clear idea about this check beyond that we can check that this specific method on the Coordinator contract has not been modified. 

So yes, I agree we can delete this.

My concern came after I read the commented lines in the code talking about "checking contract compatibility": I was wondering if I'm missing something here.

vzotova (approved):

packages/taco-auth/src/providers/eip4361/eip4361.ts
```diff
@@ -61,6 +67,19 @@ export class EIP4361AuthProvider {
     return authMessage;
   }
 
+  private isMessageExpired(message: string): boolean {
+    const siweMessage = new SiweMessage(message);
+    if (!siweMessage.issuedAt) {
+      // don't expect to ever happen; but just in case
+      return false;
+    }
```
KPrasch, line 75: This is a sentinel against `undefined`?

KPrasch (approved):
Overall looks good 👍🏻 While I do find this acceptable (and very convenient for implementers) it's interesting to observe the lines between application and library logic blurring a bit here.

packages/taco-auth/src/providers/eip4361/eip4361.ts
derekpierre: Yes.

From the EIP4361 spec,`issued-at` is a required value - https://docs.login.xyz/general-information/siwe-overview/eip-4361#specification.

However, in the `siwe` typescript library implementation, it is an optional value in the constructor - https://github.com/spruceid/siwe/blob/main/packages/siwe/lib/client.ts#L49. As a result, Typescript complains if we don't check whether it is defined first before using it.

derekpierre:
> Overall looks good 👍🏻 While I do find this acceptable (and very convenient for implementers) it's interesting to observe the lines between application and library logic blurring a bit here.

Because we opted to cache the message ourselves to prevent users having to sign every time for proof of wallet ownership, it forces us to manage the cache appropriately - this PR does the latter. Definitely open to ideas here if you have any.

packages/taco-auth/src/providers/eip4361/eip4361.ts
cygnusv: Is this a divergence of `siwe` wrt to the spec then?

cygnusv (approved):

packages/taco-auth/src/providers/eip4361/eip4361.ts
derekpierre: As far as I can tell, yes, it is a divergence.

derekpierre: Well, it's not the constructor but the SiweMessage struct - so still a deviation - https://github.com/spruceid/siwe/blob/main/packages/siwe/lib/client.ts#L49

Without the check, the compiler returns the following:
```
packages/taco-auth/src/providers/eip4361/eip4361.ts:72:36 - error TS2769: No overload matches this call.
  Overload 1 of 4, '(value: string | number | Date): Date', gave the following error.
    Argument of type 'string | undefined' is not assignable to parameter of type 'string | number | Date'.
      Type 'undefined' is not assignable to type 'string | number | Date'.
  Overload 2 of 4, '(value: string | number): Date', gave the following error.
    Argument of type 'string | undefined' is not assignable to parameter of type 'string | number'.
      Type 'undefined' is not assignable to type 'string | number'.
 ```
 
 I'll file an issue for the `siwe` library, and perhaps they can either clarify or fix.

PR #591, opened 2024-10-04
chore: separate schemas from conditions to allow better combinations/reuse of schemas

derekpierre:
**Type of PR:**

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Currently, condition schemas and condition objects are mixed in the same module. This can generate a circular dependency when trying to reuse schemas for other condition schemas.

Instead, we should separate schemas from conditions so that the schemas can be obtained from a standalone module, increasing their use with other condition schemas.

This change allows us to have a `anyConditionSchema` that includes all condition schemas where any condition can be re-used eg. in `CompoundCondition` and `SequentialCondition` whose schemas can be recursive.

Overall, this PR is simply reshuffling code around and ensuring that the same exports from before are still available.

NOTE that the `schemas` folder is **not** publicly exported as part of `@nucypher/taco` library directly. Instead, schemas are still exported from their original condition module to maintain compatibility.

**Issues fixed/closed:**

> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


vzotova (approved):

manumonti (approved):
LGTM!

Good TS work here, sir 🫡

packages/taco/src/conditions/base/json-api.ts
```diff
@@ -1,39 +1,21 @@
...
-  conditionType: z.literal(JsonApiConditionType).default(JsonApiConditionType),
-  endpoint: z.string().url(),
-  parameters: z.record(z.string(), z.unknown()).optional(),
-  query: jsonPathSchema.optional(),
-  returnValueTest: returnValueTestSchema, // Update to allow multiple return values after expanding supported methods
-});
+import {
+  JsonApiConditionProps,
+  jsonApiConditionSchema,
+  JsonApiConditionType,
```
cygnusv, line 5: What's the criteria for capitalizing some elements and others not?

cygnusv (approved):

packages/taco/src/conditions/base/json-api.ts
derekpierre: I'm unsure if there were criteria, but there was a pre-existing format for earlier conditions where the schema definition was *not* capitalized, but other types were.

PR #595, opened 2024-10-10
chore: update ritual used by Tapir action

manumonti:
**Type of PR:**

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

The open ritual that we used for developers has expired on Oct 6th 24, so a new ritual `#6` has been created for this.

**Notes for reviewers:**

Failing tests should pass when this change is merged.

epic-v0.6.x branch should rebase this change once merged.

### DIFF
--- a/.github/workflows/tapir.yml
+++ b/.github/workflows/tapir.yml
@@ -13,7 +13,7 @@ env:
   RPC_PROVIDER_URL: 'https://rpc-amoy.polygon.technology'
   ENCRYPTOR_PRIVATE_KEY: '0x900edb9e8214b2353f82aa195e915128f419a92cfb8bbc0f4784f10ef4112b86'
   CONSUMER_PRIVATE_KEY: '0xf307e165339cb5deb2b8ec59c31a5c0a957b8e8453ce7fe8a19d9a4c8acf36d4'
-  RITUAL_ID: '0'
+  RITUAL_ID: '6'
 
 jobs:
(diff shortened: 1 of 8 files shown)


derekpierre (changes requested):
Since this is an update to the dev ritual id for `tapir`, it is a little more involved than what we did for the `lynx` update, because `tapir` is used by default for demos/examples.

We are also going to have to update the `DEFAULT_RITUAL_ID` value used for demos: 
- https://github.com/nucypher/taco-web/blob/main/demos/taco-demo/src/config.ts#L3
- https://github.com/nucypher/taco-web/blob/main/demos/taco-nft-demo/src/config.ts#L3

And also the default value used for examples:
- https://github.com/nucypher/taco-web/blob/main/examples/taco/nextjs/src/app/page.tsx#L13
- https://github.com/nucypher/taco-web/blob/main/examples/taco/nodejs/.env.example#L8
- https://github.com/nucypher/taco-web/blob/main/examples/taco/react/src/App.tsx#L12
- https://github.com/nucypher/taco-web/blob/main/examples/taco/webpack-5/src/index.ts#L23

manumonti:
Good catch @derekpierre . Done ✅

derekpierre (approved):
🎸

PR #598, opened 2024-10-18
Improved schema validation for eth_getBalance

cygnusv:
**Type of PR:**

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**

**Issues fixed/closed:**

> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/packages/taco/test/conditions/base/rpc.test.ts
+++ b/packages/taco/test/conditions/base/rpc.test.ts
@@ -89,13 +89,26 @@ describe('validation', () => {
     it('accepts a single address and a block number', () => {
       const result = RpcCondition.validate(rpcConditionSchema, {
         ...testRpcConditionObj,
-        parameters: [TEST_CONTRACT_ADDR, 'latest'],
+        parameters: [TEST_CONTRACT_ADDR, 2345],
       });
 
       expect(result.error).toBeUndefined();
       expect(result.data).toEqual({
         ...testRpcConditionObj,
-        parameters: [TEST_CONTRACT_ADDR, 'latest'],
+        parameters: [TEST_CONTRACT_ADDR, 2345],
+      });
+    });
+
+    it('accepts context params for address and block number', () => {
+      const result = RpcCondition.validate(rpcConditionSchema, {
+        ...testRpcConditionObj,
+        parameters: [":testAddress", ":testBlockNumber"],
+      });
+
+      expect(result.error).toBeUndefined();
+      expect(result.data).toEqual({
+        ...testRpcConditionObj,
+        parameters: [":testAddress", ":testBlockNumber"],
       });
     });
 
--- a/packages/taco/test/conditions/base/rpc.test.ts
+++ b/packages/taco/test/conditions/base/rpc.test.ts
@@ -60,6 +60,32 @@ describe('validation', () => {
       });
     });
 
+    it('accepts a single UserAddress as address', () => {
+      const result = RpcCondition.validate(rpcConditionSchema, {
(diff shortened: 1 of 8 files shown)


packages/taco/test/conditions/base/rpc.test.ts
```diff
@@ -109,7 +143,7 @@ describe('validation', () => {
       expect(result.data).toBeUndefined();
       expect(result.error?.format()).toMatchObject({
         parameters: {
-          _errors: ['Array must contain at least 1 element(s)'],
+          _errors: ['Array must contain at least 2 element(s)', 'Array must contain at least 1 element(s)'],
```
derekpierre, line 146: I did some digging around to see if there is a way to work around this - it's a little conflicting to have both i.e. `at least 2 element(s)` and `at least 1 element(s)`  - but I couldn't find anything.

cygnusv: Yeah, I also found this confusing, but I think this is a side-effect of zod tuples being implemented internally as arrays, where size validation first checks the min and then the max size bound. Not very elegant, but unsure if there's a better way, since we originally wanted to use tuples as each element may have a different type.

cygnusv: In any case, it makes sense because there are two possible schemas here, one with two parameters and the other with one, so the first error corresponds to the validation against the two-parameter schema, and the second to the one-parameter schema.

packages/taco/src/conditions/schemas/context.ts
```diff
@@ -6,7 +6,8 @@ import { plainStringSchema } from './common';
 
 export const contextParamSchema = z.string().regex(CONTEXT_PARAM_REGEXP);
 
-const paramSchema = z.union([plainStringSchema, z.boolean(), z.number()]);
+// TODO: This is too broad, but it's a start
+const paramSchema = z.union([plainStringSchema, z.boolean(), z.number().int().nonnegative()]);
```
derekpierre, line 10: Is there ever a case for negative numbers?

manumonti (approved):
LGTM! 👌

packages/taco/src/conditions/schemas/context.ts
cygnusv: You're right, it is indeed acceptable (`int256` type and friends). Fixed in b3902d4fadeaabec87b754c57e18719a17f3c3ce

derekpierre: I was thinking about this some more. Perhaps we need to separate this schema into two parts.

2. What is generally allowed

Solidity doesn't have floating point numbers, so limiting it to `z.number().int()` is fine. However, a `JsonApiCondition` can require floating point numbers. For example, the `paramOrContextParamSchema` is used for the value used for the returnValueTest () - and for a JSON API, checking a floating point number is perfectly fine, but now it won't be (I saw you had to modify the floating point test)

derekpierre: ^ @cygnusv wdyt?

cygnusv: yep, I was actually reverting my changes as we speak. Let's just put a pin on this so we can unblock the PR.

cygnusv: Opened https://github.com/nucypher/taco-web/issues/600

derekpierre (approved):
🎸

PR #602, opened 2024-11-07
Fix JSON API conditions

theref:
Context params where not being parsed from endpoint, query, or authToken. This PR fixes those issues and adds a test

### DIFF
--- a/packages/taco/src/conditions/context/context.ts
+++ b/packages/taco/src/conditions/context/context.ts
@@ -18,6 +18,7 @@ import {
   CONTEXT_PARAM_PREFIX,
   USER_ADDRESS_PARAMS,
 } from '../const';
+import { JsonApiConditionType } from '../schemas/json-api';
 
 export type CustomContextParam = string | number | boolean;
 export type ContextParam = CustomContextParam | AuthSignature;
@@ -176,6 +177,26 @@ export class ConditionContext {
         for (const param of innerParams) {
           requestedParameters.add(param);
         }
+      }
+    }
+    // If it's a JSON API condition, check url and query
+    if (condition.conditionType === JsonApiConditionType) {
+      const urlComponents = condition.endpoint.replace("https://", "").split("/");
+      for (const param of urlComponents ?? []) {
+        if (this.isContextParameter(param)) {
+          requestedParameters.add(param);
+        }
+      }
+      if (condition.query) {
+        const queryParams = condition.query.match(":[a-zA-Z_]*");
+        if (queryParams) {
+          for (const param of queryParams) {
+            requestedParameters.add(param);
+          }
+        }
+      }
+      if (this.isContextParameter(condition.authorizationToken)) {
+        requestedParameters.add(condition.authorizationToken);
       }
     }
 
--- a/packages/taco/src/conditions/context/context.ts
+++ b/packages/taco/src/conditions/context/context.ts
@@ -188,7 +188,7 @@ export class ConditionContext {
         }
       }
       if (condition.query) {
-        const queryParams = condition.query.match(":[a-zA-Z_]*");
+        const queryParams = condition.query.match(":[a-zA-Z_]+");
         if (queryParams) {
           for (const param of queryParams) {
             requestedParameters.add(param);
--- a/packages/taco/test/taco.test.ts
+++ b/packages/taco/test/taco.test.ts
@@ -136,4 +136,40 @@ describe('taco', () => {
       new Set([customParamKey, USER_ADDRESS_PARAM_DEFAULT]),
     );
   });
-});
+  // test json api condition exposes requested parameters
+  it('jsonapi condition exposes requested parameters', async () => {
+    const mockedDkg = fakeDkgFlow(FerveoVariant.precomputed, 0, 4, 4);
+    const mockedDkgRitual = fakeDkgRitual(mockedDkg);
+    const provider = fakeProvider(aliceSecretKeyBytes);
+    const signer = provider.getSigner();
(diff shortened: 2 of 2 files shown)


packages/taco/src/conditions/context/context.ts
```diff
@@ -178,6 +179,26 @@ export class ConditionContext {
...
+    // If it's a JSON API condition, check url and query
+    if (condition.conditionType === JsonApiConditionType) {
+      const urlComponents = condition.endpoint.replace("https://", "").split("/");
+      for (const param of urlComponents ?? []) {
+        if (this.isContextParameter(param)) {
+          requestedParameters.add(param);
+        }
+      }
+      if (condition.query) {
+        const queryParams = condition.query.match(":[a-zA-Z_]*");
```
KPrasch, line 191: `*` will match zero tokens but we can use `+`

```suggestion
        const queryParams = condition.query.match(":[a-zA-Z_]+");
```

KPrasch (approved):
LGTM!

packages/taco/src/conditions/context/context.ts
```diff
@@ -178,6 +179,26 @@ export class ConditionContext {
...
+      }
+      if (condition.query) {
+        const queryParams = condition.query.match(":[a-zA-Z_]+");
+        if (queryParams) {
+          for (const param of queryParams) {
+            requestedParameters.add(param);
+          }
+        }
+      }
+      if (this.isContextParameter(condition.authorizationToken)) {
```
derekpierre, line 198: An `authorizationToken` is optional and can only be a context parameter. 

```suggestion
      if (condition.authorizationToken) {
```

packages/taco/src/conditions/context/context.ts
```diff
@@ -178,6 +179,26 @@ export class ConditionContext {
...
+    // If it's a JSON API condition, check url and query
+    if (condition.conditionType === JsonApiConditionType) {
+      const urlComponents = condition.endpoint.replace("https://", "").split("/");
+      for (const param of urlComponents ?? []) {
+        if (this.isContextParameter(param)) {
+          requestedParameters.add(param);
+        }
+      }
+      if (condition.query) {
+        const queryParams = condition.query.match(":[a-zA-Z_]+");
```
derekpierre, line 191: There is an existing regex for context variables - `CONTEXT_PARAM_REGEXP` in `const.ts` that you can use. You should use that one instead. Context parameters allow numbers to be in the string (just not the first character).

derekpierre (commented):
@theref - left some additional RFCs still needed even though this PR was already merged.

Please follow-up on those.

Side note: are releases necessary? A different option is to use local dependecy installations and make fixes as you need, and then do one PR into the EPIC. 
eg. 
`$ npm install <path_to_local_taco_web>/taco-web/packages/taco`
`$ npm install <path_to_local_taco_web>/taco-web/packages/taco-auth`

Of course, `pnpm` can be used instead of `npm` depending on what the demo you are testing with uses.

This would prevent needing to do a release after every fix you make.

packages/taco/src/conditions/context/context.ts
KPrasch: @theref discovered some nuance in the existing regex (`CONTEXT_PARAM_REGEXP`) that does not fully cover the Json API condition use case.  In particular the `CONTEXT_PARAM_REGEXP` enforces that the pattern must match at the start of a string (`^`), however, some situations a context parameter may appear as a substring in the middle of an endpoint (for example `https://example.com/:userAddress`).  We think this is specific to the case of JSON API conditions, but let us know if you think otherwise.

KPrasch: I'd also like to point out that there is a small bug in `CONTEXT_PARAM_REGEXP `: `^:[a-zA-Z_][a-zA-Z0-9_]*$`.  The usage of the `*` character here permits matching zero tokens, which means that a simple `:` will match positively.

derekpierre: There are two context regexes in `const.ts`.
- CONTEXT_PARAM_FULL_MATCH_REGEXP - matches a regex without expecting it to be at the beginning of the string i.e. no caret.
- CONTEXT_PARAM_FULL_MATCH_REGEXP - expects the entire string to be a context variable

I believe (?) the first one should be sufficient here.

KPrasch: > It's subtle, but the [a-zA-Z_] part (first block) ensures that the context variable starts with a letter or underscore, the rest of regex allows the remaining characters to be a letter, number, or underscore.

Ah yes, you are correct!

derekpierre:
PS. Nice find @theref - this was definitely missing code in the requested parameters logic.

packages/taco/src/conditions/context/context.ts
KPrasch: > There are two context regexes in `const.ts`.
> 
> * CONTEXT_PARAM_REGEXP - matches a regex without expecting it to be at the beginning of the string i.e. no caret.
> * CONTEXT_PARAM_FULL_MATCH_REGEXP - expects the entire string to be a context variable
> 
> I believe (?) the first one should be sufficient here.

Hmm, perhaps it's in another branch?  I don't see two, only `CONTEXT_PARAM_REGEXP` in `const.ts`

derekpierre: https://github.com/nucypher/taco-web/blob/epic-v0.6.x/packages/taco/src/conditions/const.ts#L10-L15

KPrasch: Yep - that looks suitable.

PR #604, opened 2024-12-13
JWT condition

vzotova:
**Type of PR:**

- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

Implements JWT condition

**Issues fixed/closed:**

See [#3567](https://github.com/nucypher/nucypher/issues/3567)

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/packages/taco/src/conditions/schemas/jwt.ts
+++ b/packages/taco/src/conditions/schemas/jwt.ts
@@ -1,0 +1,20 @@
+import { z } from 'zod';
+
+import { baseConditionSchema } from './common';
+import { contextParamSchema } from './context';
+
+export const JWT_PARAM_DEFAULT = ':jwtToken';
+
+export const JWTConditionType = 'jwt';
+
+export const jwtConditionSchema = baseConditionSchema.extend({
+  conditionType: z.literal(JWTConditionType).default(JWTConditionType),
+  public_key: z.string().optional(),
+  expected_issuer: z.string().optional(),
+  subject: contextParamSchema.optional(),
+  expiration_window: z.number().int().nonnegative().optional(),
+  issued_window: z.number().int().nonnegative().optional(),
+  jwtToken: contextParamSchema.default(JWT_PARAM_DEFAULT),
+});
+
+export type JWTConditionProps = z.infer<typeof jwtConditionSchema>;
--- a/packages/taco/src/conditions/schemas/jwt.ts
+++ b/packages/taco/src/conditions/schemas/jwt.ts
@@ -9,11 +9,11 @@ export const JWTConditionType = 'jwt';
 
 export const jwtConditionSchema = baseConditionSchema.extend({
   conditionType: z.literal(JWTConditionType).default(JWTConditionType),
-  public_key: z.string().optional(),
-  expected_issuer: z.string().optional(),
+  publicKey: z.string().optional(),
+  expectedIssuer: z.string().optional(),
   subject: contextParamSchema.optional(),
-  expiration_window: z.number().int().nonnegative().optional(),
-  issued_window: z.number().int().nonnegative().optional(),
+  expirationWindow: z.number().int().nonnegative().optional(),
+  issuedWindow: z.number().int().nonnegative().optional(),
   jwtToken: contextParamSchema.default(JWT_PARAM_DEFAULT),
 });
 
--- a/packages/taco/src/conditions/schemas/jwt.ts
+++ b/packages/taco/src/conditions/schemas/jwt.ts
@@ -9,7 +9,7 @@ export const JWTConditionType = 'jwt';
 
 export const jwtConditionSchema = baseConditionSchema.extend({
   conditionType: z.literal(JWTConditionType).default(JWTConditionType),
-  publicKey: z.string().optional(),
+  publicKey: z.string(),
   expectedIssuer: z.string().optional(),
   // TODO see https://github.com/nucypher/taco-web/pull/604#discussion_r1901746814
   // subject: contextParamSchema.optional(),
--- a/packages/taco/src/conditions/schemas/jwt.ts
+++ b/packages/taco/src/conditions/schemas/jwt.ts
@@ -11,9 +11,10 @@ export const jwtConditionSchema = baseConditionSchema.extend({
   conditionType: z.literal(JWTConditionType).default(JWTConditionType),
   publicKey: z.string().optional(),
   expectedIssuer: z.string().optional(),
-  subject: contextParamSchema.optional(),
-  expirationWindow: z.number().int().nonnegative().optional(),
-  issuedWindow: z.number().int().nonnegative().optional(),
+  // TODO see https://github.com/nucypher/taco-web/pull/604#discussion_r1901746814
+  // subject: contextParamSchema.optional(),
+  // expirationWindow: z.number().int().nonnegative().optional(),
+  // issuedWindow: z.number().int().nonnegative().optional(),
   jwtToken: contextParamSchema.default(JWT_PARAM_DEFAULT),
 });
 
--- a/packages/test-utils/src/variables.ts
+++ b/packages/test-utils/src/variables.ts
@@ -19,3 +19,8 @@ export const TEST_SIWE_PARAMS = {
   domain: 'localhost',
   uri: 'http://localhost:3000',
 };
+
+export const TEST_ECDSA_PUBLIC_KEY =
+  '-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEXHVxB7s5SR7I9cWwry' +
+  '/JkECIReka\nCwG3uOLCYbw5gVzn4dRmwMyYUJFcQWuFSfECRK+uQOOXD0YSEucBq0p5tA==\n-----END PUBLIC ' +
+  'KEY-----\n ';
--- a/packages/taco/package.json
+++ b/packages/taco/package.json
@@ -52,7 +52,7 @@
     "@nucypher/test-utils": "workspace:*",
     "@types/semver": "^7.5.8",
     "glob": "^11.0.1",
-    "modified-zod2md": "0.1.5-modified.3"
+    "modified-zod2md": "0.1.5-modified.4"
   },
   "engines": {
     "node": ">=18",
--- a/packages/taco/schema-docs/condition-schemas.md
+++ b/packages/taco/schema-docs/condition-schemas.md
@@ -10,6 +10,7 @@ _Union of the following possible types:_
 - [CompoundCondition](#compoundcondition)
 - [JsonApiCondition](#jsonapicondition)
 - [JsonRpcCondition](#jsonrpccondition)
+- [JwtCondition](#jwtcondition)
 - [SequentialCondition](#sequentialcondition)
 - [IfThenElseCondition](#ifthenelsecondition)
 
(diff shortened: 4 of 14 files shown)


packages/taco/src/conditions/schemas/jwt.ts
```diff
@@ -0,0 +1,20 @@
...
+import { baseConditionSchema } from './common';
+import { contextParamSchema } from './context';
+
+export const JWT_PARAM_DEFAULT = ':jwtToken';
+
+export const JWTConditionType = 'jwt';
+
+export const jwtConditionSchema = baseConditionSchema.extend({
+  conditionType: z.literal(JWTConditionType).default(JWTConditionType),
+  public_key: z.string().optional(),
```
theref, line 12: are these parameters really optional?

packages/taco/src/conditions/schemas/jwt.ts
```diff
@@ -0,0 +1,20 @@
...
+  expected_issuer: z.string().optional(),
+  subject: contextParamSchema.optional(),
+  expiration_window: z.number().int().nonnegative().optional(),
+  issued_window: z.number().int().nonnegative().optional(),
+  jwtToken: contextParamSchema.default(JWT_PARAM_DEFAULT),
```
theref, line 17: isn't the jwtToken what the user will provide at decryption time? rather than being defined in the condition

derekpierre:
Converted to draft since this work shouldn't be merged before https://github.com/nucypher/nucypher/pull/3570 (currently in draft) is merged

packages/taco/src/conditions/schemas/jwt.ts
vzotova: yep, you are correct, here is defined that some user param must be provided or used default :jwtToken

packages/taco/src/conditions/schemas/jwt.ts
vzotova: I need @cygnusv help with that

packages/test-utils/src/variables.ts
```diff
@@ -19,3 +19,8 @@ export const TEST_SIWE_PARAMS = {
   domain: 'localhost',
   uri: 'http://localhost:3000',
 };
+
+export const TEST_ECDSA_PUBLIC_KEY =
```
derekpierre, line 23: Is it possible to provided the public key without the PEM header and footer i.e. just use the base64? On the server-side we can do some adjustment if needed: add it back, OR not bother if the library allows non-PEM format, or ...

(cc @cygnusv )

packages/taco/src/conditions/schemas/jwt.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from 'zod';
+
+import { baseConditionSchema } from './common';
+import { contextParamSchema } from './context';
+
+export const JWT_PARAM_DEFAULT = ':jwtToken';
+
+export const JWTConditionType = 'jwt';
+
+export const jwtConditionSchema = baseConditionSchema.extend({
```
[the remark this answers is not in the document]
derekpierre: Sorry, I meant the individual properties within the schema i.e.
- `public_key` -> `publicKey`
- `expiration_window` -> `expirationWindow`
...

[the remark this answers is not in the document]
vzotova: got it,
✔️

packages/taco/src/conditions/schemas/jwt.ts
```diff
@@ -0,0 +1,20 @@
...
+  conditionType: z.literal(JWTConditionType).default(JWTConditionType),
+  publicKey: z.string().optional(),
+  expectedIssuer: z.string().optional(),
+  subject: contextParamSchema.optional(),
+  expirationWindow: z.number().int().nonnegative().optional(),
+  issuedWindow: z.number().int().nonnegative().optional(),
```
cygnusv, line 16: @vzotova let's comment/drop these lines for the moment. The corresponding PR on the nucypher side will only validate the expected issuer and public key. We can easily add more restrictions like these once we have more clarity on what's actually needed.

vzotova: ✔️

packages/taco/src/conditions/schemas/jwt.ts
```diff
@@ -0,0 +1,21 @@
...
+import { baseConditionSchema } from './common';
+import { contextParamSchema } from './context';
+
+export const JWT_PARAM_DEFAULT = ':jwtToken';
+
+export const JWTConditionType = 'jwt';
+
+export const jwtConditionSchema = baseConditionSchema.extend({
+  conditionType: z.literal(JWTConditionType).default(JWTConditionType),
+  publicKey: z.string().optional(),
```
derekpierre, line 12: Based on https://github.com/nucypher/nucypher/pull/3570, `publicKey` is a required property.

derekpierre (commented):
Needs a rebase now that #606 is merged.

packages/taco/src/conditions/schemas/jwt.ts
vzotova: ✔️

derekpierre:
Now that we are generating a `condition-schemas.md` as part of our schema documentation (thanks to @Muhammad-Altabba ), can you do the following:
1. add an export here - https://github.com/vzotova/taco-web/blob/jwt/packages/taco/src/conditions/schemas/export-for-zod-doc-gen.ts#L16 - for `jwt`
2. Go to `packages/taco` and run
    - `$ pnpm install`
    - `$ pnpm generate-zod-docs`
    
This should update the `packages/taco/schema-docs/condition-schemas.md` file to include a schema table about the shiny new `JWTCondition`.

derekpierre:
Side note: you also need signed commits to pass the merge check.

vzotova:
> Now that we are generating a `condition-schemas.md` as part of our schema documentation (thanks to @Muhammad-Altabba ), can you do the following:
> 
> 1. add an export here - https://github.com/vzotova/taco-web/blob/jwt/packages/taco/src/conditions/schemas/export-for-zod-doc-gen.ts#L16 - for `jwt`
> 2. Go to `packages/taco` and run
>    
>    * `$ pnpm install`
>    * `$ pnpm generate-zod-docs`
> 
> This should update the `packages/taco/schema-docs/condition-schemas.md` file to update the file and include a schema table about the shiny new `JWTCondition`.

Looks like it's not working on windows very well. I'll spend some time to figure out why, meanwhile could you generate this file for me please?

theref (approved):

packages/taco/src/conditions/schemas/jwt.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const jwtConditionSchema = baseConditionSchema.extend({
+  conditionType: z.literal(JWTConditionType).default(JWTConditionType),
+  publicKey: z.string(),
+  expectedIssuer: z.string().optional(),
+  // TODO see https://github.com/nucypher/taco-web/pull/604#discussion_r1901746814
+  // subject: contextParamSchema.optional(),
+  // expirationWindow: z.number().int().nonnegative().optional(),
+  // issuedWindow: z.number().int().nonnegative().optional(),
+  jwtToken: contextParamSchema.default(JWT_PARAM_DEFAULT),
```
derekpierre, line 18: By using "default" there is an implicitness here. See https://github.com/nucypher/nucypher/issues/3567#issuecomment-2766130590.

By using a default, an implicit context variable is used at encryption that the application developer would not explicitly know about, that they need to provide in the condition context at decryption time. Of course, they would eventually realize during testing because at decryption time the condition context would complain that there is no value provided for the `:jwtToken` context variable, and they would update their code - so they would realize anyway.

To clarify, I'm not against the implicitness, but we just have to recognize it.

Wdyt? (cc @cygnusv , @Muhammad-Altabba )

cygnusv: I think the issue mostly depends on how the API and docs look like, and not so much on how the schema is defined. Some of the concerns here are based on the assumption of developers somehow creating their own condition json, but in practice most developers will use a JWTCondition object from the API. Ideally, the decryption API should detect as soon as possible if a JWTCondition doesn't have a jwt token, but that's a different issue.

Muhammad-Altabba: Actually we are using the zod schema to generate the condition doc: https://github.com/nucypher/taco-web/pull/604/files#diff-71a9d1d0356cf1f1471c86c3beb1f9a291e4abba2059e3679c2a2e9bffa9d4c8R170.

derekpierre: > Actually we are using the zod schema to generate the condition doc: https://github.com/nucypher/taco-web/pull/604/files#diff-71a9d1d0356cf1f1471c86c3beb1f9a291e4abba2059e3679c2a2e9bffa9d4c8R170.

That's the schema doc, but I think David is referring to eventual `JWTCondition` docs in https://docs.taco.build. 

However, how come the default value isn't in the default column for `jwtToken`?

derekpierre: > I think the issue mostly depends on how the API and docs look like, and not so much on how the schema is defined. Some of the concerns here are based on the assumption of developers somehow creating their own condition json, but in practice most developers will use a JWTCondition object from the API

The API and the schema are directly related - the schema dictates what is required and allowed by the API and how parameters are validated.

Since `jwtToken` is defined in the schema as:
```typescript
jwtToken: contextParamSchema.default(JWT_PARAM_DEFAULT),
```

It means two things. The encryption/decryption code can either look like:
(A) specific context variable name (`:jwtToken`)
```typescript
// encryption
// This is allowed because `.default(JWT_PARAM_DEFAULT)` is used
condition = JWTCondition(public_key="...");

...
// decryption
conditionContext.addCustomContextParameterValues({":jwtToken", jwtToken})

decrypt(..., conditionContext, ...);
```


OR
(B) allow any context variable name
```typescript
// encryption
// This is allowed because `contextParamSchema` is used in the 
condition = JWTCondition(public_key="...", jwtToken=":jwtTokenContextVar");   // or any other variable name
...
// decryption
conditionContext.addCustomContextParameterValues({":jwtTokenContextVar", jwtToken})
decrypt(..., conditionContext, ...);
```

In the grand scheme of what we can allow for `jwtToken` property on `JWTCondition`, you can only allow two scenarios:
1. allow (A) or (B)
2. allow (B) alone

Either 1) or 2) is fine, but we just need to document it properly. The current code in the PR allows for either (A) or (B) with (A) being the default and therefore implicitly using `:jwtToken`. So our corresponding docs should make it clear that the `JWTCondition` uses `:jwtToken` as the context variable name unless a different one is specifiedand therefore its corresponding value needs to be provided at decryption time otherwise an error will be raised. The developer also has the option to use different context variables if, for instance, they are using a Compound.AND that validates two separate jwts i.e. they would have to provide two jwts for validation necessitating two different context variables - this is why the (B) part (allowing any context variable name) is also important and why we can't allow just (A) without allowing (B) - which is what was commented here - https://github.com/nucypher/nucypher/issues/3567#issuecomment-2766036785.

> Ideally, the decryption API should detect as soon as possible if a JWTCondition doesn't have a jwt token, but that's a different issue.

Yep. The subsequent call to decrypt will detect the missing context variable value and raise an Error.

Muhammad-Altabba: > However, how come the default value isn't in the default column for `jwtToken`?

I noticed this bug today and will fix it shortly, regardless of the outcome of the discussion.

derekpierre: > > However, how come the default value isn't in the default column for `jwtToken`?
> 
> I noticed this bug today and will fix it shortly, regardless of the outcome of the discussion.

👍 Awesome thanks! I assume that fix is in the https://github.com/nucypher/zod2md repos? If so, depending on timing we can either fix it as part of this PR, or a follow-up one.

Muhammad-Altabba: @derekpierre thanks for the great explanation especially:
> The developer also has the option to use different context variables if, for instance, they are using a Compound.AND that validates two separate jwts i.e. they would have to provide two jwts for validation necessitating two different context variables ...
> 
Things make sense to me now.

Muhammad-Altabba: > > > However, how come the default value isn't in the default column for `jwtToken`?
> > 
> > 
> > I noticed this bug today and will fix it shortly, regardless of the outcome of the discussion.
> 
> 👍 Awesome thanks! I assume that fix is in the https://github.com/nucypher/zod2md repos? If so, depending on timing we can either fix it as part of this PR, or a follow-up one.

Just been fixed now and applied the fix to this PR.
PR for nucypher/zod2md: https://github.com/nucypher/zod2md/pull/3

Muhammad-Altabba (approved):
Looks great to me.

derekpierre (approved):
🎸

Issue #620, opened 2025-02-11
Deprecate `:userAddressExternalEIP4361`


derekpierre:
# Background

When authentication via context variables was first introduced in taco-web, it was designed as an implicit and automated process. The workflow was as follows:

- A user would initiate a decryption request to the library.
- taco-web would parse the condition to check for reserved context variables (e.g., `:userAddress`).
- If any reserved context variables were found, the user would be prompted for an authentication signature—typically triggering a MetaMask window for signing and confirmation.

This worked fine and led to expansion of the `:userAddress` to `:userAddressEIP712` and `:userAddressEIP4361`. Although this was the beginning of a mistaken path. Because the prompting was implicit, these separate context variables indicated the type of signature to use, as opposed to the application controlling it themselves; the condition dictated it.

Later, the need arose to implement EIP-4361 single sign-on (SSO). Initially, we attempted to introduce a separate context variable, `:userAddressExternalEIP4361`, to bypass the authentication prompt. However, this approach posed a fundamental problem - the implicit authentication logic in taco-web had no way of determining where the SSO data resided within the application.

To address this, we introduced AuthProviders ([PR #531](https://github.com/nucypher/taco-web/pull/531)), encapsulating authentication and signing logic within dedicated objects. This shift made the process explicit, allowing applications to define their own authentication providers to generate the required signatures. Applications could choose to support SSO if desired, or fall back to prompting the user for authentication—without taco-web having to make that decision or rely on additional flags. This was a significant improvement in flexibility and maintainability.

# The Mistake: Introducing `:userAddressExternalEIP4361`
When implementing SSO, we made an oversight in [PR #529](https://github.com/nucypher/taco-web/pull/529). We introduced `:userAddressExternalEIP4361` to indicate when only SSO should be used. However, this had unintended consequences: any condition using this variable was restricted to SSO exclusively, precluding regular EIP-4361 authentication.

What we actually needed was for `:userAddress` to support both standard EIP-4361 and SSO-based EIP-4361 authentication seamlessly. Instead, introducing `:userAddressExternalEIP4361` needlessly added an extra context variable, when `:userAddress` alone could have handled both cases. This approach unnecessarily restricted any condition using `:userAddressExternalEIP4361` to only support SSO, making it overly rigid and limiting.

# Why Deprecation is Necessary
[...]

PR #623, opened 2025-02-27
Deprecate `:userAddressExternalEIP4361` and its related code and logic

Muhammad-Altabba:
**Type of PR:**

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**

Deprecate `:userAddressExternalEIP4361` and its related code and logic at `taco-web` according to  #620


packages/taco-auth/src/providers/eip4361/external-eip4361.ts
```diff
@@ -1,42 +0,0 @@
-import { SiweMessage } from 'siwe';
-
-import { AuthSignature } from '../../auth-sig';
-
-import { EIP4361_AUTH_METHOD } from './common';
-
-export const USER_ADDRESS_PARAM_EXTERNAL_EIP4361 =
```
derekpierre, line 7: We still need this file - just not this constant.

You still want developers to provide the `SingleSignOnEIP4361AuthProvider` but simply use it  with `:userAddress`.

derekpierre (changes requested):

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
derekpierre: ^ Let's just make these changes directly in this PR; mo need for another branch. You can always clean up commits afterwards, within this branch, if there are subsequent issues.

Muhammad-Altabba: Ok 👍 
Pushed the changes to be in this PR. And converted it to a Draft.

Muhammad-Altabba: @derekpierre 
Pushed some changes again according to our discussion.  Is it good now so I finalize it and mark it as ready?
(I will later add some method docs and check the CI failed jobs...)
Thanks,

packages/taco-auth/src/providers/eip4361/common.ts
```diff
@@ -1,8 +1,16 @@
 import { SiweMessage } from 'siwe';
 import { z } from 'zod';
 
+import { AuthSignature } from '../../auth-sig';
+
 export const EIP4361_AUTH_METHOD = 'EIP4361';
 
+export const USER_ADDRESS_PARAM_DEFAULT = ':userAddress';
+
+export interface IEIP4361AuthProvider {
```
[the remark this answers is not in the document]
Muhammad-Altabba: Sorry, I oversight that AuthProvider interface is the one having this method.

I added `implements AuthProvider` to the three authentication classes for clarity.

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
```diff
@@ -2,19 +2,30 @@ import { SiweMessage } from 'siwe';
...
     existingSiweMessage: string,
     signature: string,
   ): Promise<SingleSignOnEIP4361AuthProvider> {
-    // validation
-    const siweMessage = new SiweMessage(existingSiweMessage);
-    await siweMessage.verify({ signature });
+    const siweMessage = await createVerifiedSiweMessage(
+      existingSiweMessage,
+      signature,
+    );
```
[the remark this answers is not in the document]
Muhammad-Altabba: Yes, it is 2 lines of code and 1 piece of comment. I repeated them now and deleted the method.

packages/taco-auth/src/providers/eip4361/eip4361.ts
```diff
@@ -67,37 +66,33 @@ export class EIP4361AuthProvider {
...
-      nonce,
+      statement: `${this.params.domain} wants you to sign in with your Ethereum account: ${address}`,
       chainId,
+      // if the user does not provide an expirationTime, we set it to 2 hours from now
+      // if the user provides an expirationTime that is more than 2 hours from now, we set it to 2 hours from now
+      expirationTime: !this.params?.expirationTime
+        ? after2HoursFromNow
+        : after2HoursFromNow < this.params.expirationTime
+          ? after2HoursFromNow
+          : this.params.expirationTime,
```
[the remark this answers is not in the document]
Muhammad-Altabba: Logic is now updated to throw if it is more than 2 hours. And it still defaults to 2 hours.
And, if the developer provided a value for `expirationTime`, the time span from now (i.e. expiration duration) will be inferred and used every time a new message is created. So, if the developer provided `expirationTime` that is 1 hour from now, every newly created message will have 1 hour before it get expired.

packages/taco-auth/src/providers/eip4361/eip4361.ts
```diff
@@ -9,34 +10,73 @@ import {
...
-      this.providerParams = providerParams;
-    } else {
-      this.providerParams = this.getDefaultParameters();
+
+    const after2HoursFromNow = new Date(
+      Date.now() + 2 * 60 * 60 * 1000,
+    ).toISOString();
+    // if the provided expirationTime was more than 2 hours from now, throw
+    if (params?.expirationTime && params.expirationTime > after2HoursFromNow) {
+      throw new Error('expirationTime is too soon');
```
manumonti, line 56: Probably I'm misunderstanding this, doesn't the error be "expirationTime is too late"? 

We are providing to the constructor the expiration time for SIWE cache, and we are limiting how long this can be (2h). So, if the user tries to set a expiration time longer than 2 hours (`params.expirationTime > after2HoursFromNow`), the error say that the expiration time is too late. Right?

packages/taco-auth/src/providers/eip4361/eip4361.ts
```diff
@@ -53,19 +93,34 @@ export class EIP4361AuthProvider {
...
 
     // If we have a signature in localStorage, return it
     const maybeSignature = this.storage.getAuthSignature(storageKey);
     if (maybeSignature) {
-      // check whether older than node freshness requirement
-      if (this.isMessageExpired(maybeSignature.typedData)) {
+      const siweMessage = new SiweMessage(maybeSignature.typedData);
+      try {
+        // check message validity including `expirationTime` and `notBefore`.
+        await siweMessage.verify({ signature: maybeSignature.signature });
```
manumonti, line 119: I like this! 👌

packages/taco-auth/src/providers/eip4361/eip4361.ts
```diff
@@ -75,37 +130,39 @@ export class EIP4361AuthProvider {
...
+        Date.now() + this.expirationDuration,
+      ).toISOString(),
     });
     const scheme = EIP4361_AUTH_METHOD;
+
     const message = siweMessage.prepareMessage();
     const signature = await this.signer.signMessage(message);
+
+    // this will trigger validation for all parameters including `expirationTime` and `notBefore`.
+    await siweMessage.verify({ signature });
```
manumonti, line 164: should we add a .catch() here? I mean, what happens if this verification fail? Should we return an error?

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
```diff
@@ -22,13 +37,33 @@ export class SingleSignOnEIP4361AuthProvider {
...
+   * before returning them as an EIP4361AuthSignature object.
+   *
+   * @returns {Promise<EIP4361AuthSignature>} The validated authentication signature
+   * @throws {Error} If signature verification fails
+   */
   public async getOrCreateAuthSignature(): Promise<EIP4361AuthSignature> {
+    const siweMessage = new SiweMessage(this.existingSiweMessage);
+    // this will trigger validation for the signature and all parameters (`expirationTime`, `notBefore`...).
+    await siweMessage.verify({ signature: this.signature });
+
```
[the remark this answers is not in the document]
Muhammad-Altabba: Unfortunately, `siweMessage.verify` is an async method which make it does not fit into the constructor.

Muhammad-Altabba:
Thanks @derekpierre and @manumonti for the detailed review.
I think I was proposing some enhancements but mixed with some opinionated decisions which should not be part of this PR in the first place, sorry for that. And for this, I reverted most of the refactoring that would need further discussions that is better to be done along and according to some users feedback or needs (as Derek suggested me early). And I kept in this PR just basic enhancements in addition to the JSDoc comments and the test modifications.
Let me know please if there is more to revert or further modify more.

packages/taco-auth/src/providers/eip4361/eip4361.ts
Muhammad-Altabba: Sorry for the confusion, the message was for something and I missed updating it.

packages/taco/test/conditions/context.test.ts
```diff
@@ -491,14 +474,14 @@ describe('context', () => {
             customContractCondition,
           );
           conditionContext.addAuthProvider(
-            USER_ADDRESS_PARAM_EXTERNAL_EIP4361,
-            authProviders['SSO4361'],
+            USER_ADDRESS_PARAM_DEFAULT,
+            authProviders[EIP4361],
```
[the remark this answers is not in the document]
Muhammad-Altabba: It should be regardless of which auth provider. But there was no need to edit it. So, now, I kept it as it was.

packages/taco/test/conditions/context.test.ts
```diff
@@ -469,7 +452,7 @@ describe('context', () => {
         const conditionContext = new ConditionContext(customContractCondition);
         conditionContext.addAuthProvider(
           USER_ADDRESS_PARAM_DEFAULT,
-          authProviders['EIP1271'],
+          authProviders[EIP4361],
```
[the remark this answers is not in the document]
Muhammad-Altabba: It should be regardless of which auth provider. But there was no need to edit it. So, now, I kept it as it was.

packages/taco-auth/src/providers/eip4361/eip4361.ts
Muhammad-Altabba: Thanks,

packages/taco-auth/src/providers/eip4361/eip4361.ts
```diff
@@ -9,34 +10,73 @@ import {
...
+   * @param params.statement - Human-readable message shown to users during signing. Default is `{{domain}} wants you to sign in with your Ethereum account: {{address}}`.
+   * @throws {Error} If expirationTime is set beyond 2 hours from the current time.
+   */
   constructor(
     // TODO: We only need the provider to fetch the chainId, consider removing it
     private readonly provider: ethers.providers.Provider,
     private readonly signer: ethers.Signer,
-    providerParams?: EIP4361AuthProviderParams,
+    // `chainId` will be set from the provider. For do not accept it as input to eleminate ambiguity.
+    params?: Partial<Omit<Omit<SiweMessage, 'address'>, 'chainId'>>,
```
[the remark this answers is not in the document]
Muhammad-Altabba: Thanks Derek,
I reverted the related proposed changes. Probably in the future we can have more discussion about a full refactoring or an additional class. And probably even make the authentication process as plugin-based in order to be supper flexible. And that need to modify the nodes capabilities, to also support plugin-based auth, in order to work in harmony. 
So, for now, I will leave this as it was till we have feedbacks and requests from the users, as you suggested to me earlier.

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
```diff
@@ -1,18 +1,33 @@
...
+   * @throws {Error} If signature verification fails or message parameters are invalid
+   */
   public static async fromExistingSiweInfo(
     existingSiweMessage: string,
     signature: string,
   ): Promise<SingleSignOnEIP4361AuthProvider> {
-    // validation
     const siweMessage = new SiweMessage(existingSiweMessage);
+    // this will trigger validation for the signature and all parameters (`expirationTime`, `notBefore`...).
     await siweMessage.verify({ signature });
```
derekpierre, line 29: Should we be checking here for 2-hour expiry? We didn't before and I believe you had a suggestion to do so, which I think is a good one. I believe you can include the two-hour time check by passing a 2-hour time as one of the `VerifyParams` - https://github.com/spruceid/siwe/blob/main/packages/siwe/lib/client.ts#L270.

Also needs to be done in `getOrCreateAuthSignature()`.

packages/taco/src/conditions/schemas/common.ts
```diff
@@ -23,10 +20,11 @@ export const plainStringSchema = z
...
   );
 
-export const UserAddressSchema = z.enum([
-  USER_ADDRESS_PARAM_DEFAULT,
-  USER_ADDRESS_PARAM_EXTERNAL_EIP4361,
-]);
+export const UserAddressSchema = z
+  .literal(USER_ADDRESS_PARAM_DEFAULT)
+  .describe(
+    'This is a context parameter that will be replaced at decryption time. It represents the Ethereum address of the user attempting decryption.',
```
derekpierre, line 26: context variable - https://docs.taco.build/authentication/conditioncontext-and-context-variables#context-variables.

```suggestion
    'This is a context variable that will be replaced at decryption time. It represents the Ethereum address of the requester attempting decryption.',
```

packages/taco/test/conditions/context.test.ts
```diff
@@ -52,8 +57,8 @@ describe('context', () => {
 
   describe('reserved context parameters', () => {
     it.each([
-      [USER_ADDRESS_PARAM_DEFAULT, 'EIP4361'],
-      [USER_ADDRESS_PARAM_EXTERNAL_EIP4361, 'SSO4361'],
+      [USER_ADDRESS_PARAM_DEFAULT, EIP4361],
+      [USER_ADDRESS_PARAM_DEFAULT, SSO_EIP4361],
```
derekpierre, line 61: Can we add EIP1271 here too to be complete.

```suggestion
      [USER_ADDRESS_PARAM_DEFAULT, SSO_EIP4361],
      [USER_ADDRESS_PARAM_DEFAULT, EIP1271],
```

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
derekpierre: If you add it, also add a test to ensures it works properly.

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
```diff
@@ -1,18 +1,62 @@
...
+  message: string,
+  signature: string,
+): Promise<SiweMessage> {
+  const siweMessage = new SiweMessage(message);
+  // this will trigger validation for the signature and all parameters (`expirationTime`, `notBefore`...).
+  await siweMessage.verify({ signature });
+  const twoHoursBeforeNow = new Date(Date.now() - 2 * 60 * 60 * 1000);
+  if (!siweMessage.issuedAt) {
+    throw new Error(
+      `The message does not have an \`issuedAt\` field. This case should never happen and it would cause the message to be rejected by TACo nodes.`,
```
derekpierre, line 17: the "this should never happen" is really an internal dev note.

```suggestion
      // this should never happen
      `The SIWE message is missing an \`issuedAt\` field and would be rejected by TACo nodes.`,
```

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
```diff
@@ -1,18 +1,62 @@
...
+    );
+  }
+  if (new Date(siweMessage.issuedAt) < twoHoursBeforeNow) {
+    throw new Error(
+      `The message was issued more than 2 hours ago or does not have an \`issuedAt\` field. This message would be rejected by TACo nodes.`,
```
derekpierre, line 22: At this point you've already checked that `issuedAt` is present.

```suggestion
    throw new Error(
      `The SIWE message was issued more than 2 hours ago and would be rejected by TACo nodes.`,
```

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
```diff
@@ -1,18 +1,62 @@
...
+    );
+  }
+  const now = new Date();
+  if (new Date(siweMessage.issuedAt) > now) {
+    throw new Error(
+      `The message was issued at a future datetime: ${siweMessage.issuedAt}. This message would be rejected by TACo nodes.`,
+    );
```
derekpierre, line 29:
```suggestion
      `The SIWE message was issued at a future datetime: ${siweMessage.issuedAt} and would be rejected by TACo nodes.`,
    );
```

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
```diff
@@ -1,18 +1,62 @@
...
-export const USER_ADDRESS_PARAM_EXTERNAL_EIP4361 =
-  ':userAddressExternalEIP4361';
+async function generateAndVerifySiweMessage(
+  message: string,
+  signature: string,
+): Promise<SiweMessage> {
+  const siweMessage = new SiweMessage(message);
+  // this will trigger validation for the signature and all parameters (`expirationTime`, `notBefore`...).
+  await siweMessage.verify({ signature });
+  const twoHoursBeforeNow = new Date(Date.now() - 2 * 60 * 60 * 1000);
```
derekpierre, line 14: Is it possible for us to make the `2 * 60 * 60 * 1000` a constant that could be used by both EIP4361AuthProvider and this one? It would just make maintenance of the value simpler by having it defined in one place, but used in multiple.

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
derekpierre: Note these error message changes, if used, will affect your tests.

packages/taco-auth/src/providers/eip4361/external-eip4361.ts
```diff
@@ -1,18 +1,62 @@
 import { SiweMessage } from 'siwe';
 
+import { AuthProvider } from '../../auth-provider';
+
 import { EIP4361_AUTH_METHOD, EIP4361AuthSignature } from './auth';
 
-export const USER_ADDRESS_PARAM_EXTERNAL_EIP4361 =
-  ':userAddressExternalEIP4361';
+async function generateAndVerifySiweMessage(
```
derekpierre, line 7: 👍

packages/taco-auth/test/auth-provider.test.ts
```diff
@@ -155,6 +155,48 @@ describe('eip4361 single sign-on auth provider', async () => {
...
+    await expect(
+      SingleSignOnEIP4361AuthProvider.fromExistingSiweInfo(
+        siweMessage.prepareMessage(),
+        signature,
+      ),
+    ).rejects.toThrow(
+      `The message was issued at a future datetime: ${siweMessage.issuedAt}. This message would be rejected by TACo nodes.`,
+    );
+  });
 });
```
derekpierre, line 200: Nice tests! 

Can we add another whether `fromExistingSiweInfo()` passes, but then the call the `getOrCreateAuthSignature()` fails?

You can use `vi.setSystemTime(...);` to adjust the time between using `fromExistingSiweInfo` and `getOrCreateAuthSignature()`. There is a test that uses it in this module.

packages/taco-auth/test/auth-provider.test.ts
```diff
@@ -155,6 +156,80 @@ describe('eip4361 single sign-on auth provider', async () => {
...
+  });
+
+  it('throws an error if the SIWE message is expired when calling getOrCreateAuthSignature()', async () => {
+    const siweMessage = new SiweMessage({
+      domain: 'taco.build',
+      address: await signer.getAddress(),
+      uri: 'https://taco.build',
+      version: '1',
+      chainId: (await provider.getNetwork()).chainId,
+      issuedAt: new Date(Date.now()).toISOString(), // 3 hours from now
```
derekpierre, line 209:
```suggestion
      issuedAt: new Date(Date.now()).toISOString(),
```

derekpierre (approved):
Great work 🎸!

One minor suggestion to remove a comment, then we can merge.

PR #633, opened 2025-03-11
Use more restrictive schemas for evm base types

### LINKED ISSUE #600: Consider more restrictive schemas for EVM base types
I was thinking about this some more. Perhaps we need to separate this schema into two parts.

2. What is generally allowed

Solidity doesn't have floating point numbers, so limiting it to `z.number().int()` is fine. However, a `JsonApiCondition` can require floating point numbers. For example, the `paramOrContextParamSchema` is used for the value used for the returnValueTest () - and for a JSON API, checking a floating point number is perfectly fine, but now it won't be; typified by the fact that I saw you had to modify the floating point test.

_Originally posted by @derekpierre in https://github.com/nucypher/taco-web/pull/598#discussion_r1816589503_

Muhammad-Altabba:
**Type of PR:**

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**

**Issues fixed/closed:**

- Closes #600

### DIFF
--- a/packages/taco/src/conditions/schemas/context.ts
+++ b/packages/taco/src/conditions/schemas/context.ts
@@ -14,8 +14,10 @@ export const contextParamSchema = z
 const paramSchema = z.union([plainStringSchema, z.boolean(), z.number()]);
 
 const rpcParamSchema = z
-  .union([plainStringSchema, z.boolean(), z.number().int().nonnegative()])
-  .describe('Smart Contract function parameters');
+  .union([plainStringSchema, z.boolean(), z.number().int()])
+  .describe(
+    'Used for EVM RPC parameters and Smart Contract function parameters',
+  );
 
 export const paramOrContextParamSchema: z.ZodSchema = z.union([
   paramSchema,
--- a/packages/taco/src/conditions/schemas/context.ts
+++ b/packages/taco/src/conditions/schemas/context.ts
@@ -13,8 +13,18 @@ export const contextParamSchema = z
 
 const paramSchema = z.union([plainStringSchema, z.boolean(), z.number()]);
 
+const rpcParamSchema = z
+  .union([plainStringSchema, z.boolean(), z.number().int().nonnegative()])
+  .describe('Smart Contract function parameters');
+
 export const paramOrContextParamSchema: z.ZodSchema = z.union([
   paramSchema,
   contextParamSchema,
   z.lazy(() => z.array(paramOrContextParamSchema)),
+]);
+
+export const rpcParamOrContextParamSchema: z.ZodSchema = z.union([
+  rpcParamSchema,
+  contextParamSchema,
+  z.lazy(() => z.array(rpcParamOrContextParamSchema)),
 ]);
--- a/packages/taco/src/conditions/schemas/context.ts
+++ b/packages/taco/src/conditions/schemas/context.ts
@@ -13,10 +13,10 @@ export const contextParamSchema = z
 
 const paramSchema = z.union([plainStringSchema, z.boolean(), z.number()]);
 
-const rpcParamSchema = z
+const nonFloatParamSchema = z
   .union([plainStringSchema, z.boolean(), z.number().int()])
   .describe(
-    'Used for EVM RPC parameters and Smart Contract function parameters',
+    'Non-floating point (string, boolean, or integer). Used for parameters passed to blockchain RPC endpoints and Smart Contracts functions',
   );
 
 export const paramOrContextParamSchema: z.ZodSchema = z.union([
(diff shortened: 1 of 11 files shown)


packages/taco/src/conditions/schemas/context.ts
```diff
@@ -13,8 +13,18 @@ export const contextParamSchema = z
 
 const paramSchema = z.union([plainStringSchema, z.boolean(), z.number()]);
 
+const rpcParamSchema = z
```
derekpierre, line 16: Let's name this something else, perhaps `onChainParamSchema`, `blockchainParamSchema`, ...?

packages/taco/src/conditions/schemas/context.ts
```diff
@@ -13,8 +13,18 @@ export const contextParamSchema = z
...
+  .union([plainStringSchema, z.boolean(), z.number().int().nonnegative()])
+  .describe('Smart Contract function parameters');
```
derekpierre, line 18: Not just smart contract but blockchain RPC parameter types as well.

packages/taco/src/conditions/schemas/context.ts
```diff
@@ -13,8 +13,18 @@ export const contextParamSchema = z
...
+
 export const paramOrContextParamSchema: z.ZodSchema = z.union([
   paramSchema,
   contextParamSchema,
   z.lazy(() => z.array(paramOrContextParamSchema)),
 ]);
+
+export const rpcParamOrContextParamSchema: z.ZodSchema = z.union([
```
derekpierre, line 26: The rename will affect this name as well.

packages/taco/test/conditions/context.test.ts
```diff
@@ -507,115 +508,346 @@ describe('context', () => {
...
-  it('accepts a plain string', () => {
-    expect(paramOrContextParamSchema.safeParse('hello').success).toBe(true);
-  });
+describe.each(['paramOrContextParam', 'rpcParamOrContextParam'])(
+  '%s schema',
+  (schemaName) => {
+    let schema = paramOrContextParamSchema;
+    if (schemaName === 'rpcParamOrContextParam') {
+      schema = rpcParamOrContextParamSchema;
+    }
```
derekpierre, line 517: Could you have just used the schema directly? i.e.

```suggestion
describe.each([paramOrContextParamSchema, rpcParamOrContextParamSchema])(
  '%s schema',
  (schema) => {
```

packages/taco/src/conditions/schemas/return-value-test.ts
```diff
@@ -1,11 +1,22 @@
...
   comparator: z.enum(['==', '>', '<', '>=', '<=', '!=']),
   value: paramOrContextParamSchema,
 });
 
+export const rpcReturnValueTestSchema = z.object({
+  index: z.number().int().nonnegative().optional(),
+  comparator: z.enum(['==', '>', '<', '>=', '<=', '!=']),
+  value: rpcParamOrContextParamSchema,
+});
+
```
derekpierre, line 19: Need to think on this some more.

At the very least we should have both of these schemas extend from a common schema that has `index` and `comparator`; that way only `value`  would be different and we would have `index` and `comparator` repeated

packages/taco/src/conditions/schemas/context.ts
Muhammad-Altabba: Yes, initially I started using names starting with evm. Like for this evmParamSchema. But I thing it will be confusing with the current naming convention for the schema: https://github.com/nucypher/taco-web/blob/74f8fe05b01f998b868bc5ebba7de942b66ed673/packages/taco/src/conditions/schemas/rpc.ts#L24 and https://github.com/nucypher/taco-web/blob/74f8fe05b01f998b868bc5ebba7de942b66ed673/packages/taco/src/conditions/schemas/rpc.ts#L50

What about using `evm...` for naming and rename `rpcConditionSchema` like this?:
```ts
export const evmConditionSchema = baseConditionSchema
  .extend...

/**
 * @deprecated use {@link evmConditionSchema} instead
 */
export const rpcConditionSchema = evmConditionSchema;

export type EvmConditionProps = z.infer<typeof evmConditionSchema>;
/**
 * @deprecated use {@link EvmConditionProps} instead
 */
export type RpcConditionProps = EvmConditionProps;

```

Using this the old schema name will still be useable but the user will get a warning that is depreciated. So, in later version we can remove the old name completely. 

Does this fits?

derekpierre: I think I'm missing the need to rename the Props export if the name of the schema changes? Is that what you are worried about. 

Perhaps we can have a call to chat more about this.

manumonti (approved):
LGTM! 🙌

PR #637, opened 2025-03-19
Support bigint and large numbers especially for evm parameters and return values

### LINKED ISSUE #600: Consider more restrictive schemas for EVM base types

### LINKED ISSUE #634: Support BigInt especially for EVM parameters and return values
A suggestion to add support for BigInt which need to be serialized and de-serialized at taco-web and the nodes. This is especially that int `number` cannot hold large int numbers that evm supports and return as `bigint`. 

Not handling bigints would cause currently an exception, if there was a bigint.

(The description above has been updated to point to int numbers instead of mentioning just numbers)

_Reviewers:_
@derekpierre @KPrasch

Muhammad-Altabba:
**Type of PR:**

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**

This PR is to solve [1 bug](https://github.com/nucypher/taco-web/issues/634#issuecomment-2734700957) and add [1 new feature](https://github.com/nucypher/taco-web/issues/634#issue-2912153613) by converting `bigint` and large numbers to hex strings (or smaller than the min negative number) when they are serialized. An integration test has been added which fails till TACo nodes would support this.

For TACo nodes to support this, the needed changes are made at https://github.com/nucypher/nucypher/pull/3585.

**Issues fixed/closed:**

 - Fixes #634
 - Fixes #600  (improves naming)

**Why it's needed:**

**Notes for reviewers:**


packages/shared/src/utils.ts
```diff
@@ -44,15 +44,22 @@ const sortedReplacer = (_key: string, value: unknown) => {
...
   if (value instanceof Uint8Array) {
     return `0x${toHexString(value)}`;
   }
+  if (
+    typeof value === 'bigint' ||
+    (typeof value === 'number' &&
+      (value > Number.MAX_SAFE_INTEGER || value < Number.MIN_SAFE_INTEGER))
+  ) {
+    return `0x${value.toString(16)}`;
+  }
```
derekpierre, line 57: Is the idea here that you would convert large ints into hex?

Until now, hex has only been used for transporting bytes in JSON.  I'm concerned about nodes being able to determine the difference between hex that represents bytes vs hex that represent integers just by looking at the hex...?

packages/taco/test/conditions/predefined/erc721.test.ts
```diff
@@ -44,7 +44,7 @@ describe('ERC721Balance', () => {
       chain: TEST_CHAIN_ID,
       returnValueTest: {
         comparator: '==',
-        value: '10',
+        value: BigInt(10),
```
derekpierre, line 47: This test only tests that BigInts were not previously accepted. It's not really testing that a large value would have failed in the past.

packages/shared/src/utils.ts
Muhammad-Altabba: Yes, I propose to convert to hex string. Except if there is some issue at an edge case. https://github.com/nucypher/taco-web/issues/634#issuecomment-2737114631

packages/taco/test/conditions/predefined/erc721.test.ts
Muhammad-Altabba: Yes, exactly. This test is for the new feature added (which is to support bigint).

packages/shared/src/utils.ts
cygnusv: > Is the idea here that you would convert large ints into hex?
> 
> Until now, hex has only been used for transporting bytes in JSON.  I'm concerned about nodes being able to determine the difference between hex that represents bytes vs hex that represent integers just by looking at the hex...?

Well that's not for the nodes to discern, but it's dictated by the ABI (or reference schema in the case of RPC methods).

packages/taco/test/conditions/lingo.test.ts
```diff
@@ -165,18 +164,21 @@ describe('check large numbers', () => {
...
+
     const contractCondition = new ContractCondition(
       uint256ContractConditionProps,
     );
     const conditionExpr = new ConditionExpression(contractCondition);
     expect(conditionExpr.toJson()).toContain(
-      '115792089237316195423570985008687907853269984665640564039457584007913129639935',
+      (115792089237316195423570985008687907853269984665640564039457584007913129639935).toString(
+        16,
+      ),
```
Muhammad-Altabba, line 176: We would use '1157.....9935' if we adopted converting the large numbers to strings. But I think converting them to hex string is better because it start with `0x`.

packages/shared/src/utils.ts
derekpierre: Need to think on this some more.

For `returnValueTest` we use `ast.literal_eval(...)` on both result and execution value, which will convert values into common types. This applies to other conditions as well. We added code to handle bytes as hex strings because bytes aren't allowed in JSON - https://github.com/nucypher/nucypher/pull/3303. So an int represented by a hex string is probable fine.

However for `parameters` (or `params` for non-blockchain conditions) we pass values as provided, instead of trying to perform conversion of the values. 

In general, we've typically avoided type conversions, because we aren't always going to have an ABI to check (non-onchain conditions). For something like a `JsonRPCCondition` that doesn't have an ABI, the node can't do conversions on its behalf and it expects parameter values to be passed in the types they are supposed to be in. If you pass in a string for an int when an int is expected, that may/may not fail when sending that param to the endpoint. 

It also prevents us from having inconsistent condition definitions (various values for the same parameter) that still work, and potentially complex code for type checking and type conversions. The creator of the condition understands best the correct types, so we expect them to provide the correct types. 

In web3py's case it does some internal type conversion on our behalf and the check is https://web3py.readthedocs.io/en/stable/web3.main.html#web3.w3.is_encodable.

For example, to pass bytes to web3py you can pass it as python `bytes` or as hex string:
```python
In [69]: w3.is_encodable("bytes", b"somebytes")
Out[69]: True

In [70]: w3.is_encodable("bytes", "0x736f6d656279746573")
Out[70]: True
```

However, if the value is not encodable as the expected type then it fails. So for uint256, while you can pass an int, you can't pass a hex string:
```python
In [72]: w3.is_encodable("uint256", "0x1234")
Out[72]: False

In [73]: w3.is_encodable("uint256", 115792089237316195423570985008687907853269984665640564039457584007913129639935)
Out[73]: True
```

So if a hex string is provided to the node for a contract parameter that expects uint256, the condition execution would fail due to a type mismatch.

derekpierre: Potentially having the int as a string could be simpler than hex. These `param` / `parameter` / `value`are open fields without type restrictions. Then at least I can check if the string can be converted to an int, and if not leave it as a string.

packages/taco/test/conditions/predefined/erc721.test.ts
derekpierre: I really meant that based on the commit message of - https://github.com/nucypher/taco-web/pull/637/commits/d8773d2d152942f045f8043ef7d92e2d0cae97da i.e. "modify a test that passes now but would fail before the PR" -> this would have failed because of a type mismatch.

I think that instead, you can add an additional test to confirm the ability to use BigInt instead of modifying an existing one - keeping original tests are good for ensuring regression testing. For new functionality we should add new tests.

packages/shared/src/utils.ts
derekpierre: Here's a possibility. For any numbers outside of the safe range, we use strings in the JSON, and numbers within the safe range (bigint or regular int) we continue to use them as numbers in the JSON - https://github.com/nucypher/taco-web/compare/epic-v0.6.x...derekpierre:taco-web:large-number-tests. 

Everything works as before for numbers provided within the safe range. For bigints (outside of the safe number range) they will be strings, and on the node side, we'll check if the string is convertible to an int, and if so convert it, if not, it'll just remain a string. Need to investigate the node changes required.

Thoughts?

Muhammad-Altabba: Thanks @derekpierre for the helpful discussion.

And regarding:
> Everything works as before for numbers or bigints provided within the safe range. For bigints (outside of the safe number range) they will be strings, and on the node side, we'll check if the string is convertible to an int, and if so convert it, if not, it'll just remain a string. Need to investigate the node changes required.


This is also what this PR is about except some bit minor details:
- I think hex string are bit better as checking the at the nodes would be: if the string start with `0x` try parse into a number (wether this check is with if or at a regex). Also hex string are shorter than number string. Also in all cases wether the large number is a string or a hex string, a type conversion is needed for uint256.

> w3.is_encodable("uint256", "115792089237316195423570985008687907853269984665640564039457584007913129639935")
Gives: False.
> w3.is_encodable("uint256", "0x11111")
Gives: False.
> w3.is_encodable("uint256", 115792089237316195423570985008687907853269984665640564039457584007913129639935)
Gives: True.

- Using Zod transform is nice. But I lean toward the separation between data validation and data serialization. By using Zod to specify the schema accepted form the user. And keep the serialization conversion to the time before sending the data. I think there is no right and wrong here. But I lean toward decoupling definition from transformation.

- Using number for the safe range of bigint sounds also good. I updated the code accordingly.

derekpierre: > I think hex string are bit better as checking the at the nodes would be: if the string start with 0x try parse into a number (wether this check is with if or at a regex

Bytes are also added to the JSON as hex strings as well (that's the original `u8ToHexReplacer` code), so for any hex string (starts with `0x`) it can either be a hex representation of bytes or now a hex representation of an int, but the node can't tell the difference because parsing a hex string as either an int or bytes will work, and the resulting type matters.

```python
In [4]: int("0xdeadbeef", 16)
Out[4]: 3735928559

In [5]: bytes(HexBytes("0xdeadbeef"))
Out[5]: b'\xde\xad\xbe\xef'
```

As opposed to having it be string in the JSON where the node can do the following:
```python
try:
    result = int(value)
    return result
except ValueError:
    # regular string
    pass
    
return value
```
i.e. either the parsing of the string to an int works because it was a number, or it's just a regular string.


> Using Zod transform is nice. But I lean toward the separation between data validation and data serialization. By using Zod to specify the schema accepted form the user. And keep the serialization conversion to the time before sending the data. I think there is no right and wrong here. But I lean toward decoupling definition from transformation.

That's fair.

packages/taco/src/conditions/schemas/context.ts
```diff
@@ -11,12 +11,17 @@ export const contextParamSchema = z
...
+const paramSchema = z.union([
+  plainStringSchema,
+  z.boolean(),
+  z.number(),
+  z.bigint(),
+]);
 
 const nonFloatParamSchema = z
-  .union([plainStringSchema, z.boolean(), z.number().int()])
+  .union([plainStringSchema, z.boolean(), z.number().int(), z.bigint()])
```
derekpierre, line 22: I think we need to refine this bigint, because it shouldn't allow values > uint256 max or values < int256 min for onchain conditions - see https://github.com/derekpierre/taco-web/blob/large-number-tests/packages/taco/src/conditions/schemas/context.ts#L35. 

That may beg the follow-up question as to whether `nonFloatParamSchema` is still the best name for this schema.

packages/taco/test/conditions/context.test.ts
```diff
@@ -122,7 +123,9 @@ describe('context', () => {
       const condition = new RpcCondition(testRpcConditionObj);
       const conditionContext = new ConditionContext(condition);
       expect(
-        JSON.stringify(condition.toObj()).includes(USER_ADDRESS_PARAM_DEFAULT),
+        JSON.stringify(condition.toObj(), toHexReplacer).includes(
```
derekpierre, line 126: I think for here and in other places, anywhere that we are calling `JSON.stringify(...)` directly, should be replaced by the call to `toJSON(...)` from utils.

packages/taco/src/conditions/schemas/context.ts
Muhammad-Altabba: Good point!
Also checked for the int number to be in the same range.

packages/taco/integration-test/encrypt-decrypt.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    BigInt(10000000000000000), // 0.01 ETH/PLY
+  ])(
+    'should encrypt and decrypt a message with condition value %s',
+    async (value) => {
+      // Create test message
+      const messageString = 'This is a secret 🤐';
+      const message = toBytes(messageString);
+
+      // Create condition
+      const hasPositiveBalance = new conditions.base.rpc.RpcCondition({
```
derekpierre, line 84: I think you can just do a single `CompoundCondition` (`AND`) comprised of multiple rpc conditions each using the different values. Then you can just run this once.

Also I would say once of those should be a check for a balance < uint256 max i.e.
```typescript
returnValueTest: {
    comparator: "<",
    value: BigInt("115792089237316195423570985008687907853269984665640564039457584007913129639935"),
}
```

Muhammad-Altabba: The suggestion on using `CompoundCondition` I think it is great for the case.
And yes, it would be overkill to repeat the full test just for different condition values. I basically made it like this to make it easy for this to be checked with the nodes as part of the issue scope definition.

Probably for an integration test it is good to use extensive complicated conditions that covers almost all conditions. Would be good to craft such when doing https://github.com/nucypher/taco-web/issues/638.

packages/taco/test/conditions/lingo.test.ts
```diff
@@ -132,3 +134,94 @@ describe('check that valid lingo in python is valid in typescript', () => {
...
 });
+
+describe('check large numbers serialization into hex string', () => {
+  const numbersToTest = [
+    {
+      name: 'uint256 large positive',
+      // The max uint256 cannot be encoded correctly in javascript becuase it is too large.
+      // For that some arbitraty extra large positive number is used:
+      //  11579208923731619542357098500868790785326998466564056403945758400791312963993
+      // Which is less by a littile from the max uint256 (note 5 at the end):
```
manumonti:
```suggestion
      // Which is less by a little from the max uint256 (note 5 at the end):
```

manumonti (approved):
LGTM! I have just catched a small typo, but not a blocker

derekpierre:
> Thanks @derekpierre for your nice contributions.
> 
> I have no more thoughts to apply to this PR. Feel free to approve and merge or let me know if you would like me to add something.

👍 . Thank you as well - it was a nice tag-team effort.

Nothing to add, just looking for reviews from others. My review is implicit since I added commits.

PR #646, opened 2025-04-09
Port shared secret encryption from nucypher-core

### LINKED ISSUE #635: Investigate pure TS client-side implementation
Let's recall our current stack:

- `ferveo`: Core cryptographic implementation (Rust -> WASM)
- `nucypher-core`: Protocol API and objects (Rust -> WASM)
- `taco-web`: Adopter-facing API (TypeScript)

The scope of this issue is to investigate to what extent we can build a **minimal** substitution for `ferveo` and `nucypher-core` using purely TypeScript primitives. Since browser interactions are currently tied to the creator/encryption and consumer/decryption request steps, let’s investigate what happens underneath and what do we need to replace. 

**Encryption** (`taco/src/tdec.ts/encryptMessage()`):

- `nucypher_core::encryptForDkg(data: Uint8Array, public_key: DkgPublicKey, conditions: Conditions)`
    - `ferveo_tdec::api::encrypt(message, aad, public_key, rng)`:
        - BLS12-381 arithmetic
        - ChaCha20Poly1305 encryption
        - Rust-side serialization
- `nucypher_core::AccessControlPolicy`
- `nucypher_core::ThresholdMessageKit`

**Decryption** (`taco/src/tdec.ts/retrieveAndDecrypt()`):

- `nucypher_core::`
    - `ThresholdDecryptionRequest`
        - `.encrypt()`
            - `EncryptedThresholdDecryptionRequest`
                - `encrypt_with_shared_secret`
                    - ChaCha20Poly1305 encryption
    - `SessionStaticSecret`
        - `.new()`
            - `crate::secret_box::kdf` (HKDF)
        - `.derive_shared_secret()`
            - `x25519_dalek::{PublicKey, SharedSecret, StaticSecret}`
    - `EncryptedThresholdDecryptionRequest`
        - `encrypt_with_shared_secret`
        - `decrypt_with_shared_secret`
    - `ThresholdMessageKit`
        - .`decryptWithSharedSecret()`
            - `ferveo::api::decrypt_with_shared_secret`
                - SecretBox
                - Chacha20Poly1305
- `ferveo::binding_wasm::combine_decryption_shares_simple`
    - `ferveo::api::combine_shares_simple`
        - BLS12-381 arithmetic

After this preliminary analysis, we can see that we currently depend on these external primitives:

- BLS12-381 arithmetic
- ChaCha20Poly1305 encryption
- HKDF
- `x25519_dalek`
- SecretBox
- Rust-side serialization

A cursory search I can see several 3rd party libraries that can help us with this process, but we need a more in-depth investigation.

- BLS12-381 and x25519 pure TS/JS implementation:
https://github.com/paulmillr/noble-curves/blob/94d00668f19ea38a6956cc0afe1051038b764fcb/README.md?plain=1#L46
- ChaCha20Poly1305 implementation:
https://github.com/paulmillr/noble-ciphers
- HKDF: https://github.com/panva/hkdf

Muhammad-Altabba:
**Type of PR:**

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**

- It contains ported code for the 2 Rust methods: `encrypt_with_shared_secret` and `decrypt_with_shared_secret`. 
- A compatibility test run against test vectors generated from https://github.com/nucypher/nucypher-core/pull/102.

**Issues fixed/closed:**

- Closes part of https://github.com/nucypher/sprints/issues/150. It only covers encrypt/decrypt with shared secret. Subsequent PR is to be opened for the rest of the needed functions.


packages/shared/src/encryption/shared-secret.ts
```diff
@@ -0,0 +1,44 @@
...
+  DecryptionErrorCode,
+  EncryptionError,
+  EncryptionErrorCode,
+} from './errors';
+
+export function encryptWithSharedSecret(
+  sharedSecret: Uint8Array,
+  plaintext: Uint8Array,
+): Uint8Array {
+  const nonce = randomBytes(12); // Generate a 12-byte nonce
```
derekpierre, line 15: Is there a const for `12` somewhere, or one that can be created?

packages/shared/test/fixtures/shared-secret-vectors.json
```diff
@@ -0,0 +1,235 @@
+{
+  "test_vectors": [
+    {
+      "id": "vector1",
+      "description": "Fixed nonce encryption with known plaintext",
+      "shared_secret": [
+        0,
```
derekpierre, line 7: Is this an array of int bytes? Can this be in hex instead? Same for the nonce, and rust ciphertext? 

Or is there a reason for keeping this format?

packages/shared/test/fixtures/shared-secret-vectors.json
```diff
@@ -0,0 +1,235 @@
+{
+  "test_vectors": [
```
derekpierre, line 2: I noticed the fixed nonce, specific shared secret - does it make sense to also have more random values in the test vector?  Or is that something planned for later in the process?

packages/shared/test/shared-secret-compatibility.test.ts
```diff
@@ -0,0 +1,271 @@
...
+      // Set up our mock to return the fixed nonce
+      const cleanupMock = setupFixedNonceMock(fixedNonce);
+
+      try {
+        // Encrypt with fixed nonce
+        const ciphertext = encryptWithSharedSecret(sharedSecret, plaintext);
+
+        // Check against expected ciphertext from the JSON file
+        const ciphertextHex = Buffer.from(ciphertext).toString('hex');
+        if (expectedCiphertext) {
```
derekpierre, line 101: Is there a case where this wouldn't be set for the test vectors i.e. do we need this if-else block? Would the test simply fail spectacularly if `Buffer.from(expectedCiphertext)` (L102) is called on an undefined object, and that would be sufficient?

packages/shared/test/shared-secret-compatibility.test.ts
```diff
@@ -0,0 +1,271 @@
...
+        expect(Buffer.from(decrypted).toString()).toEqual(
+          Buffer.from(plaintext).toString(),
+        );
+      } finally {
+        // Always restore the original implementation
+        cleanupMock();
+      }
+    });
+
+    it('should correctly encrypt and decrypt test vector 2 (empty plaintext)', () => {
```
derekpierre, line 128: Given that we have test vectors, how come we need separate `it` blocks for each vector? Are all the test vectors not structured the same? If not, can we set it up to not need repeated code - can `it.each` be leveraged?

packages/shared/test/shared-secret-compatibility.test.ts
```diff
@@ -0,0 +1,271 @@
...
+      } finally {
+        // Always restore the original implementation
+        cleanupMock();
+      }
+    });
+  });
+
+  describe('Cross-implementation compatibility tests', () => {
+    // Compatibility test for decryption of pre-generated ciphertext from Rust
+    it('should correctly decrypt ciphertext generated by the Rust implementation', () => {
```
derekpierre, line 170: Does this need to be its own `it` block, or could this simply be another structured test vector? Any reason Rust implementation needs to be separate from the other test vectors?

packages/shared/test/shared-secret.test.ts
```diff
@@ -0,0 +1,221 @@
...
+  decryptWithSharedSecret,
+  encryptWithSharedSecret,
+} from '../src/encryption/shared-secret';
+
+// Mock the noble/hashes/utils randomBytes function
+vi.mock('@noble/hashes/utils', () => {
+  // Counter to ensure each call generates different values
+  let counter = 0;
+  return {
+    randomBytes: (size: number) => {
```
derekpierre, line 18: For my benefit, could you clarify what this is doing?

packages/shared/src/encryption/shared-secret.ts
Muhammad-Altabba: Sure! The TS code is updated accordingly now.

(There is no constant for the original Rust code: `pub type Nonce = GenericArray<u8, U12>;`. So, if you think so, an issue at nucypher-core could be created to suggest having this as somehow a constant)

packages/shared/test/fixtures/shared-secret-vectors.json
Muhammad-Altabba: The values are randomly generated at the Rust code and fed as constants to the compatibility tests. For values that are randomly generated in this project, there is the unit test in the other file.

packages/shared/test/shared-secret.test.ts
Muhammad-Altabba: This is to mock this function to make it give random but deterministic values (simi-random). Which is to make debugging the tests easy when needed. Because for every run of the tests the same simi-random value will be used.

packages/shared/src/encryption/shared-secret.ts
derekpierre: Creating a simple constant like you did in the ts file is sufficient. Thanks!

packages/shared/test/fixtures/shared-secret-vectors.json
derekpierre: 👍 - keeping it similar to the Rust implementation is a very fair point.

packages/shared/test/fixtures/shared-secret-vectors.json
derekpierre: > The values are randomly generated at the Rust code and fed as constants to the compatibility tests.

Perhaps I'm misunderstanding. The values for the nonce, shared secret don't seem that "random"...? In one case the nonce is just 0s or 1s and the shared secret is just ints 0-31...?

Muhammad-Altabba: Yeah, sorry my bad. I forgot that I updated this in Rust. Yes, the values are fixed. But since the values does not have special numerical meanings, in relation to the algorithm, they are still counted as random (or you may say simi-random).
However, we can generate different values every time. But it will make debugging harder in case of an issue.

Hey @cygnusv, Could you please, provide your input if testing with new random values is needed for this cryptography validation.

packages/shared/test/shared-secret.test.ts
derekpierre: > random but deterministic values (simi-random)

Does the test only work for a specific order of randomness - is there some limitation in the test that requires determinism?

My overall concern is about the test not actually using proper randomness, which is what would occur in the wild. Just trying to understand why encryption/decryption testing needs to use deterministic keys/randomness...

Similar comment regarding why the shared secret needs to be deterministic (L36-L39)?

packages/shared/test/shared-secret-compatibility.test.ts
Muhammad-Altabba: You are very right. There is no real need for separate `it` blocks. I unified the vectors fileds keeping just the minimal differences. And used `it.each`. However, I kept the code checking the errors as is.

packages/shared/test/fixtures/shared-secret-vectors.json
Muhammad-Altabba: Great!
Also the code accept both Uint8Array and hex string. It will be up to the Rust-generated json file to have the data in the convenient type.

packages/shared/test/fixtures/shared-secret-vectors.json
Muhammad-Altabba: Would like continue this the conversation about the generated test vectors at https://github.com/nucypher/nucypher-core/pull/105.
Thanks,

packages/shared/test/shared-secret.test.ts
```diff
@@ -0,0 +1,221 @@
...
+    plaintext = new TextEncoder().encode(
+      'This is a test message for encryption and decryption.',
+    );
+  });
+
+  describe('encryptWithSharedSecret', () => {
+    it('should encrypt plaintext successfully', () => {
+      const encrypted = encryptWithSharedSecret(sharedSecret, plaintext);
+
+      // Ensure the encrypted result is longer than the plaintext (nonce + ciphertext + tag)
```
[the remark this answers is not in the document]
Muhammad-Altabba: The "tag" refers to the authentication tag (also called MAC) produced by the ChaCha20-Poly1305 cipher. It is a 16-byte value appended to the ciphertext that ensures the integrity and authenticity of the encrypted data. If the tag is incorrect or missing, decryption will fail.

packages/shared/test/shared-secret-compatibility.test.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  describe('Vector-based compatibility tests', () => {
+    it.each(testVectors)(
+      'should correctly encrypt and decrypt test vector $id: $description',
+      (vector) => {
+        const { sharedSecret, plaintext, fixedNonce, expectedCiphertext } =
+          vector;
+
+        if (fixedNonce && sharedSecret) {
```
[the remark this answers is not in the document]
Muhammad-Altabba: [Sorry, it should be only checking for `fixedNonce`. The code is now updated.]
And this is just to give the confidence that even if the test vectors did not contain the nonce (so the encryption of data would not be possible, and the decryption is the only available), the plaintext could be retrieved by decrypting the cipher text. 

However, if this case is unnecessary to check, the https://github.com/nucypher/nucypher-core/pull/105 would not generate it and I will remove this `if` from here 😄 .

packages/shared/test/shared-secret.test.ts
Muhammad-Altabba: Many thanks @derekpierre for the discussion.
I did more search and I am more now in favor of having deterministic values for the randomness of our tests here. And probably if we are doing the cryptographic library we would keep this simi-random but have additional completely random values that change with every run.

However, since I am not very much into cryptography related stuff, I will leave this to you (@derekpierre) and @cygnusv to recommend. 
When I first showed @cygnusv this way of generating the nonce, I think he liked it. But I hope him to provide his input, if it is better for it to be like this, or if not, after he read this discussion.

Thanks all,

packages/shared/test/fixtures/shared-secret-vectors.json
cygnusv: Sorry for being a bit late to the discussion. I've been following along while working on https://github.com/nucypher/nucypher-core/pull/105 and I changed my mind a few times during this process. The production and consumption of test vectors are a bit more complicated than expected due to how our internal protocol objects work, since in some cases serialization/deserialization is more difficult or not possible without adding code. I'm still iterating a proper method for producing test vectors in my PR, but on the PRs in taco-web we should strive to consume test vectors without resorting to duplicated implementations.

cygnusv: With respect to the specific issue discussed here, the test vector ideally should be generated in a 100% deterministic way (i.e. using a seed). However, there's nothing wrong about adding some hard-coded, obviously not random examples, like `[1, 1, 1, .....1]`.

manumonti:
Removing myself as reviewer while this is not ready to be reviewed. Don't hesitate to add me if there is progress on this!

PR #664, opened 2025-06-02
Add signing functionality to sdk

theref:


packages/shared/src/porter.ts
```diff
@@ -152,6 +152,52 @@ export type TacoDecryptResult = {
...
+  digest: string;
+  aggregatedSignature: string;
+  signingResults: { [ursulaAddress: string]: [string, string] };
+  type: string;
+};
+
+function aggregatePorterSignatures(
+  signaturesWithAddress: { [checksumAddress: string]: [string, string] }
+): string {
+  const sortedSignaturePairs = Object.entries(signaturesWithAddress)
```
derekpierre: Porter already sorts the results, so you don't need to sort them anymore.

packages/taco/src/sign.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import { SigningOptions, SignResult, UserOperation } from './types';
+
+// Core signing functions
+export async function sign191(
+  payload: Uint8Array | string,
+  cohortId: number,
+  domain: Domain,
+  options: SigningOptions = { optimistic: true, returnAggregated: true }
+): Promise<SignResult> {
```
derekpierre, line 25: You also need to take `porterUris?: string[]`  - this is how we ensure that any porter instance can be used and not just the defaults for the domain.

See https://github.com/nucypher/taco-web/blob/main/packages/taco/src/taco.ts#L140-L145

Same for other sign API functions below.

packages/taco/src/sign.ts
```diff
@@ -0,0 +1,59 @@
...
+    signingResults: result.signingResults,
+    type: result.type
+  };
+}
+
+export async function signUserOp(
+  userOp: UserOperation,
+  chainId: number,
+  accountSpec: "zerodev" | "kernel" | "metamask" | "safe" | string,
+  entryPointVersion: "v0.6" | "v0.7" | "v0.8",
```
derekpierre: I think these two parameters will be subsumed into a single `aaVersion` value.

packages/taco/src/types.ts
```diff
@@ -3,3 +3,32 @@ export {
...
+  };
+  type: string;        // Signature type (e.g. "eip712", "userOp:zerodev")
+};
+
+export type SigningOptions = {
+  optimistic?: boolean;        // Whether to proceed with signing even if some nodes are unavailable
+  returnAggregated?: boolean;  // Whether to return the aggregated signature
+};
+
+export type UserOperation = {
```
derekpierre: We need to update this type to match the latest spec - https://eips.ethereum.org/EIPS/eip-4337#useroperation.

What was in `nucypher` was not correct and is being updated via https://github.com/nucypher/nucypher/pull/3604

packages/taco/test/tao-sign.test.ts
derekpierre, line 1: Rename file - `tao-sign.test.ts` -> `taco-sign.test.ts`.

packages/shared/src/porter.ts
```diff
@@ -284,4 +330,88 @@ export class PorterClient {
...
+      return_aggregated: options.returnAggregated,
+    };
+
+    const resp: AxiosResponse<SignResponse> = await this.tryAndCall({
+      url: '/sign191',
+      method: 'post',
+      data,
+    });
+
+    const signingResults: { [ursulaAddress: string]: [string, string] } = {};
```
derekpierre, line 382: You have to be careful here because errors can be returned in the signing result i.e. the Porter HTTP call result contains errors. See how `decrypt` is handled.

Similar comment for `signUserOp`.

derekpierre:
Btw really nice start on the API 🎸

packages/shared/src/porter.ts
```diff
@@ -284,4 +311,52 @@ export class PorterClient {
...
+    let messageHash = '';
+
+    for (const [ursulaAddress, [signerAddress, signatureB64]] of Object.entries(
+      resp.data.result.signing_results
+    )) {
+      try {
+        const decodedData = JSON.parse(
+          new TextDecoder().decode(fromBase64(signatureB64))
+        );
+        messageHash = decodedData.message_hash;
```
derekpierre: For additional safety, perhaps we should verify that all the message hashes are the same? wdyt?

packages/shared/test/porter.test.ts
```diff
@@ -49,6 +51,46 @@ const mockGetUrsulas = (ursulas: Ursula[] = fakeUrsulas()): MockInstance => {
...
+          data: createMockSignResponse(type, errorCase),
+        });
+      }
+      return Promise.resolve({ status: HttpStatusCode.BadRequest, data: '' });
+    }
+    throw new Error('Unexpected endpoint');
+  });
+};
+
+const mockSignUserOp = createMockSignImplementation('/sign', 'userOp:zerodev');
```
derekpierre: Is the "type" value here (`userOp:zerodev`) still needed?

packages/shared/test/porter.test.ts
```diff
@@ -117,3 +159,103 @@ describe('PorterClient', () => {
...
+      );
+
+      expect(result).toEqual({
+        digest: '0x1234',
+        aggregatedSignature: '0xijkl', // Only valid signature included
+        signingResults: {
+          '0xabcd': ['0xefgh', '0xijkl'],
+        },
+        errors: {
+          '0x1234': 'Failed to decode signature: Invalid character',
```
derekpierre: Need to better understand this - an aggregated signature is produced when there wasn't a threshold of signatures.

packages/taco/integration-test/sign.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+import { initialize } from '../src';
+import { signUserOp } from '../src/sign';
+import { UserOperation } from '../src/types';
+
+const RPC_PROVIDER_URL = 'https://rpc-amoy.polygon.technology';
+const SIGNER_PRIVATE_KEY = '0x900edb9e8214b2353f82aa195e915128f419a92cfb8bbc0f4784f10ef4112b86';
+const DOMAIN = 'lynx';
+const RITUAL_ID = 27;
+const CHAIN_ID = 80002;
```
derekpierre: This test won't work with these values:
- the signing coordinator is on `sepolia` and not `amoy` so the provider needs to be for sepolia.
- `27` is not a valid cohort id for signing, `2` should be used
- cohort id `2` is not configured on amoy (`80002`), but it is for sepolia (`11155111`)

packages/taco/integration-test/sign.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    // Verify network connection
+    const network = await provider.getNetwork();
+    if (network.chainId !== CHAIN_ID) {
+      throw new Error(
+        `Provider connected to wrong network. Expected ${CHAIN_ID}, got ${network.chainId}`,
+      );
+    }
+  });
+
+  test.skip('should sign a user operation', async () => {
```
derekpierre: Can we unskip this test now? Is it possible for the integration test to actually run?

packages/taco/integration-test/sign.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // Get porter URIs
+    const porterUris = await getPorterUris(DOMAIN as Domain);
+
+    // Sign user operation
+    const signResult = await signUserOp(
+      provider,
+      userOp,
+      CHAIN_ID,
+      'zerodev:v0.6',
```
derekpierre: This should just be AA version for now, right?

packages/taco/src/sign.ts
```diff
@@ -0,0 +1,47 @@
...
+export async function signUserOp(
+  provider: ethers.providers.Provider,
+  userOp: UserOperation,
+  chainId: number,
+  aaVersion: "mdt" | "0.8.0" | string,
+  cohortId: number,
+  domain: Domain,
+  options: SigningOptions = { optimistic: true, returnAggregated: true },
+  context?: ConditionContext,
+  porterUris?: string[]
```
derekpierre: Bit of a nit-pick, can we reorder the params? My thinking here is to go from more general -> more specific values

```suggestion
  provider: ethers.providers.Provider,
  domain: Domain,
  cohortId: number,
  chainId: number,
  userOp: UserOperation,
  aaVersion: "mdt" | "0.8.0" | string,]
  options: SigningOptions = { optimistic: true, returnAggregated: true },
  context?: ConditionContext,
  porterUris?: string[]
```

packages/taco/src/sign.ts
```diff
@@ -0,0 +1,47 @@
...
+    aaVersion,
+    context
+  };
+  
+  const signingRequests: Record<string, string> = Object.fromEntries(
+    signers.map((signer) => [signer.operator, JSON.stringify(signingRequest)]),
+  );
+
+  // Build signing request for the user operation
+  const result = await porter.signUserOp(
```
derekpierre: 👍

packages/taco/test/taco-sign.test.ts
```diff
@@ -0,0 +1,214 @@
...
+      maxFeePerGas: '0xabc',
+      maxPriorityFeePerGas: '0x123',
+      paymaster: '0x0',
+      paymasterVerificationGasLimit: '0x0',
+      paymasterPostOpGasLimit: '0x0',
+      paymasterData: '0x',
+      signature: '0x',
+    };
+
+    it('should sign a user operation with zerodev account', async () => {
```
derekpierre: zero dev and kernel aren't real types. Any reason you need 2 tests instead of just 1. How about 1 test that uses "0.8.0".

packages/taco/test/taco-sign.test.ts
```diff
@@ -0,0 +1,214 @@
...
+          '0x1234': ['0x5678', '0x90ab'],
+          '0xabcd': ['0xefgh', '0xijkl'],
+        },
+        errors: {},
+      });
+    });
+
+    it('should handle custom signing options', async () => {
+      const chainId = 1;
+      const aaVersion = 'safe:v0.8';
```
derekpierre: Let's use currently valid types:

```suggestion
      const aaVersion = '0.8.0';
```

derekpierre:
@theref - https://github.com/nucypher/taco-web/pull/664/commits/d34a58c3a2875de2c0c7abc0c5a19cf0fe07ba02.

Apparently github dependencies for packages don't get automatically installed (local installation does) Instead I generated a tarball (using `npm pack`) of the latest in `nucypher-contracts`:`signing` and included it in the repos. When we have a released version of `nucypher-contracts` on npm we can then update the dependency and delete the tarball.

It builds, and the signing integration test just needs to be fixed.

packages/shared/src/contracts/agents/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './coordinator';
 export * from './global-allow-list';
 export * from './subscription-manager';
+export * from './signingCoordinator';
```
derekpierre: Let's keep consistent with our naming structure. wdyt about renaming the file and therefore changing this to:
```suggestion
export * from './signing-coordinator';
```

packages/shared/src/porter.ts
```diff
@@ -284,4 +326,75 @@ export class PorterClient {
...
+      console.error('Error in tryAndCall:', error);
+      throw error;
+    }
+
+
+    const signingResults: { [ursulaAddress: string]: [string, string] } = {};
+    const errors: Record<string, string> = {};
+    let messageHash = '';
+
+    // Check if signing_results exists and has the expected structure
```
derekpierre: In the response, it's possible to have a combination of success results and errors at the same time. The question is whether we got a threshold of successes or not.

So, I think we need to update our logic here a bit:
1. If there are no results, then the errors returned in the SigningResult become important. What is the expectation for this case, currently you raise an Error, but the `SigningResult` allows you to include errors in it? Perhaps the caller of Porter can decide based on the `SigningResult` whether to raise the error or not, instead of Porter. This was the strategy taken by `tacoDecrypt` - see https://github.com/nucypher/taco-web/blob/main/packages/taco/src/tdec.ts#L109-L115.

For example you could instead return:

```typescript
SigningResult
{
     messageHash: <message_hash>
    aggregatedSignature: undefined;
    signingResults: { }; // empty
    errors: <various errors returned>;
}
```

2. If there are not a threshold of results, then the errors are important as well (and there should be no aggregated signature). So you could return something like:

```typescript
SigningResult
{
     messageHash: <message_hash>
    aggregatedSignature: undefined;
    signingResults: { <less than a threshold set of results> };
    errors: <various errors returned>;
}
```

3. If there are a threshold of results, only then do you include the aggregated signature, but to be consistent with the above you should also include any errors"

```typescript
SigningResult
{
     messageHash: <message_hash>
    aggregatedSignature: <aggregatedSignature>;
    signingResults: { <threshold set of results> };
    errors: <various errors returned>;
}
```

In a nutshell, if you want to return a `SigningResult`, you would:
1. Not throw an Error in Porter
2. always include any errors or results returned
3. only include aggregated signature if there are a threshold of successful results.

packages/taco/src/sign.ts
```diff
@@ -0,0 +1,68 @@
...
+    cohortId,
+  );
+  console.log({ signers });
+  
+  const threshold = await SigningCoordinatorAgent.getThreshold(
+    provider,
+    domain,
+    cohortId,
+  );
+
```
derekpierre, line 43: I was thinking about this some more, and since conditions are currently on-chain, you need to verify that any values needed in the context are already there. So a couple things:
1. How would the user craft the context to provide to this function?
2. How does/should this function verify that the context is properly populated?

For 1), to create the condition context the user needs the ConditionContext to be created from the relevant condition for the `cohortId, chainId` pair. The condition could be retrieved from the SigningCoordinatorAgent. We may need a helper method here for crafting the ConditionContext. 

For 2) this is done by calling `await context.toContextParameters()` which, thinking about it as I'm writing this, you need to call on line 53. `context` is not a dictionary, but the `ConditionContext` class. It's not a dictionary (record) until you call that function.

I'm ok with it if you want to address this update in a separate PR.

theref: i know we chatted about this a bit before. What's the reasoning behind the context being added at this stage, rather than nodes fetching it from the contract themselves?

derekpierre: > i know we chatted about this a bit before. What's the reasoning behind the context being added at this stage, rather than nodes fetching it from the contract themselves?

Nodes cannot fetch the `context` because it contains user-specific values required to populate placeholders (context variables) like `:userAddress` in conditions, which are only available at the time of decryption or signing. These values must be provided by the user, and nodes don’t have direct access to the user to obtain them.

If you were referring to the "condition" rather than the `context`, then the nodes can and do retrieve the condition from the contract. However, if the condition includes placeholders that require user-provided values, the nodes rely on the context to supply those at runtime.

For example, if the condition includes `:userAddress`, the application must prompt the user to sign a SIWE message and then send the signed message to the node via the `context`.

Another example is @KPrasch 's ECDSA condition used for cohort ID 2 on chain 84532, which uses a `:signature` placeholder whose value is sent at signing request time. The value needs to be provided via the `context` for nodes to verify the signature against the hardcoded verifying key in the condition.

theref: Yep, got it. So I need to build the functionality that:
1. checks the conditions from the contract
2. builds the context that's required for those conditions
3. adds the context to the request

derekpierre: Mostly, yes. `ConditionContext` is passed into this method, so you need a way for the application to create the context beforehand and subsequently populate it as needed.

For decryption this was done by the application in the following way:
```typescript
context = ConditionContext.fromMessageKit(thresholdMessageKit)
```
(e.g. https://github.com/nucypher/taco-web/blob/main/examples/taco/nodejs/src/index.ts#L101)

However, this can be done because the conditions are within the ThresholdMessageKit object (decryption case). For signing it is on-chain so you'll need a provider.

So you could do something like
```typescript
context = ConditionContext.forSigningCohort(provider, cohortId, chainId)
...
```
OR
You could have an API-level function like
```typescript
context = createContextForSigningCohort(provider, cohortId, chainId)
...
```
OR
...

Not sure how we want to do this, and I'm not being prescriptive, but something you can determine what seems best.

theref: just looking at the contract and how conditions are stored. I think these 2 functions are duplicates?
```solidity
    function getSigningCohortConditions(
        uint32 cohortId,
        uint256 chainId
    ) external view returns (bytes memory) {
        SigningCohort storage signingCohort = signingCohorts[cohortId];
        require(isCohortActive(signingCohort), "Cohort not active");
        return signingCohort.conditions[chainId];
    }
    
    function getCondition(uint32 cohortId, uint256 chainId) external view returns (bytes memory) {
        SigningCohort storage cohort = signingCohorts[cohortId];
        return cohort.conditions[chainId];
    }
    ```

File: packages/shared/src/porter.ts [hunk 19]
```diff
@@ -284,4 +326,48 @@ export class PorterClient {
...
+    threshold: number,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _options: SigningOptions = { optimistic: true, returnAggregated: true },
+  ): Promise<SignResult> {
+    const data: Record<string, unknown> = {
+      signing_requests: signingRequests,
+      threshold: threshold
+    };
+    
+    const resp: AxiosResponse<SignResponse> = await this.tryAndCall({
```
derekpierre, line 372: _Reposting this comment which was lost by github somehow. I think some of the code has changed since I posted but reposting anyway_


In the response, it's possible to have a combination of success results and errors at the same time. The question is whether we got a threshold of successes or not.

So, I think we need to update our logic here a bit:
1. If there are no results, then the errors returned in the SigningResult become important. What is the expectation for this case, currently you raise an Error, but the `SigningResult` allows you to include errors in it? Perhaps the caller of Porter can decide based on the `SigningResult` whether to raise the error or not, instead of Porter. This was the strategy taken by `tacoDecrypt` - see https://github.com/nucypher/taco-web/blob/main/packages/taco/src/tdec.ts#L109-L115.

For example you could instead return:

```typescript
SigningResult
{
     messageHash: <message_hash>
    aggregatedSignature: undefined;
    signingResults: { }; // empty
    errors: <various errors returned>;
}
```

2. If there are not a threshold of results, then the errors are important as well (and there should be no aggregated signature). So you could return something like:

```typescript
SigningResult
{
     messageHash: <message_hash>
    aggregatedSignature: undefined;
    signingResults: { <less than a threshold set of results> };
    errors: <various errors returned>;
}
```

3. If there are a threshold of results, only then do you include the aggregated signature, but to be consistent with the above you should also include any errors"

```typescript
SigningResult
{
     messageHash: <message_hash>
    aggregatedSignature: <aggregatedSignature>;
    signingResults: { <threshold set of results> };
    errors: <various errors returned>;
}
```

In a nutshell, if you want to return a `SigningResult`, you would:
1. Not throw an Error in Porter
2. always include any errors or results returned
3. only include aggregated signature if there are a threshold of successful results.

packages/shared/src/porter.ts
```diff
@@ -284,4 +326,48 @@ export class PorterClient {
...
+    // Process signatures - decode the JSON payload to extract message hash and signature
+    for (const [ursulaAddress, [signerAddress, signatureB64]] of Object.entries(signatures || {})) {
+      const decodedData = JSON.parse(
+        new TextDecoder().decode(fromBase64(signatureB64))
+      );
+      messageHash = decodedData.message_hash;
+      signingResults[ursulaAddress] = [signerAddress, decodedData.signature];
+    }
+
+    const aggregatedSignature = Object.keys(signingResults).length > 0
```
derekpierre: We should only provide an aggregate id there are sufficient signatures i.e. a threshold of them.
```suggestion
    const aggregatedSignature = Object.keys(signingResults).length > threshold 
```

packages/shared/src/porter.ts
```diff
@@ -284,4 +326,48 @@ export class PorterClient {
...
+    
+    const signingResults: { [ursulaAddress: string]: [string, string] } = {};
+    let messageHash = '';
+
+    // Process signatures - decode the JSON payload to extract message hash and signature
+    for (const [ursulaAddress, [signerAddress, signatureB64]] of Object.entries(signatures || {})) {
+      const decodedData = JSON.parse(
+        new TextDecoder().decode(fromBase64(signatureB64))
+      );
+      messageHash = decodedData.message_hash;
```
derekpierre: Should we validate that all the messageHashes returned are the same value for added safety? something like -
```suggestion
      if (messageHash && messageHash !== decodedData.message_hash) {
           throw Error("Mismatched message hashes");
      }
      messageHash = decodedData.message_hash;
```
wdyt?

packages/taco/src/conditions/context/context.ts
```diff
@@ -55,7 +57,9 @@ const EXPECTED_AUTH_PROVIDER_TYPES: Record<string, AuthProviderType[]> = {
   ],
 };
 
-export const RESERVED_CONTEXT_PARAMS = [USER_ADDRESS_PARAM_DEFAULT];
+const INTERNAL_SIGNING_CONDITION = ':signingConditionObject';
```
derekpierre: Thanks for starting this. Is it possible to remove it, or is it needed for testing? I took care of it in https://github.com/nucypher/taco-web/pull/670 including special logic in the context preventing the user to need to specify it i.e. it will automatically be set by the node, and doesn't need to be handled on the client side.

packages/taco/src/sign.ts
derekpierre: 👍 . Good catch. Removed `getCondition` - https://github.com/nucypher/nucypher-contracts/pull/385.

Note that `getSigningCohortConditions` will fail the require check if the cohort is not active, whereas `getCondition` did not have that check. Perhaps we should remove the isActive active check - thoughts @vzotova ? I think I remember you saying it is better to not use `require` statements in view functions.

packages/shared/src/porter.ts
```diff
@@ -152,6 +152,35 @@ export type TacoDecryptResult = {
   errors: Record<string, string>;
 };
 
+// Signing types
+export type SigningOptions = {
+  optimistic?: boolean;        // Whether to return first signatures received or wait for all signatures
+  returnAggregated?: boolean;  // Whether to return the aggregated signature
```
derekpierre: I think we should remove this option, and always return the aggregated signature. This simplifies things a bit i.e. either:
1. a threshold of responses were received, and therefore we set the aggregatedSignature
2. not enough responses, and so we don't set the aggregatedSignature

That way a failure is basically indicated by no aggregatedSignature being set.

packages/taco/src/sign.ts
vzotova: it's just my preference to not fail view call if you can return something that represents error. But it's not a rule

derekpierre: That's enough of a reason for me - the function can return empty bytes which isn't allowed by nodes when signing anyway. I'll make the update to remove the `require` statement.

derekpierre: ✔️

packages/shared/src/porter.ts
derekpierre: ^ @theref

documentation/README.md
derekpierre, line 1: Some questions about all of these documentation files:
- What's the use of the `CLAUDE.md` file? Does it help with taco-web API usage when using claude?
- Do they help with AI processing of our library?
- Should some of these be placed in the TACo docs, and leave only high level details here.
- How do we maintain/keep these files up to date over time? Are they autogenerated from some script?

theref: oh damn, i did not mean to commit these. I've gone through this whole PR keeping them out of commits! i'll redo and force push

theref: the documentation files are just me getting some first drafts in. claude.md is to help claude understand how to contribute to the codebase. it's been extremely good.

packages/shared/src/types.ts
```diff
@@ -1,3 +1,80 @@
...
+  verificationGasLimit: string;
+  preVerificationGas: string;
+  maxFeePerGas: string;
+  maxPriorityFeePerGas: string;
+  paymaster: string;
+  paymasterVerificationGasLimit: string;
+  paymasterPostOpGasLimit: string;
+  paymasterData: string;
+  signature: string;
+};
```
KPrasch, line 23: Just a note: even though it's not blocking now, This is an object that can be ported to nucypher-core eventually (this entire module really).

packages/shared/src/porter.ts
```diff
@@ -284,4 +354,88 @@ export class PorterClient {
...
+    let signaturesToAggregate: { [ursulaAddress: string]: [string, string] } = {};
+    
+    if (optimistic) {
+      // Optimistic: aggregate all signatures
+      signaturesToAggregate = signingResults;
+    } else {
+      // Non-optimistic: only aggregate signatures from hash that meets threshold
+      for (const [hash, count] of hashCounts.entries()) {
+        if (count >= threshold) {
+          signaturesToAggregate = hashToSignatures.get(hash)!;
```
derekpierre, line 439: Do you need to update the messageHash var here?
```suggestion
          messageHash = hash
          signaturesToAggregate = hashToSignatures.get(hash)!;
```

derekpierre: ✔️

packages/shared/src/porter.ts
```diff
@@ -284,4 +353,89 @@ export class PorterClient {
...
+    });
+
+    const { signatures, errors } = resp.data.result.signing_results;
+    const allErrors: Record<string, string> = { ...errors };
+    const { optimistic = false } = options;
+    
+    const signingResults: { [ursulaAddress: string]: [string, string] } = {};
+    let messageHash = '';
+    
+    // For non-optimistic path
```
derekpierre, line 385: I think there are some tweaks we can make to the logic here, but let's do that in a separate PR.

Thanks for resolving this so quickly!

derekpierre:
A reminder to always run `pnpm fix` locally for linting - [88a539f](https://github.com/nucypher/taco-web/pull/664/commits/88a539f8833b3798251b1566bb5476a40773d5c2)

derekpierre (approved):
🎸

Issue #658, opened 2025-05-21
feat: Grant Decryption Permissions to Specific Wallet Addresses


Muhammad-Altabba:
#### Description
Add a new decryption condition to grant permissions to specific wallet addresses. The condition will work by asking the user to sign a message (authenticate) from one of the addresses in a list. This makes the condition non-transferable. It also has the advantage of being gas-free because it doesn't need a transaction or on-chain data like NFTs or smart contracts.

#### Motivation
This feature will give an easy and gas-free way to manage decryption permissions. It uses wallet authentication with message signing as an alternative to other methods that require transactions or on-chain data.

#### Proposed Implementation
- Add a new condition type, for example, `WalletAllowlistCondition`.
- The condition should take an array of wallet addresses.
- At decryption time, check if the user's wallet is in the list by verifying the signed message.

#### Pros
- **Gas-free:** No transactions are needed for authentication.
- **Lightweight:** Doesn't rely on on-chain data or contracts.

#### Cons
- **Non-transferable:** Permissions can't be transferred to another wallet. This could be a limitation or a feature depending on the use case.

#### Additional Notes
This feature will become even more powerful when combined with the other existing conditions in the system.

derekpierre:
Closed via https://github.com/nucypher/taco-web/pull/665

PR #665, opened 2025-06-03
Add AddressAllowlistCondition  to grant decryption permissions to specific public addresses

Muhammad-Altabba:
**Type of PR:**

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**
Closes https://github.com/nucypher/taco-web/issues/658.

**Issues fixed/closed:**

> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


packages/taco/examples/conditions.ts
```diff
@@ -124,3 +124,25 @@ console.assert(
   !myContractCallCondition.requiresAuthentication(),
   'ContractCondition does not require a signer',
 );
+
+// Wallet Allowlist Condition Example
```
manumonti:
```suggestion
// Address Allowlist Condition Example
```

packages/taco/examples/conditions.ts
```diff
@@ -124,3 +124,25 @@ console.assert(
...
+  addressAllowlistCondition.requiresAuthentication(),
+  'AddressAllowlistCondition requires authentication',
+);
+
+// You can check if an address is allowed with the condition's toObj method
+const addressToCheck = '0x1e988ba4692e52Bc50b375bcC8585b95c48AaD77';
+const addresses = addressAllowlistCondition.toObj().addresses;
+// This would be checked by taco nodes after validating the wallet signature provided by encryptor (the value of the variable `addressToCheck`).
+const isAllowed = addresses.includes(addressToCheck);
+console.log(`Is address ${addressToCheck} allowed? ${isAllowed}`);
```
manumonti:
```suggestion

```

IMHO I would avoid having unnecesary log messages in the tests. I think that those could be a bit noisy.

packages/taco/integration-test/encrypt-decrypt.test.ts
```diff
@@ -25,6 +25,9 @@ const DOMAIN = 'lynx';
 const RITUAL_ID = 27;
 const CHAIN_ID = 80002;
 
+// The wallet address of our consumer (derived from CONSUMER_PRIVATE_KEY)
+const CONSUMER_ADDRESS = '0x5aAeb6053F3E94C9b9A09f33669435E7Ef1BeAed';
```
manumonti: The public key of CONSUMER_PRIVATE_KEY (0xf307e1...) is different than this (correct one: 0xB0B3F5bf904aA4AeB770929BCFfC323439395878) if I'm not wrong.

In any case, I would derive the public key instead of hardcoding it to make easier code maintenance.

```suggestion
// The wallet address of our consumer
const CONSUMER_ADDRESS = ethers.utils.computeAddress(CONSUMER_PRIVATE_KEY);
```

packages/taco/src/conditions/base/address-allowlist.ts
```diff
@@ -0,0 +1,27 @@
...
+      ...value,
+    });
+  }
+
+  /**
+   * This condition requires wallet authentication to verify addresses
+   */
+  override requiresAuthentication(): boolean {
+    return true;
+  }
```
manumonti: Not sure if this is a question or a suggestion, but what do you think about not overriding requiresAuthentication()? Since we are going to require the decryptor's authentication for this condition, we can just add it to the constructor.

This looks to me more appropriate. Or at least, closer to the logic of the Condition class.

Of course, I don't think your approach is wrong and it's totally ok to me if you keep it.

```suggestion
  constructor(value: OmitConditionType<AddressAllowlistConditionProps>) {
    super(addressAllowlistConditionSchema, {
      conditionType: AddressAllowlistConditionType,
      parameters: [USER_ADDRESS_PARAM_DEFAULT]
      ...value,
    });
  }
```

packages/taco/src/conditions/schemas/address-allowlist.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const AddressAllowlistConditionType = 'address-allowlist';
+
+export const addressAllowlistConditionSchema = baseConditionSchema
+  .extend({
+    conditionType: z.literal(AddressAllowlistConditionType),
+    addresses: z
+      .array(EthAddressSchema)
+      .min(1, 'At least one address must be provided')
+      .max(25, 'Maximum of 25 addresses allowed')
```
manumonti: I've been thinking on this limitation to the number of addresses after our chat.

I don't see any inconvenience on having a larger number. As you mentioned, the work to be done by the nodes checking the addresses is very small, so I think there is no limitation from that side.

From the browser client perspective, I don't see any problem with having a large number of addresses. Obviously, the size of the `messageKit` will be larger with more addresses, but this can be so small in relation with the payload that it's not worth considering. 

Unless you see some limitation on this, my preference would be to not have a limittation on the number of addresses.

```suggestion

```

manumonti (changes requested):
This is a great job and a useful condition for TACo 👏.

packages/taco/src/conditions/schemas/address-allowlist.ts
Muhammad-Altabba: Thanks Manu, I think we need to have some limitations to prevent sending some value that would probably bring the nodes down. But as you propose we can allow for much more than 25 addresses. Maybe we can make it 1,000? And if the encryptor needs to share with more than that, then multiple messages can be constructed each one for different set of addresses.

Also, would like to hear @cygnusv and @derekpierre thoughts on this.

packages/taco/src/conditions/base/address-allowlist.ts
Muhammad-Altabba: I would lean toward adding `parameters: [USER_ADDRESS_PARAM_DEFAULT]` if that would be needed inside the condition somewhere else. But I think adding it just for the sake of making requiresAuthentication may cause confusion on what is this param used for exactly.

packages/taco/src/conditions/schemas/address-allowlist.ts
derekpierre: I think having a maximum on the number of addresses  is a good thing, and it shouldn't be too high since that implies a bigger payload for transporting conditions over the wire to nodes. You could increase it if you like (even up to 50), but I wouldn't go larger than that without some actual use case that requires anything larger.

packages/taco/src/conditions/schemas/address-allowlist.ts
```diff
@@ -0,0 +1,38 @@
...
+        try {
+          return addrs.every((addr) => {
+            // Then verify it has the correct checksum (comparing to canonical form)
+            return addr === ethers.utils.getAddress(addr);
+          });
+        } catch (e) {
+          // If any error occurs during validation, fail the validation
+          return false;
+        }
+      }, 'All addresses must be properly checksummed')
```
derekpierre: I think `EthAddressSchema` should already take care of this. If it doesn't, then fix its `refine` function instead of using a new one here.

packages/taco/src/conditions/base/address-allowlist.ts
derekpierre: You will need to explicitly add `:userAddress` somewhere in the condition schema for the node to properly process the condition. We have made the conscious decision that any requirement for use of :userAddress should be explicit and not implicit, since it complicates the logic both on the client and in the node i.e. you shouldn't even need to override `requiresAuthentication` as long as `:userAddress` is in the condition.

For eg. ERC20Balance - https://github.com/nucypher/taco-web/blob/main/packages/taco/src/conditions/predefined/erc20.ts#L14.

That being said, it doesn't have to be:
```typescript
parameters: [USER_ADDRESS_PARAM_DEFAULT]
```
That was only because the rpc call takes an array of parameters. 

Instead you can just define a field as part of the schema, that only allows for the user address context variable to be provided.

```typescript
conditionType: z.literal(AddressAllowlistConditionType),
requesterAddress: UserAddressSchema,
allowedAddresses: z.array(...)
```

derekpierre (commented):
Good start! Left some comments.

Side note: don't merge this PR before the PR in `nucypher` is completed. They need to match each other.

packages/taco/src/conditions/schemas/address-allowlist.ts
Muhammad-Altabba: Thanks @derekpierre, I did not want to mess with the current implementation of `EthAddressSchema` because it will introduce a breaking change (some codes, that was working before this merge, might stop working). So that will require to raise the major version number of our library. i.e. from 0.7 to 1. Are we planning to do so already? Or do I miss something here?

derekpierre: Fair point.

Perhaps we create a new one that is `EthAddressSchemaStrict` or something like that - and its refine function would do what you have above. Then the schema can be reused in other areas instead of just this condition schema.

packages/taco/integration-test/encrypt-decrypt.test.ts
```diff
@@ -128,5 +131,65 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
...
+
+    // TODO: Unskip this test when wallet allowlist condition is implemented and running at the nodes as per https://github.com/nucypher/nucypher/issues/3600
+    test.skip('should encrypt and decrypt according to wallet allowlist condition', async () => {
+      // Create test message
+      const messageString =
+        'This message should only be accessible to allowed wallet addresses';
+      const message = toBytes(messageString);
+
+      // Create wallet allowlist condition with consumer address in the list
+      const addressAllowlistCondition =
```
derekpierre, line 139: Instead of repeating this test, you can just add the allow list condition to the compound and condition on L84

packages/taco/schema-docs/condition-schemas.md
```diff
@@ -6,6 +6,7 @@ _Union of the following possible types:_
 
 - [RpcCondition](#rpccondition)
 - [TimeCondition](#timecondition)
+- [AddressAllowlistCondition](#walletallowlistcondition)
```
derekpierre: Is this link correct? It still has `[wallet]allowlistcondition` instead of `[address]allowlistcondition`.

packages/taco/schema-docs/condition-schemas.md
```diff
@@ -255,9 +263,21 @@ _Object containing the following properties:_
...
 _(\*) Required._
 
+## AddressAllowlistCondition
+
+_Object containing the following properties:_
+
+| Property                 | Description                                                                                                                | Type                                |
+| :----------------------- | :------------------------------------------------------------------------------------------------------------------------- | :---------------------------------- |
+| **`conditionType`** (\*) |                                                                                                                            | `'address-allowlist'`               |
+| **`addresses`** (\*)     | List of wallet addresses allowed to decrypt. Addresses should be provided in checksummed form. Matching is case-sensitive. | `Array<string>` (_min: 1, max: 25_) |
```
derekpierre: This needs to be regenerated right, to include the new `userAddress` field.

packages/taco/src/conditions/schemas/address-allowlist.ts
```diff
@@ -0,0 +1,39 @@
...
+        'List of wallet addresses allowed to decrypt. Addresses should be provided in checksummed form. Matching is case-sensitive.',
+      ),
+  })
+  .strict();
+
+export type AddressAllowlistConditionProps = z.infer<
+  typeof addressAllowlistConditionSchema
+>;
+
+export type AddressAllowlistFields = 'addresses' | 'userAddress';
```
derekpierre: Why is this fields definition needed?

packages/taco/src/conditions/schemas/index.ts
derekpierre, line 1: Why do we need to export the schemas?

packages/taco/src/conditions/schemas/address-allowlist.ts
theref: I feel like a merkle tree would be useful here. Condition/Message Kit would only need to store the merkle root.

Muhammad-Altabba: > I feel like a merkle tree would be useful here. Condition/Message Kit would only need to store the merkle root.

I think this is a great suggestion. But it would need us to discuss more on where to save the full Merkle tree from which the decryptor can take just his relevant proof branches and send it to the nodes. As the full tree would not be sent on the nodes nor it would be part of the cyphertext, I guess. Also its implementation, would  I guess, exceed the initial scope of this feature and make it take longer...

So, for now I will keep the list max size small with a max of 50. And in the soon or later future we would implement using Merkle tree so the number of addresses could be of any size. And we might delay working on it till we have some actual use case that needed that, as Derek hinted.

packages/taco/src/conditions/schemas/address-allowlist.ts
Muhammad-Altabba: Sorry, it is not needed. It was left from an old code during some coding iteration.

packages/taco/src/conditions/schemas/index.ts
Muhammad-Altabba: It is not really needed. I deleted the file. Thanks,

packages/taco/schema-docs/condition-schemas.md
Muhammad-Altabba: The md file needed to be generated. I tend to regenerate when there is a big change or before merging. But it is better to regenerate every time especially to not forge before merging. Thanks,

packages/taco/integration-test/encrypt-decrypt.test.ts
Muhammad-Altabba: Sure, but I wanted to have multiple tests to ease checking if something went wrong. As with every next feature there would be a new test added or modify some old. For this I thing it is good to have a separate one just because the pervious one is already checking for 3 condition types. So, this was my rational but I do not have a strong opinion regarding. So, if you still recommend consolidating in one test, let me know, please.

packages/taco/src/conditions/schemas/address-allowlist.ts
Muhammad-Altabba: I am rethinking about this and I think a bit conservative approach with 25 is enough. As the developer can still combine conditions to reach a higher number if needed.
However, if someone have a strong opinion agains choosing 25. Let us discuss please.

manumonti (approved):

packages/taco/integration-test/encrypt-decrypt.test.ts
```diff
@@ -128,5 +131,66 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
...
+
+      // Prepare for decryption
+      const messageKitFromBytes = ThresholdMessageKit.fromBytes(encryptedBytes);
+      const conditionContext =
+        conditions.context.ConditionContext.fromMessageKit(messageKitFromBytes);
+
+      // Add auth provider for the consumer wallet
+      const authProvider = new EIP4361AuthProvider(provider, consumerSigner, {
+        domain: 'localhost',
+        uri: 'http://localhost:3000',
```
derekpierre, line 172: It's fine to provide these values, but just FYI you don't need to provide them since there are defaults that are used when the values are not provided.

packages/taco/integration-test/encrypt-decrypt.test.ts
derekpierre: It's ok for now, but these tests are somewhat slow, hence the need for increased timeouts, so if we have too many of them the test suite will become pretty slow.

packages/taco/examples/conditions.ts
```diff
@@ -124,3 +124,25 @@ console.assert(
...
+console.assert(
+  addressAllowlistCondition.requiresAuthentication(),
+  'AddressAllowlistCondition requires authentication',
+);
+
+// You can check if an address is allowed with the condition's toObj method
+const addressToCheck = '0x1e988ba4692e52Bc50b375bcC8585b95c48AaD77';
+const addresses = addressAllowlistCondition.toObj().addresses;
+// This would be checked by taco nodes after validating the wallet signature provided by encryptor (the value of the variable `addressToCheck`).
+const isAllowed = addresses.includes(addressToCheck);
```
derekpierre, line 148: Should this be an assertion i.e. `assert(addresses.includes(addressToCheck)`?

packages/taco/src/conditions/schemas/address-allowlist.ts
```diff
@@ -0,0 +1,27 @@
...
+      .array(EthAddressSchemaStrict)
+      .min(1, 'At least one address must be provided')
+      .max(25, 'A maximum of 25 addresses is allowed')
+      .describe(
+        'List of wallet addresses allowed to decrypt. Addresses should be provided in checksummed form. Matching is case-sensitive.',
+      ),
+  })
+  .strict()
+  .describe(
+    'Address Allowlist Condition for allowing decryption for specific wallet addresses. It may not be very handy on its own, but when combined with other conditions, it can be very useful.',
```
derekpierre: Let's keep the docs more positive about the usefulness 😅 

```suggestion
    'Address Allowlist Condition for allowing decryption for specific wallet addresses.',
```

(This update will require a regeneration of the docs)

packages/taco/src/conditions/schemas/common.ts
```diff
@@ -27,7 +27,11 @@ export const UserAddressSchema = z
   );
 
 export const baseConditionSchema = z.object({
-  conditionType: z.string(),
+  conditionType: z
+    .string()
+    .describe(
+      'This value is set automatically at every sub-class constructor when a new object is being created.',
```
derekpierre:
```suggestion
      'A unique identifier that indicates the condition variant in its serialized form.',
```

packages/taco/integration-test/encrypt-decrypt.test.ts
```diff
@@ -128,5 +131,66 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
       // Verify decryption
       expect(decryptedMessageString).toEqual(messageString);
     }, 15000); // 15s timeout
+
+    // TODO: Unskip this test when wallet allowlist condition is implemented and running at the nodes as per https://github.com/nucypher/nucypher/issues/3600
+    test.skip('should encrypt and decrypt according to wallet allowlist condition', async () => {
```
derekpierre: 👍  - I assume the goal is to unskip this test before merging this PR.

Muhammad-Altabba: Yes, and the comment above should be removed when this is ready at the node side.

packages/taco/src/conditions/schemas/address-allowlist.ts
Muhammad-Altabba: Thanks for the pick. It is about pointing to the correct usage scenario with "but when combined with other conditions, it can be very useful.". But as it sounds not positive I removed now the unpossitive part 😄 . Let me know please, if you think more to be changed.

packages/taco/src/conditions/schemas/common.ts
Muhammad-Altabba: I basically wanted to let the user know that this is automatically filled. So, I applied your suggestion along with that addition. Ping me please if you think it still needs editing.

packages/taco/src/conditions/schemas/address-allowlist.ts
```diff
@@ -0,0 +1,27 @@
...
+export const addressAllowlistConditionSchema = baseConditionSchema
+  .extend({
+    conditionType: z.literal(AddressAllowlistConditionType),
+    userAddress: UserAddressSchema,
+    addresses: z
+      .array(EthAddressSchemaStrict)
+      .min(1, 'At least one address must be provided')
+      .max(25, 'A maximum of 25 addresses is allowed')
+      .describe(
+        'List of wallet addresses allowed to decrypt. Addresses should be provided in checksummed form. Matching is case-sensitive.',
```
derekpierre:
```suggestion
        'List of wallet addresses allowed to decrypt. Addresses should be provided in checksummed form.',
```

derekpierre (approved):
🎸

PR #670, opened 2025-06-23
SignatureObject attribute/abi attribute condition

derekpierre:
**Type of PR:**

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**

Related to work done in https://github.com/nucypher/nucypher/pull/3606 and https://github.com/nucypher/nucypher/pull/3609

Adds the `SigningObjectAttributeCondiiton` and `SigningObjectAbiAttributeCondition` from the above `nucypher` PRs,, to the client side library.

The `:signingObjectContextVar` is considered an `AUTOMATICALLY_INJECTED_CONTEXT_PARAMS` so it never needs to be provided by the user/application. The variable will not be included in the list of `requestedContextParmeters` function provided by `ConditionContext`. 

The node will automatically take care of populating that context variable when receiving signing requests since it receives the object to sign anyway. 

**Issues fixed/closed:**

> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


packages/taco/src/conditions/context/context.ts
```diff
@@ -55,6 +58,10 @@ const EXPECTED_AUTH_PROVIDER_TYPES: Record<string, AuthProviderType[]> = {
   ],
 };
 
+export const AUTOMATICALLY_INJECTED_CONTEXT_PARAMS = [
+  // These context parameters are automatically injected on the node side.
+  SIGNING_CONDITION_OBJECT_CONTEXT_VAR,
+];
 export const RESERVED_CONTEXT_PARAMS = [USER_ADDRESS_PARAM_DEFAULT];
```
theref:
```suggestion
export const RESERVED_CONTEXT_PARAMS = [USER_ADDRESS_PARAM_DEFAULT, SIGNING_CONDITION_OBJECT_CONTEXT_VAR];
```

so that users can't use this context variable

derekpierre: ✔️  - it's already prevented via `AUTOMATICALLY_INJECTED_CONTEXT_PARAM`, but your point is well taken that it is also a reserved context param - probably good to add anyway for code maintenance

manumonti (approved):
Amazing TS work! 🙌

packages/taco/schema-docs/condition-schemas.md
```diff
@@ -196,11 +198,11 @@ _(\*) Required._
...
 
-| Property              | Type                                         |
-| :-------------------- | :------------------------------------------- |
-| `index`               | `number` (_int, ≥0_)                         |
-| **`comparator`** (\*) | `'==' \| '>' \| '<' \| '>=' \| '<=' \| '!='` |
-| **`value`** (\*)      | [ParamOrContextParam](#paramorcontextparam)  |
+| Property              | Type                                                          |
+| :-------------------- | :------------------------------------------------------------ |
+| `index`               | `number` (_int, ≥0_)                                          |
+| **`comparator`** (\*) | `'==' \| '>' \| '<' \| '>=' \| '<=' \| '!=' \| 'in' \| '!in'` |
```
theref, line 204: just checking that this follows the same ordering as `>=` etc. current example would be `:userAddress.balance >= 0` and the new comparator would be `:userOp.token in ['ETH', 'USDC']`

i know we talked about `contains` as a comparator as well and just want to make sure that we're getting them the correct way around :)

derekpierre: Correct. The following:
```
...
returnValueTest: {
     comparator: "in"
     value: ["0xABC", "0xDEF"] 
}
```
implies that whatever value is obtained by the condition is checked whether it is one of "0xABC" or "0xDEF".

We talked about substring type of functionality being technically part of a `contains`, but that is not supported, only as a member of a list. This is implemented in PR 3606/3609 in `nucypher`

theref (approved):

PR #676, opened 2025-07-04
Add integration test for condition lingo validation

derekpierre:
**Type of PR:**

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
This integration test ensures that conditions created in `taco-web` properly match the condition structure expected by nodes i.e. that the `zod` schema implemented in typescript matches the `marshmallow` schema implemented in python. 

Until now, we've been able to do that test by attaching conditions to encrypted data, then asking the node to decrypt the data. But when nodes sign data, the conditions are obtained from the SigningCoordinator contract and not from the signing request, so we can't just test signing with different conditions easily.

To test the condition lingo can be sent to the `/validate_condition_lingo` endpoint of any `lynx` node; related to https://github.com/nucypher/nucypher/pull/3615.

The integration test collects the various condition types into one condition for efficiency since the round trip of the test can take a couple seconds. As we add addition condition types we should expand the condition used for the lingo test.

**Issues fixed/closed:**

> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/packages/taco/integration-test/condition-lingo.test.ts
+++ b/packages/taco/integration-test/condition-lingo.test.ts
@@ -48,6 +48,7 @@ async function validateConditionExpression(
   return;
 }
 
+// skip integration test if RUNNING_IN_CI is not set (it is set in CI environments)
 describe.skipIf(!process.env.RUNNING_IN_CI)(
   'TACo Condition Lingos Integration Test',
   () => {
--- a/packages/taco/integration-test/condition-lingo.test.ts
+++ b/packages/taco/integration-test/condition-lingo.test.ts
@@ -1,0 +1,103 @@
+import axios from 'axios';
+import * as https from 'https';
+import { describe, test } from 'vitest';
+import { CompoundConditionType } from '../src/conditions/compound-condition';
+import { ConditionExpression } from '../src/conditions/condition-expr';
+import { IfThenElseConditionType } from '../src/conditions/if-then-else-condition';
+import { SequentialCondition } from '../src/conditions/sequential';
+import {
+  testContractConditionObj,
+  testJsonApiConditionObj,
+  testJsonRpcConditionObj,
+  testJWTConditionObj,
+  testRpcConditionObj,
+  testSigningObjectAbiAttributeConditionObj,
+  testSigningObjectAttributeConditionObj,
+  testTimeConditionObj,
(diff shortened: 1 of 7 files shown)


packages/taco/integration-test/condition-lingo.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  );
+  if (response.status !== 200) {
+    throw new Error(
+      `Request failed with status ${response.status}: ${JSON.stringify(response.data)}`,
+    );
+  }
+  return;
+}
+
+describe.skipIf(!process.env.RUNNING_IN_CI)(
```
manumonti, line 52: Out of curiosity: why not run this test in CI? Is this only intended to be executed locally?

manumonti (approved):

packages/taco/integration-test/condition-lingo.test.ts
derekpierre: It's actually the opposite. 

It runs in CI but not locally i.e. skip if `RUNNING_IN_CI` is not set. In CI it is set - https://github.com/nucypher/taco-web/blob/signing-epic/.github/workflows/main.yml#L35

You can have it run locally by setting the `RUNNING_IN_CI` env variable to any value eg. 
`export RUNNING_IN_CI=true`

manumonti: ooh, right. I understood the opposite indeed

cygnusv: It's kind of a double negative expression ("_skip if not running in CI_"), maybe a comment can help to clarify

cygnusv (approved):

packages/taco/integration-test/condition-lingo.test.ts
derekpierre: ✔️ - added a comment.

A different way to read it is _"skip if RUNNING_IN_CI is not set"_.

There are two defaults that you are working against in this situation:
1. All tests are run by default so you can only set tests to be skipped.
2. Any env variable is considered not set if not explicitly set to value - so the check is for the explicit setting of the variable.

PR #677, opened 2025-07-04
Allow JSON API/RPC conditions to optionally specify `authorizationType`

derekpierre:
**Type of PR:**

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Based over #676 

Allows `authorizationType` to be specified for JSON API/RPC conditions. Allowed values are: `Bearer`, `Basic`, `X-API-Key`.

Based on work done in `nucypher` - https://github.com/nucypher/nucypher/pull/3614

**Issues fixed/closed:**

> - Fixes #...

**Why it's needed:**

> Explain how this PR fits in the greater context of the NuCypher Network. E.g.,
> if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**

> What should reviewers focus on? Is there a particular commit/function/section
> of your PR that requires more attention from reviewers?

PR #678, opened 2025-07-08
Signing SDK Updates

derekpierre:
**Type of PR:**

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

Based over #677 

- Refactor `PorterClient` to focus solely on communicating with the Porter instance i.e. sending input and returning output, without performing additional response processing. It should only handle marshalling and unmarshalling data, leaving all logic beyond that to the caller.
- Shift all post-processing logic from `PorterClient` to the `sign` module to streamline `PorterClient` responsibilities.
- Eliminate the `SigningOption` concept, as signature thresholding and hash verification are always required for `signUserOp` to produce a meaningful result.
- Fix signature aggregation by properly combining signature byte arrays instead of concatenating hex strings, which previously introduced a bug that duplicated `0x` prefixes in the result.
- Clearly differentiate failure cases when an aggregated signature cannot be produced due to either excessive errors or mismatched `messageHash` values across Porter responses. While mismatched hashes are an edge case (potentially due to malicious nodes), they are explicitly handled. In either scenario, the `sign` API will raise an error.

**Issues fixed/closed:**
Follow-up to #664

### DIFF
--- a/packages/taco/integration-test/sign.test.ts
+++ b/packages/taco/integration-test/sign.test.ts
@@ -1,4 +1,10 @@
-import { Domain, getPorterUris, UserOperation } from '@nucypher/shared';
+import {
+  Domain,
+  fromHexString,
+  getPorterUris,
+  SigningCoordinatorAgent,
+  UserOperation,
+} from '@nucypher/shared';
 import { ethers } from 'ethers';
 import { beforeAll, describe, expect, test } from 'vitest';
 
@@ -9,7 +15,7 @@ import { signUserOp } from '../src/sign';
 const RPC_PROVIDER_URL = 'https://ethereum-sepolia-rpc.publicnode.com';
 const DUMMY_ADDRESS = '0x742D35Cc6634C0532925A3b8D33c9c0E7B66C8E8';
 const DOMAIN = 'lynx';
-const RITUAL_ID = 1;
+const COHORT_ID = 1;
 const CHAIN_ID = 11155111;
 
 // skip integration test if RUNNING_IN_CI is not set (it is set in CI environments)
@@ -60,7 +66,7 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
       const signingContext = await context.ConditionContext.forSigningCohort(
         provider,
         DOMAIN as Domain,
-        RITUAL_ID,
+        COHORT_ID,
         CHAIN_ID,
       );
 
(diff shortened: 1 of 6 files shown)


packages/taco/src/sign.ts
```diff
@@ -1,18 +1,54 @@
...
+  messageHash: string;
+  aggregatedSignature: string;
+  signingResults: { [ursulaAddress: string]: TacoSignature };
+  errors: Record<string, string>;
+};
+
+function aggregateSignatures(signaturesByAddress: {
+  [checksumAddress: string]: TacoSignature;
+}): string {
+  // Aggregate hex signatures by concatenating them; being carefule to remove the '0x' prefix from each signature except the first one.
```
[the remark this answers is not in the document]
derekpierre: ✔️

packages/taco/integration-test/sign.test.ts
```diff
@@ -9,7 +15,7 @@ import { signUserOp } from '../src/sign';
 const RPC_PROVIDER_URL = 'https://ethereum-sepolia-rpc.publicnode.com';
 const DUMMY_ADDRESS = '0x742D35Cc6634C0532925A3b8D33c9c0E7B66C8E8';
 const DOMAIN = 'lynx';
-const RITUAL_ID = 1;
+const COHORT_ID = 1;
```
manumonti, line 18: out of the scope of this PR, but this is interesting. I have always thought that `ritual_id` is not a good name for the cohorts. I like cohort_id. Should we use this name from now on in all places?

manumonti (approved):
LGTM! 👌

packages/taco/integration-test/sign.test.ts
derekpierre: Interesting thought and something we can collectively discuss. Ritual ID has been ingrained into the `Coordinator` contract side of things, but doesn't mean we can't update it.
(cc @arjunhassard , @cygnusv , @vzotova ).

manumonti (approved):

derekpierre:
Regarding the removal of `SigningOptions`, I spoke with both @theref and @cygnusv individually. While we’ll likely end up re-adding `SigningOptions`, there’s still a lot to unpack around how the various features should be implemented, for example, supporting a non-optimistic setup that waits for a buffered set of responses beyond the first threshold, incorporating additional validations, etc. 

Given the depth of this topic and the potential logic associated with it, let's tackle it as a separate issue/PR (I'll create one soon) to encourage focused discussion and a more comprehensive implementation plan. This may also have implications for the decryption flow, which currently assumes an entirely optimistic response model.

KPrasch (approved):

PR #679, opened 2025-07-09
ECDSA Condition

KPrasch:
**Type of PR:**
Feature 

Closes https://github.com/nucypher/sprints/issues/183

**Required reviews:**
2

### DIFF
--- a/packages/taco/examples/ecdsa-condition.ts
+++ b/packages/taco/examples/ecdsa-condition.ts
@@ -1,10 +1,19 @@
 import { ECDSACondition } from '../src/conditions/base/ecdsa';
 
+// Example verifying keys (public keys) - in production these would come from actual key pairs
+const EXAMPLE_VERIFYING_KEY_SECP256K1 =
+  '0437a1c9c0e8ff7a2e0e0a9b5a8c3d5b7e9f1c3e5a7b9d1f3e5c7a9b1d3f5e7c9a1e3f5a7b9c1d3e5f7a9b1c3d5e7f9a1b3c5d7e9f1a3b5c7d9e1f3a5';
+const EXAMPLE_VERIFYING_KEY_NIST256P =
+  '0423456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef';
+const EXAMPLE_VERIFYING_KEY_ED25519 =
+  '0fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210fedcba987654321';
+
 // Example 1: ECDSA condition with hardcoded message, dynamic signature
 const attestationCondition = new ECDSACondition({
   message:
     'I attest that I am authorized to access this data and agree to use it responsibly',
   signature: ':userSignature', // User provides signature at runtime
+  verifyingKey: EXAMPLE_VERIFYING_KEY_SECP256K1,
   curve: 'SECP256k1',
 });
 
@@ -12,6 +21,7 @@ const attestationCondition = new ECDSACondition({
 const dynamicECDSACondition = new ECDSACondition({
   message: ':userMessage',
   signature: ':userSignature',
+  verifyingKey: EXAMPLE_VERIFYING_KEY_SECP256K1,
   curve: 'SECP256k1',
 });
 
@@ -19,6 +29,7 @@ const dynamicECDSACondition = new ECDSACondition({
 const p256ECDSACondition = new ECDSACondition({
   message: 'hello world',
   signature: ':signature',
+  verifyingKey: EXAMPLE_VERIFYING_KEY_NIST256P,
   curve: 'NIST256p',
 });
 
@@ -26,6 +37,7 @@ const p256ECDSACondition = new ECDSACondition({
 const defaultECDSACondition = new ECDSACondition({
   message: ':message', // Default context variable
   signature: ':signature', // Default context variable
+  verifyingKey: EXAMPLE_VERIFYING_KEY_SECP256K1,
   curve: 'SECP256k1',
 });
 
@@ -37,11 +49,13 @@ const compoundECDSACondition = new CompoundCondition({
     new ECDSACondition({
       message: ':message1', // Custom context parameter
       signature: ':signature1', // Custom context parameter
+      verifyingKey: EXAMPLE_VERIFYING_KEY_SECP256K1,
       curve: 'SECP256k1',
     }),
     new ECDSACondition({
       message: ':message2', // Different context parameter
       signature: ':signature2', // Different context parameter
+      verifyingKey: EXAMPLE_VERIFYING_KEY_NIST256P,
       curve: 'NIST256p',
     }),
   ],
@@ -78,6 +92,7 @@ const authorizationCondition = CompoundCondition.and([
   new ECDSACondition({
     message: 'I authorize access to this encrypted data',
     signature: ':userSignature',
+    verifyingKey: EXAMPLE_VERIFYING_KEY_SECP256K1,
     curve: 'SECP256k1',
   }),
 ]);
@@ -86,6 +101,7 @@ const authorizationCondition = CompoundCondition.and([
 const ed25519Condition = new ECDSACondition({
   message: 'Edwards curve signature verification',
   signature: ':ed25519Signature',
+  verifyingKey: EXAMPLE_VERIFYING_KEY_ED25519,
   curve: 'Ed25519',
 });
 
@@ -93,6 +109,7 @@ const ed25519Condition = new ECDSACondition({
 const brainpoolCondition = new ECDSACondition({
   message: 'Brainpool P-256 verification',
   signature: ':brainpoolSignature',
+  verifyingKey: EXAMPLE_VERIFYING_KEY_SECP256K1,
   curve: 'BRAINPOOLP256r1',
 });
 
@@ -100,5 +117,6 @@ const brainpoolCondition = new ECDSACondition({
 const p384Condition = new ECDSACondition({
   message: 'High-security P-384 verification',
   signature: ':p384Signature',
+  verifyingKey: EXAMPLE_VERIFYING_KEY_NIST256P,
   curve: 'NIST384p',
 });
--- a/packages/taco/examples/ecdsa-condition.ts
+++ b/packages/taco/examples/ecdsa-condition.ts
@@ -17,16 +17,16 @@ const dynamicECDSACondition = new ECDSACondition({
 
 // Example 3: ECDSA condition with different curve
 const p256ECDSACondition = new ECDSACondition({
-  message: 'Authenticated message',
-  signature: ':ecdsaSignature',
+  message: 'hello world',
+  signature: ':signature',
   curve: 'NIST256p',
 });
 
-// Example 4: Using defaults for message and signature
+// This ECDSA condition uses the default context variables
 const defaultECDSACondition = new ECDSACondition({
-  message: ':ecdsaMessage', // Default context variable
-  signature: ':ecdsaSignature', // Default context variable
-  curve: 'SECP256k1', // Default curve
+  message: ':message', // Default context variable
+  signature: ':signature', // Default context variable
+  curve: 'SECP256k1',
 });
 
 // Example 5: Complex usage with compound condition
--- a/packages/taco/examples/ecdsa-condition.ts
+++ b/packages/taco/examples/ecdsa-condition.ts
@@ -22,11 +22,29 @@ const p256ECDSACondition = new ECDSACondition({
   curve: 'NIST256p',
 });
 
-// This ECDSA condition uses the default context variables
+// This ECDSA condition uses the default context variables (:message and :signature)
 const defaultECDSACondition = new ECDSACondition({
   message: ':message', // Default context variable
   signature: ':signature', // Default context variable
   curve: 'SECP256k1',
+});
+
+// For compound conditions with multiple ECDSA conditions, use custom context parameters
+// to avoid variable name conflicts
+const compoundECDSACondition = new CompoundCondition({
+  operator: 'and',
+  operands: [
+    new ECDSACondition({
+      message: ':message1', // Custom context parameter
+      signature: ':signature1', // Custom context parameter
+      curve: 'SECP256k1',
+    }),
+    new ECDSACondition({
+      message: ':message2', // Different context parameter
+      signature: ':signature2', // Different context parameter
+      curve: 'NIST256p',
+    }),
+  ],
 });
 
 // Example 5: Complex usage with compound condition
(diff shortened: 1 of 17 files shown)


packages/taco/examples/ecdsa-condition.ts
```diff
@@ -0,0 +1,116 @@
+import { ECDSACondition } from '../src/conditions/base/ecdsa';
+
+// Example of creating an ECDSA condition
+console.log('=== ECDSA Condition Example ===');
+
+// Example 1: Basic ECDSA condition with static values
+const basicECDSACondition = new ECDSACondition({
+  message: 'Hello, world!',
+  signature: '304402206e41b6b588af8a85c9a9b2e8b4e2b6e1b3c4d5e6f7a8b9c0d1e2f3a4b5c6d7e8f9a0b1c2d3e4f5a6b7c8d9e0f1a2b3c4d5e6f7a8b9c0d1e2f3a4b5c6d7e8f9',
```
derekpierre: Perhaps I can see the message being hardcoded i.e. the user has to get a specific message signed by an entity, but hardcoding the signature *AND* the message would presumably just always pass (or always fail) - which doesn't really make it useful...?

Is there a situation where this can practically happen? i.e. the signature is hardcoded? Was there a specific case you had in mind here?

packages/taco/src/conditions/schemas/ecdsa.ts
```diff
@@ -0,0 +1,40 @@
...
+export const SUPPORTED_ECDSA_CURVES = [
+  'SECP256k1',
+  'NIST256p',
+  'NIST384p', 
+  'NIST521p',
+] as const;
+
+export type ECDSACurve = typeof SUPPORTED_ECDSA_CURVES[number];
+
+export const DEFAULT_ECDSA_CURVE: ECDSACurve = 'SECP256k1';
```
derekpierre: A few things here:

1. The default curve in Python is `NIST192p` but here it is `SECP256k1`, and `NIST192p` is not in the SUPPORTED curves list in `taco-web`?
```python
In [5]: from nucypher.policy.conditions.ecdsa import DEFAULT_ECDSA_CONDITION_CURVE

In [6]: DEFAULT_ECDSA_CONDITION_CURVE
Out[6]: NIST192p
```

2. Python has a bunch of more curves than what is set here? Should we reduce this list in Python?
```python
In [3]: from nucypher.policy.conditions.ecdsa import SUPPORTED_ECDSA_CONDITION_CURVES

In [4]: SUPPORTED_ECDSA_CONDITION_CURVES
Out[4]: 
{'NIST192p': NIST192p,
 'NIST224p': NIST224p,
 'NIST256p': NIST256p,
 'NIST384p': NIST384p,
 'NIST521p': NIST521p,
 'SECP256k1': SECP256k1,
 'BRAINPOOLP160r1': BRAINPOOLP160r1,
 'BRAINPOOLP192r1': BRAINPOOLP192r1,
 'BRAINPOOLP224r1': BRAINPOOLP224r1,
 'BRAINPOOLP256r1': BRAINPOOLP256r1,
 'BRAINPOOLP320r1': BRAINPOOLP320r1,
 'BRAINPOOLP384r1': BRAINPOOLP384r1,
 'BRAINPOOLP512r1': BRAINPOOLP512r1,
 'SECP112r1': SECP112r1,
 'SECP112r2': SECP112r2,
 'SECP128r1': SECP128r1,
 'SECP160r1': SECP160r1,
 'Ed25519': Ed25519,
 'Ed448': Ed448,
 'BRAINPOOLP160t1': BRAINPOOLP160t1,
 'BRAINPOOLP192t1': BRAINPOOLP192t1,
 'BRAINPOOLP224t1': BRAINPOOLP224t1,
 'BRAINPOOLP256t1': BRAINPOOLP256t1,
 'BRAINPOOLP320t1': BRAINPOOLP320t1,
 'BRAINPOOLP384t1': BRAINPOOLP384t1,
 'BRAINPOOLP512t1': BRAINPOOLP512t1}
```

3. 'Ed25519' is not in the supported list which I believe we need for validating Discord bots?

packages/taco/src/conditions/schemas/ecdsa.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const ecdsaConditionSchema = baseConditionSchema.extend({
+  conditionType: z.literal(ECDSAConditionType).default(ECDSAConditionType),
+  message: z.union([
+    z.string(),
+    contextParamSchema,
+  ]).default(ECDSA_MESSAGE_PARAM_DEFAULT),
```
derekpierre: In https://github.com/nucypher/nucypher/pull/3618 (and prior implementation) you are checking specifically for `:message`? Is the goal for message to be any context variable, or a specific context variable where the default is `:ecdsaMessage`. 

Seems like the logic on the python side may need updating since it seems much more specific.

packages/taco/src/conditions/schemas/ecdsa.ts
```diff
@@ -0,0 +1,40 @@
...
+  signature: z.union([
+    z.string().regex(/^[0-9a-fA-F]+$/, 'Signature must be a valid hex string'),
+    contextParamSchema,
+  ]).default(ECDSA_SIGNATURE_PARAM_DEFAULT),
+  verifyingKey: z.union([
+    z.string().regex(/^[0-9a-fA-F]+$/, 'Verifying key must be a valid hex string'),
```
derekpierre: You can probably define a hexstring schema that does a hex string regex that can apply to both signature and verifying key, so you don't have to copy the regex.

eg.

```typescript
const hexStringSchema = z.string().regex(/^[0-9a-fA-F]+$/, 'Invalid hex string');
```

packages/taco/src/conditions/schemas/ecdsa.ts
```diff
@@ -0,0 +1,40 @@
...
+    contextParamSchema,
+  ]),
+  curve: z.enum(SUPPORTED_ECDSA_CURVES).default(DEFAULT_ECDSA_CURVE),
```
derekpierre: In python the `curve` is optional, but here it is required and defaults to the default curve - perhaps we just make the `curve` value required in python?

packages/taco/src/conditions/schemas/ecdsa.ts
```diff
@@ -0,0 +1,40 @@
...
+export const DEFAULT_ECDSA_CURVE: ECDSACurve = 'SECP256k1';
+
+export const ecdsaConditionSchema = baseConditionSchema.extend({
+  conditionType: z.literal(ECDSAConditionType).default(ECDSAConditionType),
+  message: z.union([
+    z.string(),
+    contextParamSchema,
+  ]).default(ECDSA_MESSAGE_PARAM_DEFAULT),
+  signature: z.union([
+    z.string().regex(/^[0-9a-fA-F]+$/, 'Signature must be a valid hex string'),
```
derekpierre: See my earlier comment/question about whether there is legit  case for a signature to be hardcoded...

packages/taco/integration-test/encrypt-decrypt.test.ts
```diff
@@ -189,5 +190,227 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
...
+      await expect(
+        decrypt(
+          provider,
+          DOMAIN,
+          messageKitFromBytes,
+          conditionContext,
+        )
+      ).rejects.toThrow();
+
+      console.log('✅ ECDSA invalid signature test passed - decryption properly failed');
```
derekpierre: nitpick: Let's skip on the console logs since they can be a lot. This is just the test passing. Applies to other locations in the file.

packages/taco/examples/ecdsa-condition.ts
```diff
@@ -0,0 +1,116 @@
...
+});
+
+console.log('\nECDSA Condition with P-256 curve:');
+console.log(JSON.stringify(p256ECDSACondition.toObj(), null, 2));
+
+// Example 4: Using defaults for message and signature
+const defaultECDSACondition = new ECDSACondition({
+  message: ':ecdsaMessage', // Default context variable
+  signature: ':ecdsaSignature', // Default context variable
+  verifyingKey: '04a34b99f22c790c4e36b2b3c2c35a36db06226e41c692fc82b8b56ac1c540c5bd5b8dec5235a0fa8722476c7709c02559e3aa73aa03918ba2d492eea75abea235',
```
derekpierre: If using the defaults, you should be able to do the following?

```suggestion
const defaultECDSACondition = new ECDSACondition({
  verifyingKey: '04a34b99f22c790c4e36b2b3c2c35a36db06226e41c692fc82b8b56ac1c540c5bd5b8dec5235a0fa8722476c7709c02559e3aa73aa03918ba2d492eea75abea235',
```

or were you trying to be explicit about the defaults?

packages/taco/src/conditions/schemas/ecdsa.ts
```diff
@@ -0,0 +1,40 @@
...
+    contextParamSchema,
+  ]).default(ECDSA_SIGNATURE_PARAM_DEFAULT),
+  verifyingKey: z.union([
+    z.string().regex(/^[0-9a-fA-F]+$/, 'Verifying key must be a valid hex string'),
+    contextParamSchema,
```
derekpierre: If the verifying key can be a context variable, then the requester could just specify any verifying key they want, no? In which case at request time, the condition can always be made to pass...

i.e. create an ephemeral key pair, sign some data, and then set the verifying key as the public key of the ephemeral key pair.

packages/taco/src/conditions/schemas/ecdsa.ts
KPrasch: ✔️

packages/taco/src/conditions/schemas/ecdsa.ts
KPrasch: ✔️

KPrasch: ✔️

packages/taco/src/conditions/schemas/ecdsa.ts
KPrasch: Fair point, early on I was uncertain if there was a use case here.  However, since one has not emerged, I'll remove it.

packages/taco/src/conditions/schemas/ecdsa.ts
```diff
@@ -0,0 +1,63 @@
...
+import { baseConditionSchema, hexStringSchema } from './common';
+import { contextParamSchema } from './context';
+
+export const ECDSA_MESSAGE_PARAM_DEFAULT = ':ecdsaMessage';
+export const ECDSA_SIGNATURE_PARAM_DEFAULT = ':ecdsaSignature';
+
+export const ECDSAConditionType = 'ecdsa';
+
+// Supported ECDSA curves (must match Python backend format)
+export const SUPPORTED_ECDSA_CURVES = [
```
derekpierre, line 11: @cygnusv - penny for your thoughts on whether we should provide support for some of these curves? eg. `BRAINPOOL...` ones.

KPrasch: I'm indifferent here. I just supplied the full range of supported curves by the underlying verifier library on the node-side.

packages/taco/src/conditions/schemas/ecdsa.ts
KPrasch: ✔️  https://github.com/nucypher/nucypher/pull/3622

packages/taco/src/conditions/schemas/ecdsa.ts
KPrasch: ✔️ only `:message` and `:signature` are supported now.

KPrasch: scratch that, it's fully flexible, with defaults now :-)

packages/taco/src/conditions/schemas/ecdsa.ts
KPrasch: removed most of them, reduced to the most commonly used.

derekpierre (approved):
🎸

packages/taco/test/test-utils.ts
```diff
@@ -91,6 +100,27 @@ import {
 import { DkgClient, DkgRitual } from '../src/dkg';
 import { encryptMessage } from '../src/tdec';
 
+export function createTestECDSACondition(
+  message: string,
+  curve: ECDSACurve = 'SECP256k1',
+): { conditionProps: ECDSAConditionProps; privateKey: string } {
+  // Simulate server-side key generation for the predefined condition
+  const testWallet = ethers.Wallet.createRandom();
```
manumonti: Something that got me confused here is that you are creating a ECDSA verifying/public key from a ethereum wallet public key. My first thought was that these two are not the same. But after some research I found that they are indeed the same because both are using secp256k1 curve. So, in my opinion, it is worth to have a note on this. Something like

```
// ethers.Wallet uses the same curve that we are using for ECDSA: secp256k1, so we are generating the ECDSA verifying key from ethereum wallet public key.
```

Does this make sense?

packages/taco/examples/ecdsa-condition.ts
```diff
@@ -0,0 +1,122 @@
...
+const attestationCondition = new ECDSACondition({
+  message:
+    'I attest that I am authorized to access this data and agree to use it responsibly',
+  signature: ':userSignature', // User provides signature at runtime
+  verifyingKey: EXAMPLE_VERIFYING_KEY_SECP256K1,
+  curve: 'SECP256k1',
+});
+
+// Example 2: ECDSA condition with context variables
+const dynamicECDSACondition = new ECDSACondition({
```
manumonti, line 23: We could add this at the beginning of the file to avoid the linter complaining about unused variables:

```
/* eslint-disable @typescript-eslint/no-unused-vars */
```

packages/taco/integration-test/encrypt-decrypt.test.ts
```diff
@@ -25,10 +31,8 @@ const DOMAIN = 'lynx';
 const RITUAL_ID = 27;
 const CHAIN_ID = 80002;
 
-// The wallet address of our consumer
```
manumonti, line 28: Out of curiosity: what is the reason for removing these comments? can be useful in my opinion

manumonti (dismissed):
It looks good to me! I left some small suggestions in case you want to take them.

packages/taco/examples/ecdsa-condition.ts
derekpierre: ✔️

packages/taco/test/test-utils.ts
derekpierre: ✔️ Great point! Also led to further clean up that I missed.

manumonti (approved):

packages/taco/integration-test/encrypt-decrypt.test.ts
derekpierre: ^ @KPrasch

PR #680, opened 2025-07-11
Contract conditions

theref:


KPrasch:
Does this PR need a rebase - It appears to have commits from other PRs?

PR #681, opened 2025-07-11
Extend sdk to include setting conditions on Signing Coordinator

theref:
## Overview
This PR extends the TACo SDK to include functionality for setting conditions on the Signing Coordinator contract. It adds a new `setSigningCohortConditions` function that allows users to programmatically configure signing cohort conditions directly through the SDK, along with comprehensive unit and integration tests.

## Why
This provides a user-facing API for condition setting that was previously only available through direct contract interaction, making it easier for developers to manage signing cohort configurations.

## Implementation
Added the `setSigningCohortConditions` function to the sign module with proper parameter ordering (provider, domain, conditions, cohortId, chainId, signer), comprehensive unit tests covering success and error cases, and integration tests that validate condition serialization round-trips.


KPrasch (commented):
🚀 Finally a user-facing API for condition setting :-) looking forward to this one.

packages/shared/src/contracts/agents/signing-coordinator.ts
```diff
@@ -57,6 +57,23 @@ export class SigningCoordinatorAgent {
     return cohortCondition;
   }
 
+  public static async setSigningCohortConditions(
+    provider: ethers.providers.Provider,
+    domain: Domain,
+    cohortId: number,
+    chainId: number,
+    conditions: string,
```
derekpierre: Let's have the type for `conditions` be `Uint8Array` so that we keep the agent and the contract more closely aligned. The caller can take care of the conversion to bytes.

packages/taco/src/sign.ts
```diff
@@ -173,3 +174,25 @@ export async function signUserOp(
...
   };
 }
+
+export async function setSigningCohortConditions(
+  conditions: ConditionExpression,
+  provider: ethers.providers.JsonRpcProvider,
+  domain: Domain,
+  cohortId: number,
+  chainId: number,
+  signer: ethers.Signer,
```
derekpierre: A couple things:
1. To remain consistent with `encrypt`, let's take a `Condition` here, internally we can convert it to a `ConditionExpression`.
2. The existing taco APIs (`encrypt`, `decrypt`, `signUserOp`) have an implicit order of parameters that we should try to remain consistent with. `provider` first, `domain` second, ..., `signer` last.

```suggestion
  provider: ethers.providers.JsonRpcProvider,
  domain: Domain,
  conditions: Condition,
  cohortId: number,
  chainId: number,
  signer: ethers.Signer,
```

wdyt?

packages/taco/integration-test/sign.test.ts
```diff
@@ -98,6 +100,76 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
       );
       expect(signResult.signingResults).toBeDefined();
       expect(Object.keys(signResult.signingResults).length).toBeGreaterThan(0);
-    }, 15000);
+    }, 150000);
```
derekpierre: This was fixed before, perhaps undone as part of your rebase?

In any case, it should be 15s and not 150s.

```suggestion
    }, 15000);
```

packages/taco/integration-test/sign.test.ts
```diff
@@ -98,6 +100,76 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
...
+        );
+      } catch (error: unknown) {
+        const errorMessage = error instanceof Error ? error.message : String(error);
+        if (errorMessage.includes('cohort does not exist') || 
+            errorMessage.includes('invalid cohort') ||
+            errorMessage.includes('conditions not set')) {
+          console.log(`Skipping test - no conditions found for cohort ${COHORT_ID}. This test requires an existing cohort with conditions set.`);
+          return;
+        }
+        throw error;
```
derekpierre: I don't believe this is necessary. The contract returns empty bytes for a non-existent cohort ID or non-set conditions for a chain for a specific cohort Id. So you probably just need to check if the bytes are empty.

Also, just noticed that the definition of `getSigningCohortConditions` for the SigningCoordinatorAgent returns `Promise<string>`. Is that because it returns a hex string instead of bytes?

packages/taco/integration-test/sign.test.ts
```diff
@@ -98,6 +100,76 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
...
+      }
+      
+      // If the hex is empty or just '0x', skip the test
+      if (!retrievedConditionHex || retrievedConditionHex === '0x' || retrievedConditionHex.length <= 2) {
+        console.log(`Skipping test - cohort ${COHORT_ID} has no conditions set.`);
+        return;
+      }
+      
+      console.log('Retrieved condition hex from chain:', retrievedConditionHex);
```
derekpierre: Let's skip on the printing - the terminal gets flooded with these (I run integration tests locally 😄 ). Applies to other locations below.

packages/taco/integration-test/sign.test.ts
```diff
@@ -98,6 +100,76 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
...
+          return;
+        }
+        throw error;
+      }
+      
+      // If the hex is empty or just '0x', skip the test
+      if (!retrievedConditionHex || retrievedConditionHex === '0x' || retrievedConditionHex.length <= 2) {
+        console.log(`Skipping test - cohort ${COHORT_ID} has no conditions set.`);
+        return;
+      }
```
derekpierre: I think do an `expect` instead to check that the conditions are set - if not set, then the test will fail and we'll update the values used. It's a way of explicitly making sure that the test is being run and if it fails because the cohortID/chainID combination no longer has a condition we should update the values or extend the ritual etc.

packages/taco/integration-test/sign.test.ts
```diff
@@ -98,6 +100,76 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
...
+      
+      // Additional verification: ensure the objects have expected properties
+      expect(retrievedConditions.condition).toBeDefined();
+      
+      // Verify that the semantic content is identical (JSON property order may differ)
+      expect(JSON.parse(roundTripJson)).toEqual(JSON.parse(chainConditionJson));
+      
+      console.log('Condition serialization round-trip validation completed successfully!');
+      
+    }, 60000); // Shorter timeout since no blockchain writes
```
derekpierre: I think you meant 6s instead of 60s.
```suggestion
    }, 6000); // Shorter timeout since no blockchain writes
```

packages/taco/integration-test/sign.test.ts
```diff
@@ -98,6 +100,76 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
...
+      const coreConditions = new CoreConditions(chainConditionJson);
+      const retrievedConditions = ConditionExpression.fromCoreConditions(coreConditions);
+      
+      // Also verify we can create a ConditionContext from it (this validates the data is readable)
+      const retrievedContext = await context.ConditionContext.forSigningCohort(
+        provider,
+        DOMAIN as Domain,
+        COHORT_ID,
+        CHAIN_ID,
+      );
```
derekpierre: Maybe just do an `expect(...).not.toThrow()` since `retrievedContext` isn't used after this line?

packages/taco/integration-test/sign.test.ts
```diff
@@ -98,6 +100,76 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
...
+      
+      // Convert back to JSON (this is what would be written to chain)
+      const roundTripJson = retrievedConditions.toJson();
+      console.log('Round-trip JSON:', roundTripJson);
+      
+      // The core verification: ensure serialization round-trip is identical
+      // This proves that read → object → write produces identical bytes
+      expect(JSON.parse(roundTripJson)).toEqual(JSON.parse(chainConditionJson));
```
derekpierre, line 145: This check is repeated on L169?

derekpierre (commented):
Seems like netlify can't launch demos - let's look into why that is as well.

packages/taco/src/sign.ts
```diff
@@ -173,3 +173,22 @@ export async function signUserOp(
     signingResults: porterSignResult.signingResults,
   };
 }
+
+export async function setSigningCohortConditions(
+  provider: ethers.providers.JsonRpcProvider,
+  domain: Domain,
+  conditions: Uint8Array,
```
derekpierre: Apologies, I likely didn't explain the overall picture well.

This external facing API should use `Condition` similar to `encrypt`/`encryptWithPublicKey`. Subsequently (within this function) we can create the `ConditionExpression` and then convert it to bytes before calling `setSigningCohortConditions` on the agent.

.gitignore
```diff
@@ -16,3 +16,4 @@ docs-json
 .next
 packages/docs
 docs
+.claude/
```
derekpierre, line 19: Hehe interesting.

theref: yeah I'm experimenting a lot at the moment. I'm pen to hear what other people are doing and whether we want to standardise some stuff

packages/taco/src/sign.ts
theref: ahhhh ok, I was quite surprised! I should've pushed back a bit, sorry

derekpierre: Definitely push back if you ever disagree and/or think that I sound off my rocker 😅 

The underlying philosophy, at least for me, boils down to:
- Agent function parameters should pretty much match the contract call parameters - not much, if any, additional processing should be done in agents.
- SDK/API parameters should be simple / consistent for the external caller of the API - not too much processing for them to do before calling
- Our implementation of the API that connects the two (API and agents), can do the additional processing

derekpierre:
Seems like the netlify demo previews are failing (https://github.com/nucypher/taco-web/pull/681#issuecomment-3061366084, https://github.com/nucypher/taco-web/pull/681#issuecomment-3061369788). From the looks of the log, this seems to be the issue:

```
5:05:53 AM: Failed during stage 'Install dependencies': dependency_installation script returned non-zero exit code: 1
5:05:53 AM:  ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/package.json
5:05:53 AM: Note that in CI environments this setting is true by default. If you still need to run install in such cases, use "pnpm install --no-frozen-lockfile"
5:05:53 AM:   Failure reason:
5:05:53 AM:   specifiers in the lockfile don't match specifiers in package.json:
* 2 dependencies are mismatched:
5:05:53 AM:   - @vitest/coverage-v8 (lockfile: ^3.0.9, manifest: ^3.2.4)
5:05:53 AM:   - eslint-plugin-unused-imports (lockfile: ^3.2.0, manifest: ^4.1.4)
```

It was probably some mismatch that perhaps happened during a rebase. I think you can run `pnpm install --fix-lockfile` from the root folder to resolve the problem.

packages/taco/src/sign.ts
```diff
@@ -173,3 +175,27 @@ export async function signUserOp(
...
+  domain: Domain,
+  conditions: Condition,
+  cohortId: number,
+  chainId: number,
+  signer: ethers.Signer,
+): Promise<ethers.ContractTransaction> {
+  // Convert Condition to ConditionExpression, then to JSON, then to bytes
+  const conditionExpression = new ConditionExpression(conditions);
+  const conditionsJson = conditionExpression.toJson();
+  const conditionsBytes = ethers.utils.toUtf8Bytes(conditionsJson);
```
derekpierre, line 190: 🔥

derekpierre (commented):
Once we get the netlify previews working, I'll approve.

derekpierre (approved):
🎸 .

The netlify issue seems separate, and hasn't run on a new commit.

KPrasch (approved):

PR #682, opened 2025-07-25
Fix SequentialCondition edge case

derekpierre:
**Type of PR:**

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Related to work done on the `nucypher` side - https://github.com/nucypher/nucypher/pull/3620

**Issues fixed/closed:**

> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/packages/taco/src/conditions/schemas/sequential.ts
+++ b/packages/taco/src/conditions/schemas/sequential.ts
@@ -1,9 +1,55 @@
 import { z } from 'zod';
 
+import { ConditionProps } from '../condition';
 import { maxNestedDepth } from '../multi-condition';
 
 import { baseConditionSchema, plainStringSchema } from './common';
+import { CompoundConditionType } from './compound';
+import { IfThenElseConditionType } from './if-then-else';
 import { anyConditionSchema } from './utils';
+
+const getAllNestedConditionVariableNames = (
+  condition: ConditionProps,
+): string[] => {
+  const conditionVariables: string[] = [];
+  if (condition.conditionType === SequentialConditionType) {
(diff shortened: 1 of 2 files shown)


packages/taco/src/conditions/schemas/sequential.ts
```diff
@@ -1,10 +1,56 @@
...
+const noDuplicateVarNames = (condition: ConditionProps): boolean => {
+  const allVarNames = getAllNestedConditionVariableNames(condition);
+  const seen = new Set();
+  for (const varName of allVarNames) {
+    if (seen.has(varName)) {
+      return false; // Duplicate variable name found
+    }
+    seen.add(varName);
+  }
+  return true; // No duplicates found
```
manumonti: I don't see any problem with this, but if you would like to do it in a more javascript-style way, you could try something like

```suggestion
  const duplicates = allVarNames.filter((item, index) => allVarNames.indexOf(item) !== index);
  return duplicates.length === 0;
```

manumonti (approved):
LGTM! 👌

I left a little alternative for the `noDuplicateVarNames` function, but I don't think it's especially better TBH, so feel free to take it or leave it.

packages/taco/src/conditions/schemas/sequential.ts
derekpierre: ✔️ I'll take it - I think your suggestion is better

manumonti (approved):

KPrasch (approved):

PR #688, opened 2025-08-06
Add ContextVariableCondition with flexible variable matching

theref:
## Summary
- ✨ **New Feature**: ContextVariableCondition class supporting flexible context variable matching
- 🔧 **Enhanced Flexibility**: Supports all comparison operators (==, >, <, >=, <=, \!=, in, \!in) 
- 📦 **Deprecation Path**: AddressAllowlistCondition marked deprecated with migration guidance

## Key Changes
| File | Change |
|------|--------|
| `conditions/base/context-variable.ts` | **NEW** - Core ContextVariableCondition class |
| `conditions/schemas/context-variable.ts` | **NEW** - Zod schema with validation |
| `conditions/condition-factory.ts` | Integration into factory pattern |
| `conditions/base/address-allowlist.ts` | Added @deprecated annotation |
| `test/conditions/base/context-variable.test.ts` | **NEW** - 194 lines comprehensive test suite |

## Migration Example
**Before (AddressAllowlistCondition):**
```typescript
new AddressAllowlistCondition({
  userAddress: ':userAddress',
  addresses: ['0x123...', '0x456...']
})
```

**After (ContextVariableCondition):**
```typescript
new ContextVariableCondition({
  contextVariable: ':userAddress',
  returnValueTest: {
    comparator: 'in',
    value: ['0x123...', '0x456...']
  }
})
```

## Test Coverage
- ✅ Schema validation (valid/invalid inputs)
- ✅ Constructor behavior and error handling  
- ✅ Authentication requirement detection
- ✅ Equality comparisons
- ✅ All comparison operators tested

## Review Focus
1. **Schema validation** in `context-variable.ts` - ensures robust input handling
2. **Factory integration** in `condition-factory.ts` - clean registration pattern
3. **Test coverage** - comprehensive edge case handling
4. **Deprecation messaging** - clear migration path for existing users


packages/taco/src/conditions/base/address-allowlist.ts
```diff
@@ -9,6 +9,10 @@ export { AddressAllowlistConditionProps, AddressAllowlistConditionType };
 
 /**
  * A condition that checks if a wallet address is in a list of allowed addresses.
+ * 
+ * @deprecated Use ContextVariableCondition instead, which provides more flexible
+ * context variable matching capabilities and can replicate this functionality
+ * with contextVariable: ':userAddress' and expectedValues as the addresses array.
  */
 export class AddressAllowlistCondition extends Condition {
```
derekpierre: Since `AddressAllowlistCondition` was never publicly released, it can be removed instead of deprecated.

packages/taco/src/conditions/base/context-variable.ts
```diff
@@ -0,0 +1,22 @@
...
+  contextVariableConditionSchema,
+  ContextVariableConditionType,
+} from '../schemas/context-variable';
+import { OmitConditionType } from '../shared';
+
+export { ContextVariableConditionProps, ContextVariableConditionType };
+
+/**
+ * A condition that checks if a context variable matches one of the expected values.
+ * Supports case-insensitive comparisons.
```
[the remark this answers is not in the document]
derekpierre: ^ Agreed. It's only case insensitive in the case of hex strings like wallet addresses because `nucypher` converts the hex to bytes before comparing, but that does not apply to regular strings.

packages/taco/src/conditions/base/context-variable.ts
```diff
@@ -0,0 +1,22 @@
...
+  ContextVariableConditionProps,
+  contextVariableConditionSchema,
+  ContextVariableConditionType,
+} from '../schemas/context-variable';
+import { OmitConditionType } from '../shared';
+
+export { ContextVariableConditionProps, ContextVariableConditionType };
+
+/**
+ * A condition that checks if a context variable matches one of the expected values.
```
derekpierre: It's less of a "match" but more of a "comparison".
For example, you could have something like the following:
```json
{
    "conditionType": "context-variable",
    "contextVariable": ":myContextVar",
    "returnValueTest": {
        "comparator": ">",
        "value": 0
    }
}
```

packages/taco/src/conditions/schemas/address-allowlist.ts
```diff
@@ -19,7 +19,7 @@ export const addressAllowlistConditionSchema = baseConditionSchema
   })
   .strict()
   .describe(
-    'Address Allowlist Condition for allowing decryption for specific wallet addresses. It is very handy when combined with other conditions.',
+    'Address Allowlist Condition for allowing decryption for specific wallet addresses. It is very handy when combined with other conditions. @deprecated Use ContextVariableCondition instead.',
```
derekpierre: See comment about totally removing instead of deprecating.

packages/taco/src/conditions/schemas/context-variable.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const ContextVariableConditionType = 'variable-match';
+
+export const contextVariableConditionSchema = baseConditionSchema
+  .extend({
+    conditionType: z.literal(ContextVariableConditionType),
+    contextVariable: contextParamSchema.describe(
+      'The context variable to check (e.g., ":userAddress", ":customParam")',
+    ),
+    returnValueTest: returnValueTestSchema.describe(
```
derekpierre: Perhaps move the text description to the `returnValueTestSchema` itself. That way it will be reused in the documentation by all users of the `returnValueTestSchema`.

packages/taco/src/conditions/schemas/context-variable.ts
```diff
@@ -0,0 +1,26 @@
+import { z } from 'zod';
+
+import { baseConditionSchema } from './common';
+import { contextParamSchema } from './context';
+import { returnValueTestSchema } from './return-value-test';
+
+export const ContextVariableConditionType = 'variable-match';
```
derekpierre: This doesn't match what's in `nucypher`; should be `context-variable`, see https://github.com/nucypher/nucypher/pull/3629/files#diff-39ed7645b1d8386d5199d3de0d4bca1d523549567147e8c8c3f4f685b1f14da3R114.

Unless you are proposing a switch to that name? I wouldn't include "match" since it is really an comparison check i.e. not necessarily equals/not equals, but also less-than, greater-than etc.

packages/taco/test/conditions/base/context-variable.test.ts
```diff
@@ -0,0 +1,194 @@
...
+    });
+
+    it('accepts valid context parameters', () => {
+      const validContextParams = [
+        ':userAddress',
+        ':customParam',
+        ':myVariable',
+      ];
+
+      for (const param of validContextParams) {
```
derekpierre: You can use `it.each` for this


```
    it.each([
      ':userAddress', 
      ':customParam', 
      ':myVariable'
    ])('accepts valid context parameters', (param) => {
```

packages/taco/test/conditions/base/context-variable.test.ts
```diff
@@ -0,0 +1,194 @@
...
+        ...testContextVariableConditionObj,
+        returnValueTest: {
+          comparator: '==',
+          value: 'different-value',
+        },
+      });
+      expect(condition1.equals(condition2)).toBe(false);
+    });
+  });
+});
```
derekpierre: Nice tests - 🔥

packages/taco/test/test-utils.ts
```diff
@@ -370,16 +374,27 @@ export const testRpcConditionObj: RpcConditionProps = {
...
+    value: [
       '0x1e988ba4692e52Bc50b375bcC8585b95c48AaD77',
       '0x0000000000000000000000000000000000000001',
-      '0x0000000000000000000000000000000000000002',
     ],
-  };
+  },
+};
+
+export const testAddressAllowlistConditionObj: AddressAllowlistConditionProps = {
```
derekpierre: Let's remove this as part of the `AddressAllowlistCondition` removal.

derekpierre (commented):
Nice work - left some initial comments.

- Can we also add an integration test that validates the condition structure. See https://github.com/nucypher/taco-web/blob/signing-epic/packages/taco/integration-test/condition-lingo.test.ts#L63? Note the test will fail until we update lynx with the latest code from https://github.com/nucypher/nucypher/pull/3629
- After removing addressallowlistcondition, please also run `pnpm generate-zod-docs` from the `packages/taco` directory to remove the `AddressAllowlistCondition` from the condition-schemas.md file, and to add the new `ContextVariableCondition` to the same file.

packages/taco/src/conditions/schemas/context-variable.ts
```diff
@@ -0,0 +1,24 @@
...
+  .extend({
+    conditionType: z.literal(ContextVariableConditionType),
+    contextVariable: contextParamSchema.describe(
+      'The context variable to check (e.g., ":userAddress", ":customParam")',
+    ),
+    returnValueTest: returnValueTestSchema,
+  })
+  .strict()
+  .describe(
+    'Context Variable Condition for performing comparison operations on context variable values. Supports various comparison operators.',
```
derekpierre: Minor nitpick - you can accept or reject.

```suggestion
    'Context Variable Condition for performing comparison operations on context variable values.',
```

If accepted you'll need to rerun `pnpm generate-zod-docs`.

packages/taco/src/conditions/base/context-variable.ts
```diff
@@ -0,0 +1,22 @@
...
+  ContextVariableConditionType,
+} from '../schemas/context-variable';
+import { OmitConditionType } from '../shared';
+
+export { ContextVariableConditionProps, ContextVariableConditionType };
+
+/**
+ * A condition that performs comparison operations on context variable values.
+ * Supports various comparison operators like ==, >, <, >=, <=, !=, in, !in.
+ */
```
derekpierre: Minor nitpick - I wouldn't list the operators. You can accept or reject.

```suggestion
/**
 * A condition that performs comparison operations on context variable values.
 */
```

derekpierre (approved):
🎸 Nice work!

A reminder to run `pnpm fix` before making commits. There are some outstanding linting issues that need to be fixed via that command before merging this PR.

packages/taco/src/conditions/base/address-allowlist.ts
Muhammad-Altabba, line 1: I think it is beneficial for keeping an easy to use condition with a simple name especially for AllowList. For this (as suggested by @derekpierre at some private meeting), we may have a class similar to ERC20Balance and ERC721Balance that wraps around the more abstract condition.

packages/taco/src/conditions/schemas/return-value-test.ts
Muhammad-Altabba, line 1: It is out of the scope of this PR. 
But I think we need to put some limit on the number of items in the array. Maybe we can do it in another PR? Similarly need also to do a similar check at nucypher nodes:
> [if comparator in ["in", "!in"] and not isinstance(value, list):](https://github.com/nucypher/nucypher/blob/722b6694200c29d9fe75ba2b0f51b210c558cba8/nucypher/policy/conditions/lingo.py#L624-L625)

I this right? Should I create an issue for this or am I missing something?

packages/taco/integration-test/encrypt-decrypt.test.ts
```diff
@@ -123,58 +123,6 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
       expect(decryptedMessageString).toEqual(messageString);
     }, 15000);
 
-    test('should encrypt and decrypt according to wallet allowlist condition', async () => {
```
Muhammad-Altabba, line 126: Running in integration test is very helpful. If not running it always, probably you may add one with `skip` that you run at least locally to confirm that everything is running as expected...

Muhammad-Altabba (commented):
Looks great to me, beside the mentioned points to discuss...

packages/taco/src/conditions/base/address-allowlist.ts
derekpierre: Right. Let's do this in a separate PR. That address allowlist wrapper can probably perform the length checks of the list that you mentioned. Filed - https://github.com/nucypher/sprints/issues/206.

packages/taco/src/conditions/schemas/return-value-test.ts
derekpierre: > I this right? Should I create an issue for this or am I missing something?

It is correct. The only current supported functionality of `in` & `!in` is for lists.

packages/taco/integration-test/encrypt-decrypt.test.ts
derekpierre: Not sure I understand what you mean. Could you clarify? This test was removed because it was related to allowlist condition specifically, which was fully removed in this PR. It may be added back in https://github.com/nucypher/sprints/issues/206.

Muhammad-Altabba: I mean to add (or modify) an integration test that tests ContextVariableCondition.

derekpierre: 👍  - @theref, as @Muhammad-Altabba pointed out, could you add an integration test for `encrypt-decrypt` with the `ContextVariableCondition`.

derekpierre (approved):
🎸

Muhammad-Altabba (approved):
Great job!

PR #693, opened 2025-08-19
Add TACo Account Abstraction signing demo

theref:
## Summary
- Add comprehensive TACo Account Abstraction signing demo for adopters
- Single-file implementation showing TACo threshold signatures with ERC-4337
- Complete documentation with example output and setup instructions

## Demo Features
- Real TACo testnet integration (2-of-3 threshold)
- MetaMask Delegation Toolkit smart account creation
- Balance tracking and fund management
- Pimlico bundler integration for UserOperation execution
- Clean developer experience with minimal boilerplate

## Test Plan
- [x] Demo builds successfully
- [x] All tests pass
- [x] Linting passes
- [x] Documentation is clear and complete


demos/taco-aa-signing/src/index.ts
```diff
@@ -0,0 +1,234 @@
...
+  createPaymasterClient,
+} from 'viem/account-abstraction';
+import { privateKeyToAccount } from 'viem/accounts';
+import { sepolia } from 'viem/chains';
+
+dotenv.config();
+
+const SEPOLIA_CHAIN_ID = 11155111;
+const TACO_DOMAIN: Domain = domains.DEVNET;
+const COHORT_ID = 1;
```
cygnusv, line 37:
```suggestion
const COHORT_ID = 2;
```
AFAIK, cohort 1 expired a few days ago, but cohort 2 should be OK until Sept 8th. Although with this I'm getting this error:
```
...
🔏 Signing with TACo...
❌ Demo failed: Unexpected end of JSON input
```

demos/taco-aa-signing/src/index.ts
```diff
@@ -49,21 +50,26 @@ async function createTacoSmartAccount(
     TACO_DOMAIN,
     COHORT_ID,
   );
-  const signers = participants.map((p: any) => p.operator as Address).sort();
+  const signers = participants.map((p) => p.operator as Address).sort();
+
+  // Create a TACo account using the first signer as placeholder address
+  // This satisfies MetaMask's signatory requirement without using a local wallet
+  const tacoAccount = createViemTacoAccount(signers[0]);
```
derekpierre: Is it possible to use the multisig contract address of the cohort instead of the first signer address or is there an internal check that is done to ensure that the address provided is an EOA?

demos/taco-aa-signing/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "taco-aa-signing-demo",
+  "version": "0.0.0",
```
derekpierre: "0.1.0"?

```suggestion
  "version": "0.1.0",
```

demos/taco-aa-signing/README.md
```diff
@@ -0,0 +1,109 @@
+# TACo Account Abstraction Demo
```
derekpierre: Should we point out that this is metamask delegation toolkit specific AA implementation? including in the folder name?

Some of the extra steps are needed because of MDT eg. viem account, local address etc.

demos/taco-aa-signing/src/taco-account.ts
```diff
@@ -0,0 +1,33 @@
...
+    },
+
+    async signTypedData() {
+      return '0x' as `0x${string}`;
+    },
+  });
+}
+
+// Legacy alias - kept for backward compatibility if needed
+export const createTacoAccount = createViemTacoAccount;
```
derekpierre: What's the backward compatibility concern?

demos/README.md
```diff
@@ -3,3 +3,4 @@
 - [`taco-demo`](./taco-demo) - A demo of the `@nucypher/taco` library.
 - [`taco-nft-demo`](./taco-nft-demo) - A demo an NFT-based condition using the
   `@nucypher/taco` library.
+- [`taco-aa-signing`](./taco-aa-signing) - A demo showing TACo distributed signing with Account Abstraction wallets.
```
derekpierre:
```suggestion
- [`taco-aa-signing`](./taco-aa-signing) - A demo showing TACo distributed signing with MetaMask Delegation Toolkit Account Abstraction wallets.
```

demos/taco-aa-signing/src/taco-account.ts
Muhammad-Altabba:
```suggestion
```
 I guess this was left by LLM by a chance as, when iterating on the task, it thought the initial work was adopted.

demos/taco-aa-signing/src/index.ts
```diff
@@ -0,0 +1,242 @@
...
+    TACO_DOMAIN,
+    COHORT_ID,
+  );
+  const signers = participants.map((p) => p.operator as Address).sort();
+
+  // Create a TACo account using the first signer as placeholder address
+  // This satisfies MetaMask's signatory requirement without using a local wallet
+  const tacoAccount = createViemTacoAccount(signers[0]);
+
+  const smartAccount = await toMetaMaskSmartAccount({
```
Muhammad-Altabba, line 64: Is it possible to have the demo as a webpage, in order to check directly with MetaMask? Or is it that we like to run the demo in the command line?

demos/taco-aa-signing/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "description": "A demo showing TACo distributed signing with Account Abstraction wallets",
+  "private": true,
+  "author": "NuCypher <dev@nucypher.com>",
+  "scripts": {
+    "check": "tsx --no-warnings src/index.ts --dry-run 2>/dev/null && echo '✓ Demo syntax check passed' || echo '✓ Demo syntax check passed'",
```
Muhammad-Altabba: Is it needed to bypass the check if it failed? I mean having ` || echo '✓ Demo syntax check passed'` at the end would mean the checks are optional?
```suggestion
    "check": "tsx --no-warnings src/index.ts --dry-run 2>/dev/null && echo '✓ Demo syntax check passed'",
```

demos/taco-aa-signing/src/index.ts
```diff
@@ -0,0 +1,242 @@
...
+    implementation: Implementation.MultiSig,
+    deployParams: [signers, BigInt(threshold)],
+    deploySalt: '0x' as `0x${string}`,
+    signatory: [{ account: tacoAccount }],
+  });
+
+  return { smartAccount, threshold };
+}
+
+async function signUserOpWithTaco(
```
Muhammad-Altabba, line 76: Can we generalize this, making it more type-safe, and have it as part of taco-web? I mean specifically for
```
  const tacoUserOp: UserOperation = {
    sender: userOp.sender,
    nonce: Number(userOp.nonce),
    factory: userOp.factory || '0x',
    factoryData: userOp.factoryData || '0x',
    callData: userOp.callData,
    callGasLimit: Number(userOp.callGasLimit),
    verificationGasLimit: Number(userOp.verificationGasLimit),
    preVerificationGas: Number(userOp.preVerificationGas),
    maxFeePerGas: Number(userOp.maxFeePerGas),
    maxPriorityFeePerGas: Number(userOp.maxPriorityFeePerGas),
    paymaster: userOp.paymaster || '0x',
    paymasterVerificationGasLimit: Number(
      userOp.paymasterVerificationGasLimit || 0,
    ),
    paymasterPostOpGasLimit: Number(userOp.paymasterPostOpGasLimit || 0),
    paymasterData: userOp.paymasterData || '0x',
    signature: '0x',
  };
  ```

File: demos/taco-aa-signing/src/index.ts [hunk 10]
```diff
@@ -0,0 +1,242 @@
...
+import { createViemTacoAccount } from './taco-account';
+
+dotenv.config();
+
+const SEPOLIA_CHAIN_ID = 11155111;
+const TACO_DOMAIN: Domain = domains.DEVNET;
+const COHORT_ID = 1;
+const AA_VERSION = 'mdt';
+
+async function createTacoSmartAccount(
```
Muhammad-Altabba, line 41: Is it possible to generalize this function and put it or part of it inside taco-web?

demos/taco-aa-signing/src/taco-account.ts
```diff
@@ -0,0 +1,33 @@
+import { type Address } from 'viem';
+import { toAccount } from 'viem/accounts';
+
+/**
+ * Creates a minimal Viem Account that serves as a placeholder for the MetaMask Smart Account.
+ * This account is never actually used for signing - all real signing happens through the TACo network
+ * via the separate signUserOpWithTaco function.
```
Muhammad-Altabba, line 7: Do you mean `signUserOp` as it is the one that we have inside taco-web?:
```suggestion
 * via the separate signUserOp function.
```

Muhammad-Altabba (commented):
Great to see AA in this nice demo 🚀 

I had some questions mostly about adding some helper functions to taco-web.

demos/taco-aa-signing/src/index.ts
theref: check what with metamask? I'm not sure what running in the browser would show here

demos/taco-aa-signing/src/index.ts
derekpierre: Extended cohort id 1 for another 6 months.

demos/taco-aa-signing/src/index.ts
theref: so this is to do with converting from a Viem UserOp. I'm opening to making the sdk more general, but I think we should wait until your Viem PR is in. This is just demo code for an adopter

demos/taco-aa-signing/src/index.ts
theref: this is specific to Metamask (i don't know if other AA wallets will require the same). It's also viem specific, so if we do this, Viem becomes a dependency.

Muhammad-Altabba (approved):
LGTM.
Also it is good to have a follow-up task to check and add possible abstracted code to taco-web, if it was generic enough to be so.

derekpierre:
You will need to rebase this over `signing-epic` now that #737 is merged.

demos/taco-mdt-aa-signing/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "start": "tsx src/index.ts",
+    "dev": "tsx src/index.ts --debug",
+    "type-check": "echo \"Note: Full type-check skipped due to viem/delegation-toolkit type incompatibilities\""
+  },
+  "dependencies": {
+    "@metamask/delegation-toolkit": "^0.11.0",
+    "@metamask/delegation-utils": "^0.11.0",
+    "@nucypher/shared": "^0.6.0-alpha.2",
+    "@nucypher/taco": "^0.7.0-alpha.2",
+    "@nucypher/taco-auth": "^0.4.0-alpha.2",
```
derekpierre: Can we update to use the latest alpha.4 changes...

demos/taco-mdt-aa-signing/src/index.ts
```diff
@@ -0,0 +1,252 @@
...
+    maxFeePerGas: Number(userOp.maxFeePerGas),
+    maxPriorityFeePerGas: Number(userOp.maxPriorityFeePerGas),
+    paymaster: userOp.paymaster || '0x',
+    paymasterVerificationGasLimit: Number(
+      userOp.paymasterVerificationGasLimit || 0,
+    ),
+    paymasterPostOpGasLimit: Number(userOp.paymasterPostOpGasLimit || 0),
+    paymasterData: userOp.paymasterData || '0x',
+    signature: '0x',
+  };
```
derekpierre: With alpha.4 packages this should not be needed and would be a great test for those changes. The goal was to accept user operations from various libraries.

packages/shared/src/contracts/agents/signing-coordinator.ts
```diff
@@ -77,6 +77,32 @@ export class SigningCoordinatorAgent {
...
+    const childAddress = await coordinator.getSigningCoordinatorChild(chainId);
+
+    // Create a contract instance for the child (using generic Contract interface)
+    const childContract = new ethers.Contract(
+      childAddress,
+      [
+        // ABI for the cohortMultisigs function
+        'function cohortMultisigs(uint32) view returns (address)',
+      ],
+      provider,
```
derekpierre: This only works if the child contract is on the same chain as the signing coordinator i.e. only the ETH chain. If the child chain is different the same provider cannot be used, because the provider is used for calling the signing coordinator, and therefore won't be on the same chain.

The best you can do here is return the address of the signing coordinator child contract and not actually make a call on the child contract if it is on a different chain.

Unless a separate provider is passed for the relevant child chain.

derekpierre (dismissed):

packages/shared/src/contracts/agents/signing-coordinator.ts
derekpierre: Removed the function.

derekpierre:
@theref I rebased this PR over `signing-epic` and used the latest alpha.5 releases. Also made some tweaks to the logic.

PR #701, opened 2025-08-28
Fix ES module import issues

Muhammad-Altabba:
Fix ES module import issues
+ Add pre-build script to automate the step of adding .js and /index.js to local imports. (once this PR approved the updated files will be pushed).
+ Add script to assign `type: "module"` to package.json at the ES folder.
+ Add and organize the scripts to run before and after build
+ Update some imports of ethers objects to be es compatible

To not make noise with the edited files, once approved or requested will run `pnpm build` which should then:
+ Modify source files to have imports compatible with ES modules.

**Helpful reading:** https://www.embedthis.com/blog/sensedeep/how-to-create-single-source-npm-module.html

**Type of PR:**

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**

**Issues fixed/closed:**

Fixes https://github.com/nucypher/sprints/issues/229

**Why it's needed:**

**Notes for reviewers:**


full-check-like-in-ci.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
manumonti: Can we move this script into the `scripts` folder?

scripts/fix-source-imports.ts
```diff
@@ -0,0 +1,97 @@
...
+      // Recursively process subdirectories
+      fixSourceImports(fullPath);
+    } else if (file.endsWith('.ts')) {
+      // Fix imports in .ts files
+      let content = fs.readFileSync(fullPath, 'utf8');
+      let modified = false;
+
+      // Fix all relative import/export statements to include .js extension
+      const importExportRegex =
+        /((?:import|export)(?:\s+type)?\s*(?:{[^}]*}\s*)?(?:(?:\*\s*as\s+\w+|\w+)\s*)?from\s+['"])([^'"]*?)(['"])/g;
```
manumonti: ...but the language is that of Mordor, which I will not utter here. 🤣

manumonti (changes requested):
Looks good! But something that is not clear to me: what was the problem before this PR? We could use ES modules before, right? What was broken? Why is it needed to add the `.js` extension to the imports?

Muhammad-Altabba:
> Looks good! But something that is not clear to me: what was the problem before this PR? We could use ES modules before, right? What was broken? Why is it needed to add the `.js` extension to the imports?

Thanks for your question. Actually the issue was when an ECMA Script project is importing taco-web packages. Sorry I missed initially pointing to the original issue in the PR description. More details on the issue is at: 
 https://github.com/nucypher/sprints/issues/229

manumonti (approved):
LGTM! 🙌

Muhammad-Altabba:
To not cause noise by the edited files (adding .js), once this PR is approved or if requested, I would run `pnpm build` which should then push a commit that:
- Modify source files to have imports compatible with ES modules.

Also by checking out this branch, when executing `pnpm build` the files will be edited locally. Let me know if I should push this commit or not yet.

derekpierre:
Seems like this PR needs a rebase...

derekpierre:
I'm less familiar with the nuances of ES, CommonJS, Typescript.

Some questions:
* Seems like you are trying to split this PR into two parts:
  1. package.json/script changes
  2. updated imports to use .js extensions etc.

    I think you can employ either of the following strategies:
       1. Do the two changes in this PR as distinct commits, and then we can review commit-by-commit
       OR
       2. this functionality could be implemented in it's own branch (based over `epic-viem`), using separate PRs for each stage, and that branch then gets merged into `epic-viem` once fully completed. That would prevent a half-finished feature from merging into the `epic-viem` branch. 
       
   wdyt?

* I'm interested in how this affects the existing `taco` demos we have in `examples/taco`. The ones that use nodejs, nextjs, react, webpack-5? Do those continue to work as they did before without any changes? If not, does that imply we have backwards compatibility issues with anyone who used `nucypher/taco` in the same way as the demos? Trying to gauge if this is a breaking change, and if so, what exactly it breaks i.e. what does this change mean for already existing usage of `taco-web`?

* Are there any npm publishing/packaging changes we need to make due to this change?

Muhammad-Altabba:
Thanks @derekpierre ,
- I just split the PR into 4 small commits for easier review. 
- There must be no breaking changes. 
- And there must be no changes needed for publishing process.

However, in the meantime, I will iterate once more on the last 2 points.

Muhammad-Altabba:
Hi @derekpierre,
I just double checked. 

All examples runs with no issues. 
And there must be no breaking changes to any project. Except in a case that should not happen. That is if the user at a CommonJS was wrongly referencing (with `import` or `require` a file from the ESM folder with `@nucypher/taco/dist/es/...` but in a CommonJS project. And because the users are not supposed to import or require except in the format `@nucypher/taco`. And because if the users were ever referencing the `.../dist/...`, in a CommonJS module project, they would reference `.../dist/cjs/...` and not `.../dist/es/...`, if ever. I think no need to raise the library major version for this. Or what do you think?

And there should be no need to modify the publishing process.

theref:
ok, in theory i approve this, but I'd like to see the changes that it will make :) @Muhammad-Altabba could you open a PR into this branch that contains: 1 commit with the result of the `pnpm build`

Muhammad-Altabba:
> ok, in theory i approve this, but I'd like to see the changes that it will make :) @Muhammad-Altabba could you open a PR into this branch that contains: 1 commit with the result of the `pnpm build`

https://github.com/nucypher/taco-web/pull/724
Thanks,

theref (approved):
I'm happy with this once https://github.com/nucypher/taco-web/pull/724 is also included :)

cygnusv (approved):

Issue #690, opened 2025-08-08
Add an easy to use condition for AddressAllowList based on ContextVariableCondition


Muhammad-Altabba:
I think it is beneficial for keeping an easy to use condition with a simple name especially for AllowList. For this (as suggested by @derekpierre at some private meeting), we may have a class similar to ERC20Balance and ERC721Balance that wraps around the more abstract condition.

_Originally posted by @Muhammad-Altabba in https://github.com/nucypher/taco-web/pull/688#discussion_r2261172934_

Muhammad-Altabba:
A sub-task of: https://github.com/nucypher/sprints/issues/206

PR #709, opened 2025-09-19
Add an easy to use condition for AddressAllowlist based on ContextVariableCondition

Muhammad-Altabba:
**Type of PR:**

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**

- Fixes https://github.com/nucypher/taco-web/issues/690

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/packages/taco/schema-docs/condition-schemas.md
+++ b/packages/taco/schema-docs/condition-schemas.md
@@ -6,6 +6,7 @@ _Union of the following possible types:_
 
 - [RpcCondition](#rpccondition)
 - [TimeCondition](#timecondition)
+- [AddressAllowlistCondition](#addressallowlistcondition)
 - [ContextVariableCondition](#contextvariablecondition)
 - [ContractCondition](#contractcondition)
 - [EcdsaCondition](#ecdsacondition)
@@ -394,6 +395,20 @@ _(\*) Required._
 Optional operations to perform on the obtained result
 
 _Array of at least 1  and  at most 5 [VariableOperation](#variableoperation) items._ (_optional_)
+
+## AddressAllowlistCondition
+
+Address Allowlist Condition for allowing decryption for specific wallet addresses. It is very handy when combined with other conditions.
+
+_Object containing the following properties:_
+
+| Property                 | Description                                                                                                                                     | Type                                |
+| :----------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------- | :---------------------------------- |
+| **`conditionType`** (\*) |                                                                                                                                                 | `'address-allowlist'`               |
+| **`userAddress`** (\*)   | This is a context variable that will be replaced at decryption time. It represents the Ethereum address of the requester attempting decryption. | [UserAddress](#useraddress)         |
+| **`addresses`** (\*)     | List of wallet addresses allowed to decrypt. Addresses should be provided in checksummed form.                                                  | `Array<string>` (_min: 1, max: 25_) |
+
+_(\*) Required._
 
 ## More resources
 
--- a/packages/taco/schema-docs/condition-schemas.md
+++ b/packages/taco/schema-docs/condition-schemas.md
@@ -6,7 +6,6 @@ _Union of the following possible types:_
 
 - [RpcCondition](#rpccondition)
 - [TimeCondition](#timecondition)
-- [AddressAllowlistCondition](#addressallowlistcondition)
 - [ContextVariableCondition](#contextvariablecondition)
 - [ContractCondition](#contractcondition)
 - [EcdsaCondition](#ecdsacondition)
@@ -398,17 +397,9 @@ _Array of at least 1  and  at most 5 [VariableOperation](#variableoperation) ite
 
 ## AddressAllowlistCondition
 
-Address Allowlist Condition for allowing decryption for specific wallet addresses.
+List of allowed wallet addresses. Addresses should be provided in checksummed form.
 
-_Object containing the following properties:_
-
-| Property                 | Description                                                                                                                                     | Type                                |
-| :----------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------- | :---------------------------------- |
-| **`conditionType`** (\*) |                                                                                                                                                 | `'address-allowlist'`               |
-| **`userAddress`** (\*)   | This is a context variable that will be replaced at decryption time. It represents the Ethereum address of the requester attempting decryption. | [UserAddress](#useraddress)         |
-| **`addresses`** (\*)     | List of wallet addresses allowed to decrypt. Addresses should be provided in checksummed form.                                                  | `Array<string>` (_min: 1, max: 25_) |
-
-_(\*) Required._
+_Array of at least 1  and  at most 25 `string` items._
 
 ## More resources
 
--- a/packages/taco/schema-docs/condition-schemas.md
+++ b/packages/taco/schema-docs/condition-schemas.md
@@ -398,7 +398,7 @@ _Array of at least 1  and  at most 5 [VariableOperation](#variableoperation) ite
 
 ## AddressAllowlistCondition
 
-Address Allowlist Condition for allowing decryption for specific wallet addresses. It is very handy when combined with other conditions.
+Address Allowlist Condition for allowing decryption for specific wallet addresses.
 
 _Object containing the following properties:_
 
--- a/packages/taco/src/conditions/predefined/address-allowlist.ts
+++ b/packages/taco/src/conditions/predefined/address-allowlist.ts
@@ -1,0 +1,49 @@
+import {
+  ContextVariableCondition,
+  ContextVariableConditionProps,
+  ContextVariableConditionType,
+} from '../base/context-variable';
+import { Condition, ERR_INVALID_CONDITION } from '../condition';
+import {
+  AddressAllowlistConditionProps,
+  addressAllowlistConditionSchema,
+  AddressAllowlistConditionType,
+} from '../schemas/address-allowlist';
+import { OmitConditionType } from '../shared';
+
+export {
+  AddressAllowlistConditionProps,
+  AddressAllowlistConditionType,
+} from '../schemas/address-allowlist';
+
+/**
+ * A predefined condition that checks if a user's address is in an allowlist.
+ * This condition uses a ContextVariableCondition internally to check if the
+ * specified context variable (e.g., ':userAddress') is in the provided list
+ * of allowed addresses.
+ */
+export class AddressAllowlistCondition extends ContextVariableCondition {
+  constructor(value: OmitConditionType<AddressAllowlistConditionProps>) {
+    const { data, error } = Condition.validate(
+      addressAllowlistConditionSchema,
+      {
+        conditionType: AddressAllowlistConditionType,
+        ...value,
+      },
+    );
+    if (error) {
+      throw new Error(ERR_INVALID_CONDITION(error));
+    }
+
+    const contextVariableCondition: ContextVariableConditionProps = {
+      conditionType: ContextVariableConditionType,
+      contextVariable: data.userAddress,
+      returnValueTest: {
+        comparator: 'in',
+        value: data.addresses,
+      },
+    };
+
+    super(contextVariableCondition);
+  }
+}
--- a/packages/taco/src/conditions/predefined/address-allowlist.ts
+++ b/packages/taco/src/conditions/predefined/address-allowlist.ts
@@ -1,49 +1,42 @@
+import { USER_ADDRESS_PARAM_DEFAULT } from '@nucypher/taco-auth';
+
 import {
   ContextVariableCondition,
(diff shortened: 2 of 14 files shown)


packages/taco/src/conditions/predefined/address-allowlist.ts
```diff
@@ -0,0 +1,49 @@
...
+      contextVariable: data.userAddress,
+      returnValueTest: {
+        comparator: 'in',
+        value: data.addresses,
+      },
+    };
+
+    super(contextVariableCondition);
+  }
+}
```
Muhammad-Altabba, line 42: Do you think better to override `toObj()` to something like:
```ts
  public override toObj(): AddressAllowlistConditionProps {
    const { data, error } = Condition.validate(this.schema, this.value);
    if (error) {
      throw new Error(ERR_INVALID_CONDITION(error));
    }
    return {
      conditionType: AddressAllowlistConditionType,
      userAddress: data.contextVariable,
      addresses: data.returnValueTest.value,
    };
  }
```

packages/taco/src/conditions/schemas/address-allowlist.ts
```diff
@@ -0,0 +1,27 @@
...
+      .array(EthAddressSchemaStrict)
+      .min(1, 'At least one address must be provided')
+      .max(25, 'A maximum of 25 addresses is allowed')
+      .describe(
+        'List of wallet addresses allowed to decrypt. Addresses should be provided in checksummed form.',
+      ),
+  })
+  .strict()
+  .describe(
+    'Address Allowlist Condition for allowing decryption for specific wallet addresses. It is very handy when combined with other conditions.',
```
manumonti: nitpick, but maybe it's too much text? "It is very handy..." sounds to me more like an opinion than an objective information.
```suggestion
    'Address Allowlist Condition for allowing decryption for specific wallet addresses.',
```

packages/taco/src/conditions/predefined/address-allowlist.ts
manumonti: Mmmh honestly I don't have a preferred option. Both should work fine, right?

manumonti (approved):
This looks great! I like it. I've left just a nitpicky comment, but looks goot do me.

packages/taco/src/conditions/predefined/address-allowlist.ts
Muhammad-Altabba: Oh sorry, I looked at it again. toObj() should not be override here. Because this defines how the object will be serialized to JSON, and I guess it would cause the nodes to not be able to identify the object.

theref:
Could we please clean up these 3 commits? They're 1 line each and don't have meaningful commit messages.

Screenshot 2025-10-02 at 11 29 03

packages/taco/src/conditions/schemas/address-allowlist.ts
```diff
@@ -0,0 +1,27 @@
+import { EthAddressSchemaStrict } from '@nucypher/shared';
+import { z } from 'zod';
+
+import { baseConditionSchema, UserAddressSchema } from './common';
+
+export const AddressAllowlistConditionType = 'address-allowlist';
+
+export const addressAllowlistConditionSchema = baseConditionSchema
+  .extend({
+    conditionType: z.literal(AddressAllowlistConditionType),
```
derekpierre: We don't need to define a new condition type but rather simply reuse the `context-variable` type / `ContextVariableCondition` in a specific way. Otherwise, you would need this type to also be supported on the node side, which it is not. 

Think of this as just a wrapper around setting specific values for a `ContextVariableCondition`.

Instead `AddressAllowListCondition` should simply extend `ContextVariableCondition`, and have specific defaults for the properties. See `erc20.ts` and `erc721.ts` as a template for how they extended the contract condition.

derekpierre (changes requested):

packages/taco/src/conditions/schemas/address-allowlist.ts
Muhammad-Altabba: > We don't need to define a new condition type but rather simply reuse the `context-variable` type / `ContextVariableCondition` in a specific way. Otherwise, you would need this type to also be supported on the node side, which it is not.


Defining a new schema does not necessitate the introduction of a new node side condition. It is just for client-side input validation. The integration tests pass when running `export RUNNING_IN_CI=true && p integration-test encrypt-decrypt.test.ts`.
And the proposed `AddressAllowlistCondition` is not just filling defaults similar to erc20.ts  and erc721.ts . But rather provide a meaningful-for-the-case schema which transforms into a ContextVariableCondition.

So, probably the code comments to describe this could be enhanced?
By modifying the comment for condition object to be something like?:
```
/**
 * A Client-side condition that checks if a user's address is in an allowlist and transforms the object into ContextVariableCondition.
 * The nodes process this as a standard context variable that checks if the user's address is in the allowlist.
 */
export class AddressAllowlistCondition extends ContextVariableCondition ...
```

And adding a comment for schema to be something like?:
```
/**
 * Client-side abstraction which validate the type that will be transformed internally to ContextVariableCondition.
 * So, this schema is just for the validation on an intuitive type. But the nodes receive and process ContextVariableCondition.
 */
export const addressAllowlistConditionSchema = ...
```

The whole point of the newly added condition is to provide DX with an intuitive object properties. Or do you think of another schema better than the proposed one?

Or what do you think?

derekpierre: The new schema is not the issue, it's the new condition string type `"address-allowlist"`. 

You are doing a level of indirection that I don't believe is necessary:
```typescript
export class AddressAllowlistCondition extends ContextVariableCondition {
  constructor(value: OmitConditionType<AddressAllowlistConditionProps>) {
    const { data, error } = Condition.validate(
      addressAllowlistConditionSchema,
      {
        conditionType: AddressAllowlistConditionType,
        ...value,
      },
    );
    if (error) {
      throw new Error(ERR_INVALID_CONDITION(error));
    }

    const contextVariableCondition: ContextVariableConditionProps = {
      conditionType: ContextVariableConditionType,
      contextVariable: data.userAddress,
      returnValueTest: {
        comparator: 'in',
        value: data.addresses,
      },
    };

    super(contextVariableCondition);
```
^ Here you are taking one set of props (from `addressAllowlistConditionSchema`) and converting it to another set of props (from `contextvariableschema`), and what I'm saying is that this indirection is not necessary. 

Instead, you can just directly extend the `contextvariableschema` which has the same fields and just limit the underlying values that can be set for those fields.

You don't need something like a new condition string type, because an `AddressAllowListCondition` is just a `ContextVariableCondition` with specific values. 

Using an extend type of approach saves the need for code like this in the factory:
```typescript
      case AddressAllowlistConditionType:
        return new AddressAllowlistCondition(
          props as AddressAllowlistConditionProps,
        );

```
because the serialized/deserialized type would simply be a `ContextVariableConditionType` which is already handled by the factory.

derekpierre (changes requested):
The "addess-allowlist" additional type is not needed. See comment.

packages/taco/src/conditions/schemas/address-allowlist.ts
Muhammad-Altabba: I think that not having a unique conditionType would make this condition an outlaw compared to other conditions, preventing it from being *intuitively* usable at the TACo Play Ground https://seal-app-y6kp4.ondigitalocean.app/ and other similar cases where the construction of the condition comes from a JSON object. As the user would pass the JSON of the ContextVariableCondition instead. That is not a problem for ERC20 and ERC721 Balance Conditions because it is still easy to map to the underling condition.

And not having a unique schema for the Address AllowList Condition would prevent using it at `anyConditionSchema` that is used at `CompoundCondition`, `IfThenElseCondition` and `SequentialCondition`. This would cause not being usable like this:
```
      const overallCondition = new SequentialCondition({
        conditionVariables: [
          ...,
          {
            varName: 'addressAllowlist',
            condition: {
              conditionType: AddressAllowlistConditionType,
              userAddress: ':userAddress',
              addresses: [
                '0x0000000000000000000000000000000000000000',
                '0x0000000000000000000000000000000000000001',
              ],
            },
          },
        ],
      });
      const conditionExpr = new ConditionExpression(overallCondition);
```
Right?

packages/taco/src/conditions/predefined/address-allowlist.ts
```diff
@@ -0,0 +1,42 @@
...
+  AddressAllowlistConditionProps,
+  addressAllowlistConditionSchema,
+} from '../schemas/address-allowlist';
+
+export { AddressAllowlistConditionProps } from '../schemas/address-allowlist';
+
+/**
+ * A Client-side condition that checks if a user's address is in an allowlist and transforms the object into `ContextVariableCondition`.
+ * The nodes process this as a standard context variable that checks if the user's address is in the allowlist.
+ * @remark This condition does not have a conditionType field and is not possible to be composed to/from JSON except as the underlying type `ContextVariableCondition`.
```
derekpierre:
```suggestion
 * @remark This condition doesn’t have a unique type of its own; it is simply a wrapper for creating a `ContextVariableCondition`.
```

packages/taco/schema-docs/condition-schemas.md
```diff
@@ -395,6 +396,20 @@ Optional operations to perform on the obtained result
...
 
+## AddressAllowlistCondition
+
+Address Allowlist Condition for allowing decryption for specific wallet addresses.
+
+_Object containing the following properties:_
+
+| Property                 | Description                                                                                                                                     | Type                                |
+| :----------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------- | :---------------------------------- |
+| **`conditionType`** (\*) |                                                                                                                                                 | `'address-allowlist'`               |
```
derekpierre: Looks like you need to rerun the `generate-zod-docs` script to update this file.

packages/taco/src/conditions/predefined/address-allowlist.ts
```diff
@@ -0,0 +1,42 @@
...
+    const contextVariableCondition: ContextVariableConditionProps = {
+      conditionType: ContextVariableConditionType,
+      contextVariable: USER_ADDRESS_PARAM_DEFAULT,
+      returnValueTest: {
+        comparator: 'in',
+        value: data,
+      },
+    };
+
+    super(contextVariableCondition);
```
derekpierre: minor nitpick:

```suggestion
    const contextVariableConditionProps: ContextVariableConditionProps = {
      conditionType: ContextVariableConditionType,
      contextVariable: USER_ADDRESS_PARAM_DEFAULT,
      returnValueTest: {
        comparator: 'in',
        value: data,
      },
    };

    super(contextVariableConditionProps);
```

packages/taco/src/conditions/schemas/address-allowlist.ts
```diff
@@ -0,0 +1,14 @@
+import { EthAddressSchemaStrict } from '@nucypher/shared';
+import { z } from 'zod';
+
+export const addressAllowlistConditionSchema = z
+  .array(EthAddressSchemaStrict)
+  .min(1, 'At least one address must be provided')
+  .max(25, 'A maximum of 25 addresses is allowed')
+  .describe(
+    'List of wallet addresses allowed to decrypt. Addresses should be provided in checksummed form.',
```
derekpierre: Let's not include "decrypt" in the description, since it can potentially be used as a signing condition as well.

```suggestion
    'List of allowed wallet addresses. Addresses should be provided in checksummed form.',
```

packages/taco/src/conditions/predefined/address-allowlist.ts
```diff
@@ -0,0 +1,42 @@
...
+export { AddressAllowlistConditionProps } from '../schemas/address-allowlist';
+
+/**
+ * A Client-side condition that checks if a user's address is in an allowlist and transforms the object into `ContextVariableCondition`.
+ * The nodes process this as a standard context variable that checks if the user's address is in the allowlist.
+ * @remark This condition does not have a conditionType field and is not possible to be composed to/from JSON except as the underlying type `ContextVariableCondition`.
+ */
+export class AddressAllowlistCondition extends ContextVariableCondition {
+  constructor(value: AddressAllowlistConditionProps) {
+    const { data, error } = Condition.validate(
```
derekpierre: nitpick here. I would use `AddressAllowlistCondition`. `validate` is a static function available as part of all subclasses. You'll also not need to import `Condition`.
```suggestion
    const { data, error } = AddressAllowlistCondition.validate(
```

packages/taco/test/conditions/predefined/address-allowlist.test.ts
derekpierre, line 1: Nice tests.

derekpierre (approved):
🎸 - nice work!

PR #724, opened 2025-10-08
[DRAFT] Add js extension to files import

Muhammad-Altabba:
This contains the branch a https://github.com/nucypher/taco-web/pull/701 after executing:
`pnpm build && pnpm lint:fix`

The `pnpm lint:fix` is executed because the order of few imports will be changed after adding /index.js and so.


theref:
ok perfect, I'm happy with this

theref (approved):

Issue #712, opened 2025-09-29
Investigate and fix ens resolution with viem objects


Muhammad-Altabba:
Un-skip and check the test at:
https://github.com/nucypher/taco-web/pull/700/files/46d6160b4837dcf91b6bb6f6ebfa4a4c28152662#diff-4697f0050694a3e1019fbf99dfb42b6cade6e626ede67c3eb76aa43ca1ed30fdR9

Preliminary investigation result: viem chain config does not (in the last version) include the ens contract address. For ethers.js if a viem chain object was passed and since that object would not have an ens address, the new ethers provider will fail to resolve.

PR #725, opened 2025-10-12
fix: handle missing chain and ENS registry in viem-ethers adapter

Muhammad-Altabba:
**Type of PR:**

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**

This PR improves ENS resolution handling in the viem-to-ethers adapter by:
- Making the `chain` parameter optional in [viemClientToProvider()](cci:1://file:///Users/funcy/repos/taco/taco-web/packages/shared/src/viem/ethers-adapter.ts:26:0-123:1) to support viem clients without an explicit chain configuration
- Adding a developer-friendly warning when ENS registry is not available on the configured chain
- Adding comprehensive integration tests covering three ENS resolution scenarios:
  1. When no viem chain object is provided (ethers will automatically use the provider's default chain)
  2. When viem chain has explicit ENS registry configuration 
  3. When viem chain lacks ENS registry configuration (mainnet from viem/chains)

**Issues fixed/closed:**

 - Fixes https://github.com/nucypher/taco-web/issues/712

**Notes for reviewers:**

The adapter was previously requiring a chain configuration even when not strictly necessary, and failing silently when ENS wasn't properly configured.
But in last viem version the `contracts.ensRegistry.address` is not filled. For main net for example, it contain:
```
mainnet.contracts: {
  "ensUniversalResolver": {
    "address": "0xeeeeeeee14d718c2b47d9923deab1335e144eeee",
    "blockCreated": 23085558
  },
  "multicall3": {
    "address": "0xca11bde05977b3631167028862be2a173976ca11",
    "blockCreated": 14353601
  }
}
```

- So if viem `contracts.ensRegistry.address` was not provided, the ethers `networkish.ensAddress` would not be filled and a console.warn will be logged. Because otherwise using ensUniversalResolver to get the ensRegistry will require an async contract function call.
- But, if viem `contracts.ensRegistry.address` was explicitly filled by the user, it will be utilized to fill `networkish.ensAddress`. 
- However, if the user passed just the url or an http provider, without filling the `chain` property. Then ethers.js will automatically fill the ensAddress and work with no issue. **And this is the most likely case to happen.**

The tests should describe the 3 mentioned cases well.

And the behavior would be documented when working at https://github.com/nucypher/sprints/issues/265.

### DIFF
--- a/packages/shared/src/viem/ethers-adapter.ts
+++ b/packages/shared/src/viem/ethers-adapter.ts
@@ -15,6 +15,10 @@ function createNetworkish(chain: Chain): ethers.providers.Networkish {
   // Add ENS registry address if available
   if (chain.contracts?.ensRegistry?.address) {
     networkish.ensAddress = chain.contracts.ensRegistry.address;
+  } else {
+    console.warn(
+      `No ENS registry found on chain ${chain.name} (chainId=${chain.id}) at chain.contracts.ensRegistry.address. Resolving ENS will not work on the created ethers Provider.`,
+    );
   }
 
   return networkish;
@@ -41,11 +45,10 @@ export function viemClientToProvider(
 ): ethers.providers.Provider {
   const { chain, transport }: PublicClient = client;
 
-  if (!chain) {
-    throw new Error('Client must have a chain configured');
+  let networkish: ethers.providers.Networkish | undefined;
+  if (chain) {
+    networkish = createNetworkish(chain);
   }
-
-  const networkish = createNetworkish(chain);
 
   // Note: We read minimal, commonly-present properties from transport.
   // viem's transport internals are not a public API and may change.
--- a/packages/shared/src/viem/ethers-adapter.ts
+++ b/packages/shared/src/viem/ethers-adapter.ts
@@ -17,7 +17,11 @@ function createNetworkish(chain: Chain): ethers.providers.Networkish {
     networkish.ensAddress = chain.contracts.ensRegistry.address;
   } else {
     console.warn(
-      `No ENS registry found on chain ${chain.name} (chainId=${chain.id}) at chain.contracts.ensRegistry.address. Resolving ENS will not work on the created ethers Provider.`,
+      `No ENS registry found on chain ${chain.name} (chainId=${chain.id}).\n` +
+        `With the current configuration, ENS resolution will not work on the created ethers Provider.\n` +
+        `To fix this either:\n` +
+        `  - Set chain.contracts.ensRegistry.address to the correct ENS registry address, or\n` +
+        `  - Or omit the \`chain\` data to allow automatic ENS registry detection from the provider.`,
     );
   }
 
--- a/packages/taco/integration-test/viem-to-ethers-ens.test.ts
+++ b/packages/taco/integration-test/viem-to-ethers-ens.test.ts
@@ -1,30 +1,21 @@
-import { describe, expect, test } from 'vitest';
+import { describe, expect, test, vi } from 'vitest';
 
 import { toEthersProvider } from '@nucypher/shared';
 import { createPublicClient, http } from 'viem';
 import { mainnet } from 'viem/chains';
 
-describe.skipIf(!process.env.RUNNING_IN_CI)('Viem-Ethers Adapter Integration Tests', () => {
-  describe('ENS Registry', () => {
-    // TODO: will fix this in a different PR - issue #712
-    test.skip('should properly read ENS registry contract address from viem client converted to ethers provider', async () => {
+describe.skipIf(!process.env.RUNNING_IN_CI)(
+  'Viem-Ethers Adapter ENS Integration Tests',
+  () => {
+    test('should properly resolve ENS name to address when no viem chain object is provided', async () => {
       // Test with chain that has ENS registry (mainnet)
       // Note: mainnet from viem/chains includes ENS registry configuration
       const mainnetViemClient = createPublicClient({
-        chain: mainnet,
         transport: http('https://eth.llamarpc.com'),
       });
 
       // Convert to ethers provider to verify ENS address mapping
       const ethersProvider = toEthersProvider(mainnetViemClient);
-
-      // Verify ENS registry is properly mapped from viem chain configuration
-      const expectedEnsAddress = mainnet.contracts?.ensRegistry?.address;
-      expect(ethersProvider.network.ensAddress).toBe(expectedEnsAddress);
-      // Also verify it's the known mainnet ENS registry address
-      expect(ethersProvider.network.ensAddress).toBe(
-        '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e',
-      );
 
       // Test actual ENS name resolution to verify functionality
       const resolvedAddress = await ethersProvider.resolveName('vitalik.eth');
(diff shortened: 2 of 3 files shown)


theref:
did we add support for ethers within TACo?

packages/taco/integration-test/viem-to-ethers-ens.test.ts
```diff
@@ -1,37 +1,90 @@
...
-      // Verify ENS registry is properly mapped from viem chain configuration
-      const expectedEnsAddress = mainnet.contracts?.ensRegistry?.address;
-      expect(ethersProvider.network.ensAddress).toBe(expectedEnsAddress);
-      // Also verify it's the known mainnet ENS registry address
+      // Test actual ENS name resolution to verify functionality
+      const resolvedAddress = await ethersProvider.resolveName('vitalik.eth');
+      expect(resolvedAddress).toBeTruthy();
+      // Currently vitalik.eth is "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045".
+      // But it may change in the future. So we only check if it's a valid Ethereum address.
+      expect(resolvedAddress).toMatch(/^0x[a-fA-F0-9]{40}$/); // Valid Ethereum address format
```
derekpierre: I believe that you can use `ethers.utils.isAddress(address)` instead to validate the address.

packages/taco/integration-test/viem-to-ethers-ens.test.ts
```diff
@@ -1,37 +1,90 @@
...
       expect(ethersProvider.network.ensAddress).toBe(
         '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e',
       );
 
       // Test actual ENS name resolution to verify functionality
       const resolvedAddress = await ethersProvider.resolveName('vitalik.eth');
       expect(resolvedAddress).toBeTruthy();
-      // Currently vitalik.eth is "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045".
-      // But it may change in the future. So we only check if it's a valid Ethereum address.
       expect(resolvedAddress).toMatch(/^0x[a-fA-F0-9]{40}$/); // Valid Ethereum address format
```
derekpierre: Same comment here about using `ethers.utils.isAddress(address)`

packages/taco/integration-test/viem-to-ethers-ens.test.ts
```diff
@@ -1,37 +1,90 @@
...
+
+    test('should properly handle viem chain with explicit ENS registry configuration', async () => {
+      const chainWithEns = {
+        ...mainnet,
+        contracts: {
+          ...mainnet.contracts,
+          ensRegistry: {
+            // Manually providing ENS registry since viem/chains doesn't include it
+            address:
+              '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e' as `0x${string}`,
```
derekpierre, line 40: Should we instead test that a random non-default address is configured as the resolver on ethers provider if that non-default address is specified here? 

Of course, you won't actually be able to resolve `vitalik.eth` onL61-63 but at least you'll know that the default address is overridden. Given that the default value is still specified here in the test, it is unclear that the default ens resolver address was indeed set based on the value specified here.

packages/shared/src/viem/ethers-adapter.ts
```diff
@@ -15,6 +15,10 @@ function createNetworkish(chain: Chain): ethers.providers.Networkish {
   // Add ENS registry address if available
   if (chain.contracts?.ensRegistry?.address) {
     networkish.ensAddress = chain.contracts.ensRegistry.address;
+  } else {
+    console.warn(
+      `No ENS registry found on chain ${chain.name} (chainId=${chain.id}) at chain.contracts.ensRegistry.address. Resolving ENS will not work on the created ethers Provider.`,
```
derekpierre: Given that this is just a log message,and not raising an error, what exactly is the failure case? Does it mean that if someone provides a ens address that at some point when the provider is used resolution will fail and an error will be thrown? When does that failure happen exactly and what does that failure mode look like i.e. what is the error?

packages/taco/integration-test/viem-to-ethers-ens.test.ts
Muhammad-Altabba: The last version of viem does not fill the `ensRegistry`. More details about the behavior is in the PR description.

packages/shared/src/viem/ethers-adapter.ts
Muhammad-Altabba: Yes it will fail later if ens operation was performed. For that, I tried to make the failure case clear in the warn message `"... Resolving ENS will not work on the created ethers Provider."` But, now I made it more clear. Let me know please, if it could be better.

Also, I just I updated the test to try to resolve and expect the error.
The full error stack would be like:
> Error: network does not support ENS (operation="getResolver", network="Ethereum", code=UNSUPPORTED_OPERATION, version=providers/5.8.0)
 ❯ Logger.makeError ../../../node_modules/.pnpm/@ethersproject+logger@5.8.0/node_modules/@ethersproject/logger/src.ts/index.ts:269:28
 ❯ Logger.throwError ../../../node_modules/.pnpm/@ethersproject+logger@5.8.0/node_modules/@ethersproject/logger/src.ts/index.ts:281:20
 ❯ JsonRpcProvider.<anonymous> ../../../node_modules/.pnpm/@ethersproject+providers@5.8.0_bufferutil@4.0.9_utf-8-validate@5.0.10/node_modules/@ethersproject/providers/src.ts/base-provider.ts:1989:20
 ❯ step ../../../node_modules/.pnpm/@ethersproject+providers@5.8.0_bufferutil@4.0.9_utf-8-validate@5.0.10/node_modules/@ethersproject/providers/lib/base-provider.js:48:23
 ❯ Object.next ../../../node_modules/.pnpm/@ethersproject+providers@5.8.0_bufferutil@4.0.9_utf-8-validate@5.0.10/node_modules/@ethersproject/providers/lib/base-provider.js:29:53
 ❯ fulfilled ../../../node_modules/.pnpm/@ethersproject+providers@5.8.0_bufferutil@4.0.9_utf-8-validate@5.0.10/node_modules/@ethersproject/providers/lib/base-provider.js:20:58

manumonti: +1 on this. Also, I'm not sure how usual is to have a Viem Chain that doesn't include ENS registry, but I would avoid warning messages unless we know that ENS is indeed going to be used.

packages/taco/integration-test/viem-to-ethers-ens.test.ts
derekpierre: My comment is less about the `ensRegistry` not getting filled, it's that we should verify that the ens registry address can be properly set/overridden based on the viem config.

i.e. if you do something like this on L33

```typescript
const chainWithEns = {
        ...mainnet,
        contracts: {
          ...mainnet.contracts,
          ensRegistry: {
            // Manually providing ENS registry since viem/chains doesn't include it
            address:
              '0x12345678890...' as `0x${string}`
```

(set the ens registry address to some random address)

Then L57 can be:

```typescript
      expect(ethersProvider.network.ensAddress).toBe(
        '0x1234567890...',
      );
```

which verifies that the ensAddress was set properly based on the viem config.

I think that would be a good test. Of course, since the ens address is a random address, you can't subsequently actually resolve "vitalik.eth" so you can't actually do L62-64.

This way we actually verify that the ens resolver address in viem was set to a non-default value in the corresponding ethers provider. At the moment, the test assumes it gets set but the value is the same as the default, so it's possible the default was just being used; not that it is, but it's just not verified in the test.

packages/shared/src/viem/ethers-adapter.ts
derekpierre: I think the original message was sufficient.

I just wanted to better understand at what point in the execution stack is ens resolution performed and the error raised.

manumonti (changes requested):

packages/taco/integration-test/viem-to-ethers-ens.test.ts
Muhammad-Altabba: Thanks Derek,
Actually the value set in the test is not the same as the default in viem. Because there is no default in viem, it is just not set. As in the PR description viem just does not fill it:
```
mainnet.contracts: {
  "ensUniversalResolver": {
    "address": "0xeeeeeeee14d718c2b47d9923deab1335e144eeee",
    "blockCreated": 23085558
  },
  "multicall3": {
    "address": "0xca11bde05977b3631167028862be2a173976ca11",
    "blockCreated": 14353601
  }
}
```

And checking if an arbitrary value set at `ensRegistry` is reflected on the ethers provider exists at a unit test https://github.com/nucypher/taco-web/blob/251c3ea7d3d326f2489c5f8e455e33ab12da6a9e/packages/shared/test/viem-ethers-adapter.test.ts#L41

However in the last commit also the other test, when no value is provided at `ensRegistry`, now an exception is insured to be raised when trying to resolve an ens name.

Please, let me know if you think a new unit or integration test is still needed.

packages/shared/src/viem/ethers-adapter.ts
Muhammad-Altabba: @manumonti,
Unfortunately I do not think we can predict if ens would be used or not later. 

Also passing the `chain` data is not expected to happen as the url is sufficient and the user would just pass the rpc url and everything should work with no warning and no issues. So the warning will come just in an edge case which should alert the developer to modify the code properly and stop receiving this warning.

Except if you has a different take of this?

packages/taco/integration-test/viem-to-ethers-ens.test.ts
derekpierre: Ah gotcha - thanks for the link. Didn't realize we already tested that. No problem then.

packages/shared/src/viem/ethers-adapter.ts
manumonti: Got it. Thanks for your kind explanations!

manumonti (approved):

derekpierre (approved):
🎸

PR #728, opened 2025-10-16
Chore: few final touches on epic-viem

Muhammad-Altabba:
**Type of PR:**

- [ ] Bugfix
- [] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [ ] 2
- [x] 3

**What this does:**

This PR finalizes the AccessClient API design and viem client test by:

1. (first commit) **Moving signer from constructor to encrypt methods** - The `AccessClient` configuration no longer requires a signer/account during instantiation. Instead, signers are now passed directly to encryption methods where they're actually needed.

**API Changes:**

**Before:**
```typescript
const accessClient = new AccessClient({
  domain: DOMAIN_NAMES.TESTNET,
  ritualId: 6,
  viemClient,
  viemAccount  // ❌ Signer in constructor
});

const messageKit = await accessClient.encrypt('Hello!', condition);
```

**After:**
```typescript
const accessClient = new AccessClient({
  domain: DOMAIN_NAMES.TESTNET,
  ritualId: 6,
  viemClient  // ✅ No signer needed
});

const messageKit = await accessClient.encrypt('Hello!', condition, viemAccount);  // ✅ Signer as parameter
```

**Configuration interfaces updated:**
- Removed `viemAccount` from `AccessClientViemConfig`
- Removed `ethersSigner` from `AccessClientEthersConfig`
- Added `authSigner`/`authAccount` parameters to `encrypt()` and `encryptWithPublicKey()` methods

2. (second commit) **Simplifying test setup** - Removed unnecessary chain configuration from viem client creation across test files.

**Why it's needed:**

This refactor improves the AccessClient API design for several reasons:

1. **Better separation of concerns** - Signers are only required for encryption operations (to authenticate the encryptor), not for decryption or client instantiation. The new API reflects this reality.

2. **Cleaner configuration** - The AccessClient configuration now only contains what's truly needed for the client's core functionality (domain, ritual ID, provider/client). As the signer is not needed directly for `decrypt(...)` function.

And regarding the viem client creation at tests:
1. Passing chain was not really needed.
2. It was casing a warning because the chain passed did not have ensRegistry contract address.

**Notes for reviewers:**

- **Focus areas:**
  - `packages/taco/src/access-client/client.ts` - New encrypt method signatures with overloads for ethers/viem signers
  - `packages/taco/src/access-client/config.ts` - Removal of signer-related properties from config interfaces
  - `packages/taco/README.md` - Updated usage examples
  - `packages/taco/integration-test/access-client.test.ts` - Comprehensive test updates showing the new API pattern

### DIFF
--- a/packages/shared/test/viem-ethers-adapter.test.ts
+++ b/packages/shared/test/viem-ethers-adapter.test.ts
@@ -175,10 +175,6 @@ describe('viem ethers adapter', () => {
   describe('toEthersProvider', () => {
     it('should create provider from viem client', async () => {
       const viemClient = createPublicClient({
-        chain: {
-          id: 80002,
-          name: 'Polygon Amoy',
-        },
         transport: http('https://test.com'),
       });
       const provider = toEthersProvider(viemClient);
--- a/packages/taco/integration-test/viem-to-ethers-ens.test.ts
+++ b/packages/taco/integration-test/viem-to-ethers-ens.test.ts
@@ -71,7 +71,7 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
         transport: http(),
       });
 
-      const consoleSpy = vi.spyOn(console, 'warn');
+      const consoleSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
 
       // Convert to ethers provider
       const ethersProvider = toEthersProvider(mainnetViemClient);
@@ -87,7 +87,7 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
       expect(ethersProvider.network.ensAddress).toBeUndefined();
 
       // ENS operation is expected to fail
-      expect(ethersProvider.resolveName('vitalik.eth')).rejects.toThrow(
+      await expect(ethersProvider.resolveName('vitalik.eth')).rejects.toThrow(
         'network does not support ENS',
       );
     });
--- a/packages/taco/integration-test/viem-to-ethers-ens.test.ts
+++ b/packages/taco/integration-test/viem-to-ethers-ens.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, test, vi } from 'vitest';
+import { describe, expect, test } from 'vitest';
 
 import { toEthersProvider } from '@nucypher/shared';
 import { ethers } from 'ethers';
@@ -27,9 +27,9 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
 
       // Currently ethersProvider.network.ensAddress on mainnet is "0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e".
       expect(ethersProvider.network.ensAddress).toBeTruthy();
-    }, 15000);
+    }, 60000);
 
-    test('should properly handle viem chain with explicit ENS registry configuration', async () => {
+    test('ENS operations should fail when viem chain lacks ENS registry address', async () => {
       const chainWithEns = {
         ...mainnet,
         contracts: {
@@ -50,9 +50,6 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
       // Convert to ethers provider to verify ENS address mapping
       const ethersProvider = toEthersProvider(mainnetViemClient);
 
-      // Don't check for console.warn since the behavior might vary
-      // The important part is that ENS registry is properly set when provided
-
       // Verify ENS registry address is properly set
       expect(ethersProvider.network.ensAddress).toBe(
         '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e',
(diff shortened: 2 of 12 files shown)


manumonti (approved):
LGTM! 👌 I don't have any particular comment.

theref (approved):

packages/shared/test/viem-ethers-adapter.test.ts
```diff
@@ -175,10 +175,6 @@ describe('viem ethers adapter', () => {
   describe('toEthersProvider', () => {
     it('should create provider from viem client', async () => {
       const viemClient = createPublicClient({
-        chain: {
-          id: 80002,
-          name: 'Polygon Amoy',
-        },
```
derekpierre, line 181: > And regarding the viem client creation at tests:
> 
> Passing chain was not really needed.
> It was casing a warning because the chain passed did not have ensRegistry contract address.

Is the warning a big deal? Is passing the `chain` something that common or uncommon with viem usage?

Wondering if we keep the warning or just don't bother having it.

Muhammad-Altabba: TLDR: 
With this PR, `chain` object is still passed when its relevant code is tested. And all the occurrences where it is removed are for tests which are irrelevant to it.

The warning is just a distraction that is shown more than 10 times before the changes in this PR. So, they are not a big deal but confusing to see and they distract when checking the tests logs.
Also worth notting that those warning are raised by us when there is a `chain` passed that does not have ensRegistry contract address.
So to get red of the warning we can:
- Pass the ensRegistry contract address with every creation of viem public client when passing the `chain` like in:
https://github.com/nucypher/taco-web/blob/1bf799a914be0acb64b725cef074a967ee128df9/packages/shared/test/viem-ethers-adapter.test.ts#L40
- Or, mock the implementation of console.warn at every creation of viem public client and then restore the mock after. Like in:
https://github.com/nucypher/taco-web/blob/52db4e249eb857ef9c181e175be95f7a2f53f5ce/packages/taco/integration-test/viem-to-ethers-ens.test.ts#L74
https://github.com/nucypher/taco-web/blob/52db4e249eb857ef9c181e175be95f7a2f53f5ce/packages/taco/integration-test/viem-to-ethers-ens.test.ts#L84
- Or, to not pass the `chain` object. And this is what has been adopted in this PR.

And yes it is common to pass the `chain` object when creating a public client. But I chose the third option because passing the chain object is already tested in both unit and integration tests. And more importantly in the rest of the tests, those where the `chain` is now not passed, it is irrelevant to those tests to have the chain object or not.

I hope this address your question. However, please let me know if you have further concern or suggestion.

derekpierre: Let me clarify.

My concern isn’t about the test itself, but about common usage patterns in real applications.

Should we remove the warning message? It seems excessive and confusing if passing a chain is common practice in viem. Also, it's unclear how much value it provides since ENS isn't that useful for TACo e.g. we don’t currently support it for the `:userAddress` context variable.

packages/taco/integration-test/viem-to-ethers-ens.test.ts
```diff
@@ -27,9 +27,9 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
 
       // Currently ethersProvider.network.ensAddress on mainnet is "0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e".
       expect(ethersProvider.network.ensAddress).toBeTruthy();
-    }, 15000);
+    }, 60000);
```
derekpierre: How come this test takes so long?

(https://github.com/nucypher/taco-web/actions/runs/18656495773/job/53186956758#step:8:205)
```
✓ Viem-Ethers Adapter ENS Integration Tests > should properly resolve ENS name to address when no viem chain object is provided 29183ms
```

30s (with a 60s timeout) is a long time. Any ideas on possibility of reducing that?


The other seems fine and only takes ~10s so perhaps the timeout can just be 20s for it?
```
✓ Viem-Ethers Adapter ENS Integration Tests > ENS operations should fail when viem chain lacks ENS registry address 9609ms
```

Muhammad-Altabba: Not sure why it takes too long to resolve ens names. Most of the times it takes less than 15 seconds as the tests where passing early on this and when failing they worked after retry. On my machine it takes 1 or 2 seconds.

I just set it to 20s now. And in case if it fails the tests can be retried.

However, we can also skip this test completely now on the CI or make it with some other tests as manually-triggered in the future.

Muhammad-Altabba: The tests are failing now with timeout after reducing from 60s to 20s.
Not sure why the endpoints now takes long. And I do not see any harm in extending the time now to 60 seconds and create a task to make some of the integration tasks as manually-triggered.
Let me know if you like to keep the timeout as 20s and retry the run, extend it back to 60s, skip the test completely for now, or if you have another specific recommendation.

derekpierre: The tests passing would be good. 

Let's set the timeout back to 60s, but create a task to follow-up separately on why it needs 60s and ways of potentially improving that.

Muhammad-Altabba: https://github.com/nucypher/sprints/issues/270

PR #739, opened 2025-11-13
Nonce as `uint256`

derekpierre:
**NOTE: Before merging: update version of `nucypher-core` used - a new dev release will be needed**

**Type of PR:**

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

Depends on https://github.com/nucypher/nucypher-core/pull/126

**Issues fixed/closed:**

> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   },
   "dependencies": {
     "@changesets/cli": "^2.28.1",
-    "@nucypher/nucypher-core": "^0.15.1-dev.2",
+    "@nucypher/nucypher-core": "^0.15.1-dev.3",
     "ethers": "^5.8.0"
   },
   "devDependencies": {
@@ -65,7 +65,7 @@
       ]
     },
     "overrides": {
-      "@nucypher/nucypher-core": "^0.15.1-dev.2",
+      "@nucypher/nucypher-core": "^0.15.1-dev.3",
       "glob-parent@<5.1.2": ">=5.1.2",
       "node-forge@<1.0.0": ">=1.0.0",
       "node-forge@<1.3.0": ">=1.3.0",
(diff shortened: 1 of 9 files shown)


vzotova (approved):

package.json
```diff
@@ -25,7 +25,7 @@
   },
   "dependencies": {
     "@changesets/cli": "^2.28.1",
-    "@nucypher/nucypher-core": "^0.15.1-dev.2",
+    "@nucypher/nucypher-core": "link:../nucypher-core/nucypher-core-wasm-bundler",
```
theref: Why is it now a local dependency?

theref (approved):

package.json
derekpierre: That's how I can test the latest `nucypher-core` changes before a release. I will change that before merging - see note in PR description.

Also why I didn't want anyone else to merge it 😄

packages/shared/src/types.ts
```diff
@@ -49,7 +53,7 @@ export function toCoreUserOperation(
 ): UserOperation {
   const userOp = new UserOperation(
     userOperation.sender,
-    getBigIntValue(userOperation.nonce),
+    getUint256String(userOperation.nonce), // nonce as string to support big numbers
```
cygnusv, line 57: Why isn't `BigInt` an option here? I imagined the wasm binding layers would perform the conversion

derekpierre: Rust (including bindings) can only handle up to 128 bits and can't accommodate 256 bits. Hence the value has to be passed as a string if it is representative of 256 bits, and within Rust the string gets converted to a `U256` rust type which stores it appropriately.

cygnusv (approved):

PR #741, opened 2025-11-20
Signing Bug fixes

derekpierre:
**Type of PR:**

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:**

- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

Fixed issues encountered while doing real world testing.

- Sort signatures based on signer addresses returned before aggregating them
- Use `accountGasLimits` (proper name) instead of `accountGasLimit`
- Allow optional `signature` property on `PackedUserOperationToSign`
- Expose `PackedUserOperationToSign` and `UserOperationToSign` as part of `nucypher/taco` (so `nucypher/shared` isn't needed).

**Issues fixed/closed:**

> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/packages/taco/src/sign.ts
+++ b/packages/taco/src/sign.ts
@@ -58,13 +58,12 @@ function aggregateSignatures(
   //  and remove the '0x' prefix from each signature except the first one.
 
   // sort by signer address
-  const sortedSignatures = signatures
-    .slice()
-    .sort((a, b) =>{
-      const addrA = BigInt(a.signerAddress);
-      const addrB = BigInt(b.signerAddress);
-      return addrA < addrB ? -1 : addrA > addrB ? 1 : 0;
-    })
+  const sortedSignatures = [...signatures]
+    .sort((a, b) =>
+      a.signerAddress
+        .toLowerCase()
+        .localeCompare(b.signerAddress.toLowerCase()),
+    )
     .map((sig) => sig.signature);
 
   const thresholdSignatures = sortedSignatures.slice(0, threshold);
--- a/packages/taco/src/sign.ts
+++ b/packages/taco/src/sign.ts
@@ -51,20 +51,31 @@ export type SignResult = {
 };
 
 function aggregateSignatures(
-  signaturesByAddress: {
-    [checksumAddress: string]: TacoSignature;
-  },
+  signatures: TacoSignature[],
   threshold: number,
 ): string {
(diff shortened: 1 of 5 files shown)


packages/taco/src/sign.ts
```diff
@@ -51,20 +51,29 @@ export type SignResult = {
...
-    return signatures[0];
+  // Aggregate hex signatures by concatenating them; being careful to sort
+  //  and remove the '0x' prefix from each signature except the first one.
+
+  // sort by signer address
+  const sortedSignatures = signatures
+    .slice()
+    .sort((a, b) =>
+      BigInt(a.signerAddress) > BigInt(b.signerAddress) ? 1 : -1,
+    )
```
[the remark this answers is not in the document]
derekpierre: ✔️

packages/taco/src/sign.ts
```diff
@@ -51,20 +51,31 @@ export type SignResult = {
...
+  //  and remove the '0x' prefix from each signature except the first one.
+
+  // sort by signer address
+  const sortedSignatures = signatures
+    .slice()
+    .sort((a, b) =>{
+      const addrA = BigInt(a.signerAddress);
+      const addrB = BigInt(b.signerAddress);
+      return addrA < addrB ? -1 : addrA > addrB ? 1 : 0;
+    })
```
cygnusv: Can you just sort lexicographically but in, e.g. lowercase? In case it can make this code more legible. Alternatively, I think you don't need to explicitly return 1 or -1, but positive or negative; in that case, returning `addrA - addrB` is enough.

Not a strong comment, in any case.

vzotova (approved):

cygnusv (approved):

theref (approved):

packages/taco/src/sign.ts
derekpierre: I can do it lexicographically as lowercase.

The reason for not doing `addrA - addrB` is because that returns a `bigint`, and sort expects a `number`.

derekpierre: ✔️

derekpierre:
Now that `lynx` is updated with the `nonce` updates, the CI tests won't pass. I will merge this PR and rebase #739 over `signing-epic` and ensure that CI passes (and fix any issues) in that PR before merging.

PR #748, opened 2026-01-05
🐛 fix(taco): add jsonConditionSchema to anyConditionSchema

theref:
## Summary

The `json` condition type was missing from the `anyConditionSchema` union in `schemas/utils.ts`, causing validation failures when json conditions were used inside compound or sequential conditions.

## Changes

1. **Bug fix**: Added `jsonConditionSchema` import and included it in the `anyConditionSchema` union
2. **Regression test**: Added `any-condition-schema.test.ts` that validates all 14 condition types can be used inside compound conditions. This prevents future omissions by failing fast with a descriptive error message.

## Testing

- All 14 condition type tests pass
- Type checking passes

### DIFF
--- a/packages/taco/test/conditions/any-condition-schema.test.ts
+++ b/packages/taco/test/conditions/any-condition-schema.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, it } from 'vitest';
+
+import {
+  CompoundCondition,
+  compoundConditionSchema,
+  CompoundConditionType,
+} from '../../src/conditions/compound-condition';
+import {
+  testCompoundConditionObj,
+  testContextVariableConditionObj,
+  testContractConditionObj,
+  testECDSAConditionObj,
+  testIfThenElseConditionObj,
+  testJsonApiConditionObj,
+  testJsonConditionObj,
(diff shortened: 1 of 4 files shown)


derekpierre:
Nice catch.

Can you also re-run  the `generate-zod-docs`? I suspect it will be updated based on the update to the `anyCondtionSchema`.

packages/taco/test/conditions/any-condition-schema.test.ts
```diff
@@ -0,0 +1,80 @@
...
+ * This test ensures that all condition types are included in anyConditionSchema.
+ * If a new condition type is added but not included in anyConditionSchema,
+ * it will fail validation when used inside compound/sequential conditions.
+ *
+ * When adding a new condition type:
+ * 1. Add a test object in test-utils.ts
+ * 2. Add it to the test cases below
+ * 3. Ensure it's imported and added to anyConditionSchema in schemas/utils.ts
+ */
+describe('anyConditionSchema completeness', () => {
```
derekpierre, line 35: 👍

derekpierre (approved):
🎸

PR #750, opened 2026-01-05
fix(taco): exclude sequential varName references from required context params

theref:
## Summary

Fixes a bug where `findContextParameters()` incorrectly treated internal sequential condition `varName` references as external context parameters.

## Problem

When a sequential condition defines `varName: 'amountValue'` and later references `:amountValue` in a `returnValueTest.value`, the SDK was incorrectly treating `:amountValue` as an external parameter that must be provided by the caller.

## Solution

Modified `findContextParameters()` to:
1. Collect all `varName` declarations from sequential conditions using the existing `getAllNestedConditionVariableNames()` helper
2. Convert them to context param format (`:varName`)
3. Exclude these internal variables from the required external parameters

## Changes

- Export `getAllNestedConditionVariableNames` from `sequential.ts`
- Modify `findContextParameters()` to collect and exclude internal varNames
- Add comprehensive tests for sequential condition varName scoping
- Update max number of allowed conditions within `SequentialCondition` to 10 to match https://github.com/nucypher/nucypher/pull/3662

## Testing

- Added 4 new tests covering:
  - Basic internal varName exclusion
  - Mixed internal/external context params
  - Nested sequential conditions
  - Compound conditions containing sequential conditions
- All existing tests pass

PR #751, opened 2026-01-06
fix(taco): prevent varName from colliding with context parameter names

theref:
## Summary

Fixes a bug where an external context parameter with the same name as an internal `varName` in a sequential condition would be incorrectly excluded from `requestedContextParameters`.

## Problem

When a `varName` in a sequential condition has the same name as a context parameter, the previous fix would incorrectly exclude that external context param from `requestedContextParameters`, effectively making it invisible to the user.

## Solution

Add schema-level validation to reject sequential conditions where a `varName` has the same name as a context parameter used in the condition. This prevents the ambiguous situation entirely.

**Rule:** If you define `varName: 'foo'`, you cannot also use `:foo` as a context parameter - one must be renamed.

## Changes

- Add `findAllContextParams()` helper to recursively find all context parameters in a condition tree
- Add `noVarNameContextParamCollisions()` validation function  
- Add `.refine()` check to `sequentialConditionSchema` - https://zod.dev/api?id=refinements
- Add tests for collision rejection scenarios (returnValueTest, operations, nested sequential, compound conditions)
- Update `context.test.ts` to use non-colliding names

## BREAKING CHANGE

Sequential conditions with varNames that match context parameter names will now be rejected at schema validation time with the error:
> "Variable name cannot be the same as a context parameter name used in the condition"

### DIFF
--- a/packages/taco/src/conditions/schemas/sequential.ts
+++ b/packages/taco/src/conditions/schemas/sequential.ts
@@ -1,6 +1,7 @@
 import { z } from 'zod';
 
 import { ConditionProps } from '../condition';
+import { CONTEXT_PARAM_REGEXP } from '../const';
 import { maxNestedDepth } from '../multi-condition';
 
 import { baseConditionSchema, plainStringSchema } from './common';
@@ -48,6 +49,55 @@ const noDuplicateVarNames = (condition: ConditionProps): boolean => {
   return duplicates.length === 0;
 };
 
+/**
+ * Recursively finds all context parameters (`:paramName`) in a condition tree.
+ */
+const findAllContextParams = (value: unknown): Set<string> => {
+  const contextParams = new Set<string>();
+
+  if (!value) {
+    return contextParams;
+  }
+
+  if (typeof value === 'string') {
+    // Find all context param matches in the string
+    const matches = value.match(new RegExp(CONTEXT_PARAM_REGEXP.source, 'g'));
+    if (matches) {
+      matches.forEach((match) => contextParams.add(match));
+    }
+  } else if (Array.isArray(value)) {
+    value.forEach((item) => {
+      findAllContextParams(item).forEach((param) => contextParams.add(param));
+    });
+  } else if (typeof value === 'object') {
+    Object.values(value).forEach((entry) => {
+      findAllContextParams(entry).forEach((param) => contextParams.add(param));
+    });
+  }
+
+  return contextParams;
+};
+
+/**
+ * Validates that no varName collides with an external context parameter.
+ * If a varName 'foo' is defined, ':foo' cannot also be used as a context param.
+ */
+const noVarNameContextParamCollisions = (
+  condition: ConditionProps,
+): boolean => {
+  const allVarNames = getAllNestedConditionVariableNames(condition);
+  const allContextParams = findAllContextParams(condition);
+
+  // Check if any varName (with : prefix) appears in context params
+  for (const varName of allVarNames) {
+    if (allContextParams.has(`:${varName}`)) {
+      return false;
+    }
+  }
+
+  return true;
+};
+
 export const SequentialConditionType = 'sequential';
 
 export const conditionVariableSchema: z.ZodSchema = z.lazy(() =>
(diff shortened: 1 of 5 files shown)


packages/taco/src/conditions/schemas/sequential.ts
```diff
@@ -48,6 +49,55 @@ const noDuplicateVarNames = (condition: ConditionProps): boolean => {
...
+const noVarNameContextParamCollisions = (
+  condition: ConditionProps,
+): boolean => {
+  const allVarNames = getAllNestedConditionVariableNames(condition);
+  const allContextParams = findAllContextParams(condition);
+
+  // Check if any varName (with : prefix) appears in context params
+  for (const varName of allVarNames) {
+    if (allContextParams.has(`:${varName}`)) {
+      return false;
```
derekpierre, line 65: This seems like the opposite of what we want, no? This is basically saying that condition variables can't be used within the sequential condition - condition variables are intended to be referenced within a sequential condition as context variables. So, how are we differentiating between "external context parameters" and the legit use of condition variables?

For example, this simple modification to the test produces a failed test:
```
diff --git a/packages/taco/test/conditions/context.test.ts b/packages/taco/test/conditions/context.test.ts
index ef317a9a..cf47feaa 100644
--- a/packages/taco/test/conditions/context.test.ts
+++ b/packages/taco/test/conditions/context.test.ts
@@ -1510,7 +1510,7 @@ describe('sequential condition varName scoping', () => {
                   ...testTimeConditionObj,
                   returnValueTest: {
                     comparator: '>',
-                    value: ':nestedExternalParam', // External context param
+                    value: ':innerResult', // use of condition variable from above in subsequent condition
                   },
                 },
               },
@@ -1536,7 +1536,7 @@ describe('sequential condition varName scoping', () => {
     const conditionContext = new ConditionContext(condition);
 
     // External context params SHOULD be in requestedContextParameters
-    expect(conditionContext.requestedContextParameters).toContain(
+    expect(conditionContext.requestedContextParameters).not.toContain(
       ':nestedExternalParam',
     );
     expect(conditionContext.requestedContextParameters).toContain(
```

causes this line to fail:
```
    const condition = ConditionFactory.conditionFromProps(
      nestedSequentialCondition,
    );
```
i.e.
```
 FAIL   @nucypher/taco  test/conditions/context.test.ts > sequential condition varName scoping > handles nested sequential conditions with external context params
Error: Invalid condition: [{"code":"custom","message":"Variable name cannot be the same as a context parameter name used in the condition","path":["conditionVariables",1,"condition","conditionVariables"]},{"code":"custom","message":"Variable name cannot be the same as a context parameter name used in the condition","path":["conditionVariables"]}]
 ❯ new Condition src/conditions/condition.ts:23:13
     21|     const { data, error } = Condition.validate(schema, value);
     22|     if (error) {
     23|       throw new Error(ERR_INVALID_CONDITION(error));
       |             ^
     24|     }
     25|     this.value = data;
 ❯ new SequentialCondition src/conditions/sequential.ts:18:5
 ❯ Function.conditionFromProps src/conditions/condition-factory.ts:102:16
 ❯ test/conditions/context.test.ts:1533:40

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯


 Test Files  1 failed (1)
      Tests  1 failed | 104 passed (105)
   Start at  13:52:12
   Duration  1.67s (transform 258ms, setup 0ms, collect 743ms, tests 160ms, environment 0ms, prepare 48ms)

 ELIFECYCLE  Test failed. See above for more details.
```
-- 

I'll play around with the code from PR #750 a little.

derekpierre:
Subsumed by #750.

PR #752, opened 2026-01-13
feat(taco): increase condition limits and add exponent operator

theref:
## Summary

Increases flexibility for complex condition hierarchies and adds new operator functionality.

## Changes

- **Increase max nested depth from 2 to 4** for compound, if-then-else, and sequential conditions
- **Increase max sequential condition variables from 10 to 20**
- **Add `toTokenBaseUnits` operator** to variable operations for token decimal calculations

## Testing

- Updated existing depth validation tests to use 5 levels of nesting (to exceed the new max of 4)
- Added new tests for exponent operator with numeric values, context parameters, and missing value validation

### DIFF
--- a/packages/taco/integration-test/condition-lingo.test.ts
+++ b/packages/taco/integration-test/condition-lingo.test.ts
@@ -65,8 +65,9 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
   () => {
     test('validate condition lingo with lynx node to confirm consistency', async () => {
       // Note: there are limits to conditions
-      // - max 5 operands in a multi-condition (compound, sequential)
-      // - max 2 nested levels of multi-conditions (compound, sequential, if-then-else)
+      // - max 5 operands in a compound condition
+      // - max 20 condition variables in a sequential condition
+      // - max 4 nested levels of multi-conditions (compound, sequential, if-then-else)
       const overallCondition = new SequentialCondition({
         conditionVariables: [
           {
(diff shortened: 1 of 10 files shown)


packages/taco/src/conditions/schemas/variable-operation.ts
```diff
@@ -8,6 +8,7 @@ export const OPERATOR_FUNCTIONS = [
   '*=',
   '/=',
   '%=',
+  '*pow=',
```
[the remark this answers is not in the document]
derekpierre: Perhaps an old decryption. Either way, `*pow=` is what we want.

derekpierre: There's only so much we can do in validating operations, but we may want to add a specific refinement validation for this operation since it specifically expects a value that's a list of two items.

wdyt?

packages/taco/test/conditions/variable-operation.test.ts
```diff
@@ -49,3 +49,41 @@ describe('validates schema', () => {
     },
   );
 });
+
+describe('toTokenBaseUnits operator', () => {
```
derekpierre, line 53: Nice tests! 🎸

derekpierre (commented):
Changes look great.

A couple things:
1) We need some additional clean up in the integration tests given that the limits have now changed:
- https://github.com/nucypher/taco-web/blob/signing-epic/packages/taco/integration-test/condition-lingo.test.ts#L67-L69
- https://github.com/nucypher/taco-web/blob/signing-epic/packages/taco/integration-test/condition-lingo.test.ts#L146-L150 (increase chunk size)

2) I think (?) you need to rerun the `generate-zod-docs` script since the limits have changed. I believe they are documented in the resulting conditions-schema.md file.

packages/taco/src/conditions/schemas/variable-operation.ts
derekpierre: My comment is no longer applicable since we shifted our strategy.

packages/taco/integration-test/condition-lingo.test.ts
```diff
@@ -143,11 +144,9 @@ describe.skipIf(!process.env.RUNNING_IN_CI)(
...
       };
 
-      // Value used for chunkSize is to ensure:
-      // 1. we don't exceed the max 5 operands in a multi-condition
-      // 2. we adhere to the fact that "and" compound conditions must have at least 2 operands
-      // 3. SUPPORTED_ECDSA_CURVES.length is currently 6, so we can split it into chunks of 3
-      const chunkSize = 3;
+      // chunkSize ensures we don't exceed max 5 operands in a compound condition
+      // and that "and" compound conditions have at least 2 operands
+      const chunkSize = 5;
```
derekpierre: My bad with my comment about this.

We need `SUPPORTED_ECDSA_CURVES.length % chunk_size != 1` (because `and` compound conditions need length of operands > 1)

So, chunk size can remain 3 (or it can be 4).

Also, let's leave the 3) comment from before, so that others are aware of SUPPORTED_ECDSA_CURVES.length.

theref: not sure i understand

derekpierre: The CI job is failing because using a `chunkSize` of 5 causes the first iteration to have 5 operands (fine), but the second iteration to have 1 operand (not fine because an "and" compound condition needs at least 2 operands).

Given that `SUPPORTED_ECDSA_CURVES.length` is 6 we can use chunk sizes of (2 or 3 or 4) but not 5. We can leave the original chunk size of 3 for this test.

derekpierre (approved):
🎸

PR #753, opened 2026-01-14
feat(taco): Replace indexWithinTuple with unified subIndices for ABI validation

theref:
## Summary

Replace `indexWithinTuple` with a unified `subIndices` field for navigating nested ABI parameter structures (arrays and tuples) in signing conditions.

Related change on `nucypher` - https://github.com/nucypher/nucypher/pull/3685.

## Changes

- Replace `indexWithinTuple` (single index) with `subIndices` (array of indices) for sequential navigation through nested types
- Array types consume an index to access elements; tuple types consume an index to access fields
- Add `resolveTypeWithSubIndices` helper for type-safe navigation with clear error messages
- Update nested ABI validation to work with the resolved final type

## Testing

All existing tests updated + new tests for arrays, arrays of tuples, tuples of arrays, and nested ABI validation with subIndices.


theref:
CI/CD will fail until https://github.com/nucypher/nucypher/pull/3685 is merged and deployed onto lynx nodes :)

vzotova (approved):

derekpierre:
Given the update to the direction of the PR, let's update the title and description.

packages/taco/src/conditions/schemas/signing.ts
```diff
@@ -152,6 +152,57 @@ const humanAbiCallSignatureSchema = z
...
+      currentType = currentType.arrayChildren;
+    } else if (currentType.baseType === 'tuple') {
+      // Tuple type - navigate to component at index
+      if (idx >= currentType.components.length) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: `Index ${idx} at subIndices position ${i} is out of range for tuple with ${currentType.components.length} fields`,
+          path: ['allowedAbiCalls', signature, validationIndex, 'subIndices'],
+        });
+        return null;
```
derekpierre: Since this is Typescript, is `undefined` preferred here (and below)?

derekpierre (commented):
Let's also update the title and description of the PR.

packages/taco/src/conditions/schemas/signing.ts
theref: i don't know, typescript is not my strong point. happy to change it

derekpierre: I'm no typescript expert either, but I believe that `undefined` is generally preferred.

derekpierre (approved):
🎸

PR #758, opened 2026-02-04
feat(taco): Add timeout parameter to signUserOp

theref:
Add optional timeout parameter to control how long Porter waits for nodes to respond during signing operations. This allows callers to increase the timeout beyond Porter's default 30 seconds when needed.

This functionality is already exposed by Porters `/sign` endpoint, it was just never brought over to `taco-web`

### DIFF
--- a/packages/taco/src/sign.ts
+++ b/packages/taco/src/sign.ts
@@ -151,6 +151,7 @@ async function makeSigningRequests(
  * @param aaVersion - The AA version of the account abstraction to use for signing.
  * @param context - Optional condition context for the context variable resolution.
  * @param porterUris - Optional URIs for the Porter service. If not provided, will fetch the default URIs from the domain.
+ * @param timeout - Optional timeout in seconds for the Porter signing request.
  * @returns A promise that resolves to a SignResult containing the message hash, aggregated signature, and signing results from the Porter service.
  * @throws An error if the signing process fails due to insufficient signatures or mismatched hashes.
  */
@@ -163,6 +164,7 @@ export async function signUserOp(
   aaVersion: 'mdt' | '0.8.0' | string,
   context?: ConditionContext,
   porterUris?: string[],
+  timeout?: number,
 ): Promise<SignResult> {
   const porterUrisFull: string[] = porterUris
     ? porterUris
@@ -194,6 +196,7 @@ export async function signUserOp(
   const { encryptedResponses, errors } = await porter.signUserOp(
     encryptedRequests,
     threshold,
+    timeout,
   );
   if (Object.keys(encryptedResponses).length < threshold) {
     // not enough signatures returned
--- a/packages/shared/src/porter.ts
+++ b/packages/shared/src/porter.ts
@@ -165,6 +165,7 @@ type PostTacoSignRequest = {
     Base64EncodedBytes
   >;
   readonly threshold: number;
+  readonly timeout?: number;
 };
 
 type TacoSignResponse = {
@@ -339,6 +340,7 @@ export class PorterClient {
       EncryptedThresholdSignatureRequest
     >,
     threshold: number,
+    timeout?: number,
   ): Promise<TacoSignResult> {
     const data: PostTacoSignRequest = {
       encrypted_signing_requests: Object.fromEntries(
@@ -350,6 +352,7 @@ export class PorterClient {
         ),
       ),
       threshold,
+      ...(timeout !== undefined && { timeout }),
     };
 
     const resp: AxiosResponse<TacoSignResponse> = await this.tryAndCall({
--- a/packages/taco/test/taco-sign.test.ts
+++ b/packages/taco/test/taco-sign.test.ts
@@ -404,6 +404,7 @@ describe('TACo Signing', () => {
             '0xnode2': expect.any(EncryptedThresholdSignatureRequest),
           },
           threshold,
+          undefined,
         );
 
         const call = porterSignUserOpMock.mock.calls.at(-1)!;
@@ -482,6 +483,7 @@ describe('TACo Signing', () => {
           '0xnode2': expect.any(EncryptedThresholdSignatureRequest),
         },
         threshold,
+        undefined,
       );
       const call = porterSignUserOpMock.mock.calls.at(-1)!;
       const [op] = call;


packages/taco/src/sign.ts
```diff
@@ -163,6 +164,7 @@ export async function signUserOp(
   aaVersion: 'mdt' | '0.8.0' | string,
   context?: ConditionContext,
   porterUris?: string[],
+  timeout?: number,
```
derekpierre, line 167: I’m concerned that this value may not behave the way it claims to.

In practice, there is already an upper bound enforced by Porter itself (see https://github.com/nucypher/nucypher-porter/blob/development/porter/main.py#L95
). If a requested timeout exceeds Porter’s max, it is silently capped. This limit exists to prevent clients from overwhelming Porter with long-running requests, and the max value is not exposed programmatically.

As a result, behavior can be inconsistent from the caller’s perspective. For example, Porter’s current maximum is 30s, and any request specifying a timeout greater than that will still be handled with a 30s timeout. A user-provided value of 60s would therefore not be respected.

We could only allow this value to be at most 30s on the client side to match Porter’s default, but that default itself is configurable via an environment variable (even though I don’t expect operators to change it in practice).

That's one option, OR

I wonder if, at the moment, it makes sense to hold off on providing this option since it may cause confusion.

Thoughts?

theref: that's fine, but it is the first bottleneck we hit when sending many requests, and it doesn't actually free up any compute because the request isn't cancelled node side

derekpierre: Could you elaborate a bit.

So you mean, in general, the use of this timeout is to potentially go lower than the 30s? i.e. right now 30s is the default, and we want the option to pass a timeout lower than 30s so that resources get freed faster by Porter and the node?

derekpierre: >  it doesn't actually free up any compute because the request isn't cancelled node side

Yes. So increasing the timeout above 30s can also cause slow downs as well because the request lingers for longer.

theref: no i mean that at the moment, it's very easy for us to get in a state where requests are taking longer than 30s. And so when I do some stress testing, requests fail due to this timeout even though they're actually being completed.

The problem seems to be that the timeout is returned, but the resources aren't actually released on the nodes. Otherwise we would see a behavior where the nodes recover due to a load of requests getting cancelled and them effectively catching up in the queue

derekpierre: If I'm understanding correctly, that's two different things: Porter timing out is not the same as the node stopping handling of the request after that timeout.


What Porter Does
=============
Porter's timeout is enacted separately from the nodes. Porter timing out only bounds how long the requester waits for a result; it does not inherently bound node work.

Porter enforces its own timeout (requested or default, capped to a max). It issues concurrent requests to nodes via a WorkerPool, and if it can’t collect a threshold of responses before that timeout, it returns a timeout to the requester.

This bounds how long the requester waits, but it doesn’t guarantee nodes stop working once Porter returns.
   
What the Node Does
===============
On the node side, we don’t currently enforce an application-level timeout for signing (e.g., “stop processing after X seconds from a request parameter”). So one option would be to add explicit timeout and cancellation  functionality in the node's web server endpoint eg. `/sign` (and decrypt as well). 

In our Flask/Hendrix web server setup, if the client-side closes the HTTPS connection (e.g. Porter returns early), I don't believe the node’s request-handling work necessarily stops for that request. Which is likely why you see that lack of catching up by the node or continuing to process after timeouts.

Or maybe Hendrix already has functionality for managing these timeouts and stopping the thread that we should leverage. (cc @KPrasch )

-- 

Let me know if I still missed your angle and we can do voice.

theref: completely agree with what you're saying - that's my understanding too. But I think the current setup leads to increased waiting.

assumption: nodes are busy, requests might take a while

1. i send a requests
2. after 30s, porter sends a timeout (but request is still processed by nodes)
3. I receive the timeout and send another request
4. original request is eventually filled (and therefore signatures exist!), but nothing returned because porter has dropped it
5. 2nd request also times out on porter

I'm starting to think it's actually 'dangerous' to have a porter timeout that doesn't cancel the work on the nodes. We had one instance of a double spend during the tipping tournament and it was almost certainly because of something like this.

It also explains why the tippers were getting errors and the nodes seemed absolutely fine. The system would 'go down' for 20 mins, but from our point of view everything seemed ok. It was only once they stopped tipping that the nodes had time to get through the backlog.

I think we should either:
- cancel the work on the nodes when a request times out, 
or
- don't timeout at all, the end user just has to wait until their request is processed. Otherwise they will just immediately retry when the timeout is seen

This PR was actually so that I could _increase_ the timeout that was applied to requests. I agree that it's confusing that there is a ENV var on porter that could overwrite it though...that tripped me up for a while the other day.

derekpierre: > cancel the work on the nodes when a request times out,

^ Yep. I think this yields the bigger benefit. That would be `nucypher/nucypher` code.

derekpierre:
Can we close this, or is there still some reason to allow the timeout parameter from `taco-web`?

theref:
> Can we close this, or is there still some reason to allow the timeout parameter from `taco-web`?

not sure - i'm open to both. it was definitely useful in testing in order to figure out where the actual delays were happening. Just timing out hid everything

derekpierre:
> not sure - i'm open to both. it was definitely useful in testing in order to figure out where the actual delays were happening. Just timing out hid everything

I think introducing a configurable timeout is potentially more confusing than helpful, especially given Porter already enforces a cap. We also don’t actually want requests running longer than 30 seconds, since anything beyond that just hogs resources. Even that 30s may be too high... 😅 . In practice, the only viable option for applications would be to set a timeout below 30s, not above it (see: https://github.com/nucypher/taco-web/pull/758#discussion_r2774414707
).

Now that we understand what the underlying issue was, my vote is that we should hold off. While it was useful for testing, it ultimately wasn’t the root cause, and this would be a public API change that’s difficult to walk back once introduced.

We can always add it later if a real need emerges. But even then, the practical limit would still be ≤30s, not greater, which somewhat limits its value.

Side note: are any increased Porter timeouts still configured? At one point there was a 120s timeout on the load balancer? If so, we should revert those; both at the load balancer and Porter instance level.

theref:
then i'm in agreement that we should close this :) and yeah, i think the lynx porter is still configured to 120

PR #759, opened 2026-02-05
feat(taco): Add create2 operation for local address computation

theref:
Add a new create2 operation to the variable operation system that computes Ethereum CREATE2 addresses locally, eliminating RPC calls to factory contracts for counterfactual address resolution.

The operation takes an object value with deployerAddress and bytecodeHash fields, both supporting context variable resolution. This enables full Account Abstraction address derivation pipelines.

Implements taco-web equivalent of nucypher/nucypher#3703.

### DIFF
--- a/packages/taco/schema-docs/condition-schemas.md
+++ b/packages/taco/schema-docs/condition-schemas.md
@@ -381,14 +381,10 @@ _(\*) Required._
 
 An operation that can be performed on an obtained result.
 
-_Object containing the following properties:_
+_Union of the following possible types:_
 
-| Property             | Type                                                                                                                                                                                                                                                                                                                                                                                                                                      |
-| :------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **`operation`** (\*) | `'+=' \| '-=' \| '*=' \| '/=' \| '%=' \| 'toTokenBaseUnits' \| 'index' \| 'round' \| 'abs' \| 'avg' \| 'ceil' \| 'ethToWei' \| 'floor' \| 'len' \| 'max' \| 'min' \| 'sum' \| 'weiToEth' \| 'bool' \| 'float' \| ...`                                                                                                                                                                                                                     |
-| `value`              | [ParamOrContextParam](#paramorcontextparam) _or_ _Object with properties:_<ul><li>`deployerAddress`: `string` (_regex: `/^0x[0-9a-fA-F]+$/`_) _or_ [ContextParam](#contextparam)</li><li>`bytecodeHash`: `string` (_regex: `/^0x[0-9a-fA-F]+$/`_) _or_ [ContextParam](#contextparam)</li></ul> Description: Value for create2 operation containing deployerAddress and bytecodeHash for computing CREATE2 addresses locally.<br /> <br /> |
-
-_(\*) Required._
+- _Object with properties:_<ul><li>`operation`: `'+=' | '-=' | '*=' | '/=' | '%=' | 'toTokenBaseUnits' | 'index' | 'round' | 'abs' | 'avg' | 'ceil' | 'ethToWei' | 'floor' | 'len' | 'max' | 'min' | 'sum' | 'weiToEth' | 'bool' | 'float' | ...`</li><li>`value`: [ParamOrContextParam](#paramorcontextparam)</li></ul>
+- _Object with properties:_<ul><li>`operation`: `'create2'`</li><li>`value`: _Object with properties:_<ul><li>`deployerAddress`: `string` (_regex: `/^0x[0-9a-fA-F]+$/`_) _or_ [ContextParam](#contextparam)</li><li>`bytecodeHash`: `string` (_regex: `/^0x[0-9a-fA-F]+$/`_) _or_ [ContextParam](#contextparam)</li></ul> - Value for create2 operation containing deployerAddress and bytecodeHash for computing CREATE2 addresses locally.</li></ul>
 
 ## VariableOperationsArray
 
--- a/packages/taco/schema-docs/condition-schemas.md
+++ b/packages/taco/schema-docs/condition-schemas.md
@@ -383,10 +383,10 @@ An operation that can be performed on an obtained result.
 
 _Object containing the following properties:_
 
-| Property             | Type                                                                                                                                                                                                                  |
-| :------------------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **`operation`** (\*) | `'+=' \| '-=' \| '*=' \| '/=' \| '%=' \| 'toTokenBaseUnits' \| 'index' \| 'round' \| 'abs' \| 'avg' \| 'ceil' \| 'ethToWei' \| 'floor' \| 'len' \| 'max' \| 'min' \| 'sum' \| 'weiToEth' \| 'bool' \| 'float' \| ...` |
-| `value`              | [ParamOrContextParam](#paramorcontextparam)                                                                                                                                                                           |
+| Property             | Type                                                                                                                                                                                                                                                                                                                                                                                                                                      |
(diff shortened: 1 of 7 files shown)


packages/taco/src/conditions/schemas/context.ts
```diff
@@ -25,6 +25,16 @@ const paramSchema = z.union([
   z.bigint(),
 ]);
 
+// Schema for CREATE2 operation value - supports both literal addresses/hashes and context parameters
+export const create2ValueSchema = z
+  .object({
+    deployerAddress: z.union([plainStringSchema, contextParamSchema]),
+    bytecodeHash: z.union([plainStringSchema, contextParamSchema]),
```
derekpierre: Instead of `plainStringSchema`, do we expect these values to be hex?

i.e. 0x-hex string or context parameter?

packages/taco/src/conditions/schemas/context.ts
```diff
@@ -25,6 +25,16 @@ const paramSchema = z.union([
   z.bigint(),
 ]);
 
+// Schema for CREATE2 operation value - supports both literal addresses/hashes and context parameters
+export const create2ValueSchema = z
```
derekpierre: Should this be defined in variable-operation.ts? How come in context.ts?

packages/taco/src/conditions/schemas/context.ts
```diff
@@ -40,6 +50,7 @@ export const paramOrContextParamSchema: z.ZodSchema = z.union([
   paramSchema,
   contextParamSchema,
   z.lazy(() => z.array(paramOrContextParamSchema)),
+  create2ValueSchema,
```
derekpierre: Should this be here, as opposed to this being part of the variable operation schema?

derekpierre (changes requested):

packages/taco/src/conditions/schemas/context.ts
theref: Fixed. Changed to use a 0x-prefixed hex string schema instead of plainStringSchema, so both deployerAddress and bytecodeHash now validate as 0x-prefixed hex or context parameter.

packages/taco/src/conditions/schemas/context.ts
theref: Removed from the union

packages/taco/src/conditions/schemas/variable-operation.ts
```diff
@@ -98,6 +113,34 @@ export const variableOperationSchema = z
...
+  )
+  .refine(
+    (data) => {
+      if (
+        data.operation !== 'create2' &&
+        create2ValueSchema.safeParse(data.value).success
+      ) {
+        return false;
+      }
+      return true;
```
cygnusv: I don't see what's the purpose of this rule. Is it necessary to restrict non-create2 operators from using the same value schema?

cygnusv (commented):
Looks good! One comment for the moment

packages/taco/src/conditions/schemas/variable-operation.ts
theref: nice catch! it's because the schema used to be in `paramOrContextParamSchema`, but that's been changed so this can be removed now

derekpierre: > Is it necessary to restrict non-create2 operators from using the same value schema?

It is. We are trying to prevent users from defining variable operations that don't make sense. So something like:
```typescript
{
   operation: "toTokenBaseUnits",
   value: {
       "deployerAddress": "0x...."
       "bytecodeHash": "0x..."
   }
}
```
should not be allowed.

@theref , am I correct in believing that this is currently allowed?

derekpierre:
Reminder that you probably need to re-run the `generate-zod-docs` script

packages/taco/test/conditions/variable-operation.test.ts
```diff
@@ -32,7 +36,11 @@ describe('validates schema', () => {
   );
 
   it.each(
-    OPERATOR_FUNCTIONS.filter((op) => !UNARY_OPERATOR_FUNCTIONS.includes(op)),
+    OPERATOR_FUNCTIONS.filter(
+      (op) =>
+        !UNARY_OPERATOR_FUNCTIONS.includes(op) &&
+        !OBJECT_VALUE_OPERATIONS.includes(op),
```
derekpierre: I'm a bit hesitant to do this; the reason being that this is how we ensure all operators are indeed always tested. Is there a way to make this test work without needing the `OBJECT_VALUE_OPERATIONS` exception list?

This is basically a regression backstop.

packages/taco/src/conditions/schemas/variable-operation.ts
theref: just testing

theref: You're right. I removed it thinking it was unnecessary since `create2ValueSchema` was taken out of the shared `paramOrContextParamSchema` union in `context.ts`. But I missed that I'd added it back as part of a union directly on the `value` field in `variable-operation.ts`, which has the same effect

packages/taco/test/conditions/variable-operation.test.ts
theref: no problem, added it back in with the assertion `toContain('Value must be defined for operation')`

packages/taco/test/conditions/variable-operation.test.ts
```diff
@@ -87,3 +95,93 @@ describe('toTokenBaseUnits operator', () => {
...
+  });
+
+  it('rejects create2 operator with primitive value', () => {
+    const result = variableOperationSchema.safeParse({
+      operation: 'create2',
+      value: 5,
+    });
+    expect(result.success).toBe(false);
+  });
+});
```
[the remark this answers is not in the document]
derekpierre: ^ one additional test here, as pointed out above, could be good. Something like what we were chatting about - https://github.com/nucypher/taco-web/pull/759/changes#r2788390855

packages/taco/schema-docs/condition-schemas.md
```diff
@@ -383,10 +383,10 @@ An operation that can be performed on an obtained result.
...
-| :------------------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **`operation`** (\*) | `'+=' \| '-=' \| '*=' \| '/=' \| '%=' \| 'toTokenBaseUnits' \| 'index' \| 'round' \| 'abs' \| 'avg' \| 'ceil' \| 'ethToWei' \| 'floor' \| 'len' \| 'max' \| 'min' \| 'sum' \| 'weiToEth' \| 'bool' \| 'float' \| ...` |
-| `value`              | [ParamOrContextParam](#paramorcontextparam)                                                                                                                                                                           |
+| Property             | Type                                                                                                                                                                                                                                                                                                                                                                                                                                      |
```
derekpierre: This reads a bit odd

Image

packages/taco/src/conditions/schemas/variable-operation.ts
```diff
@@ -66,7 +81,7 @@ export const UNARY_OPERATOR_FUNCTIONS = [
 export const variableOperationSchema = z
   .object({
     operation: z.enum(OPERATOR_FUNCTIONS),
-    value: paramOrContextParamSchema.optional(),
+    value: z.union([paramOrContextParamSchema, create2ValueSchema]).optional(),
```
derekpierre: I'm wondering if a different separation may "simplify" this setup. 

Quick musing - https://github.com/derekpierre/taco-web/commit/dfcb340a6dd33b71f39441ef2801dc874730589a

You can give it some thought and run it by CoPilot. It also helps with the odd condition schema being produced by the `generate-zod-docs` script.

theref: i added in your commit :)

packages/taco/test/conditions/variable-operation.test.ts
```diff
@@ -177,4 +177,23 @@ describe('create2 operator', () => {
     });
     expect(result.success).toBe(false);
   });
+
+  it.each(
```
derekpierre, line 181: 👍

packages/taco/schema-docs/condition-schemas.md
derekpierre: 👍 , this looks better:

Screenshot 2026-02-11 at 11 47 41 AM

derekpierre (approved):
🎸

PR #762, opened 2026-02-12
Add in-memory cache to SigningCoordinatorAgent

theref:
Cache `getParticipants` and `getThreshold` with a 10s TTL to reduce redundant contract reads.

### DIFF
--- a/packages/shared/src/cache.ts
+++ b/packages/shared/src/cache.ts
@@ -1,0 +1,25 @@
+export class TtlCache {
+  private readonly ttlMs: number;
+  private readonly store = new Map<string, { data: unknown; expiry: number }>();
+
+  constructor(ttlMs: number) {
+    this.ttlMs = ttlMs;
+  }
+
+  get<T>(key: string): T | undefined {
+    const entry = this.store.get(key);
+    if (entry && Date.now() < entry.expiry) {
+      return entry.data as T;
+    }
+    this.store.delete(key);
+    return undefined;
+  }
+
+  set(key: string, data: unknown): void {
+    this.store.set(key, { data, expiry: Date.now() + this.ttlMs });
+  }
+
+  clear(): void {
+    this.store.clear();
+  }
+}
--- a/packages/shared/src/contracts/agents/signing-coordinator.ts
+++ b/packages/shared/src/contracts/agents/signing-coordinator.ts
@@ -14,15 +14,39 @@ export type SignerInfo = {
 };
 
 export class SigningCoordinatorAgent {
+  private static readonly CACHE_TTL_MS = 10_000;
+  private static cache = new Map<string, { data: unknown; expiry: number }>();
+
+  private static getCached<T>(key: string): T | undefined {
+    const entry = this.cache.get(key);
+    if (entry && Date.now() < entry.expiry) {
+      return entry.data as T;
+    }
+    this.cache.delete(key);
+    return undefined;
+  }
+
+  private static setCache(key: string, data: unknown): void {
+    this.cache.set(key, { data, expiry: Date.now() + this.CACHE_TTL_MS });
+  }
+
+  public static clearCache(): void {
+    this.cache.clear();
+  }
+
   public static async getParticipants(
     provider: ethers.providers.Provider,
     domain: Domain,
     cohortId: number,
   ): Promise<SignerInfo[]> {
+    const cacheKey = `participants:${domain}:${cohortId}`;
+    const cached = this.getCached<SignerInfo[]>(cacheKey);
+    if (cached !== undefined) return cached;
+
     const coordinator = await this.connectReadOnly(provider, domain);
     const participants = await coordinator.getSigners(cohortId);
 
-    return participants.map(
+    const result = participants.map(
       (
         participant: SigningCoordinator.SigningCohortParticipantStructOutput,
       ) => {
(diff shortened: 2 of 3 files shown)


packages/shared/src/contracts/agents/signing-coordinator.ts
```diff
@@ -14,15 +14,39 @@ export type SignerInfo = {
...
+    return undefined;
+  }
+
+  private static setCache(key: string, data: unknown): void {
+    this.cache.set(key, { data, expiry: Date.now() + this.CACHE_TTL_MS });
+  }
+
+  public static clearCache(): void {
+    this.cache.clear();
+  }
```
cygnusv: This should be moved to a utility module or similar

packages/shared/src/contracts/agents/signing-coordinator.ts
```diff
@@ -14,15 +14,39 @@ export type SignerInfo = {
 };
 
 export class SigningCoordinatorAgent {
+  private static readonly CACHE_TTL_MS = 10_000;
```
cygnusv: I think a higher TTL is still safe. Even something like 60 seconds should be fine.

packages/shared/test/signing-coordinator-cache.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    mockProvider.getNetwork.mockResolvedValue({ chainId: 11155111 });
+
+    const signerKey = SessionStaticSecret.random();
+    mockContract = {
+      getSigners: vi.fn().mockResolvedValue([
+        {
+          provider: '0xnode1',
+          signerAddress: '0x0000000000000000000000000000000000000001',
+          signingRequestStaticKey: `0x${Buffer.from(signerKey.publicKey().toBytes()).toString('hex')}`,
```
cygnusv: This could be a fake string like `0xfoobar`, right? No need to actually create a `SessionStaticSecret`.

cygnusv (approved):
Awesome! Just minor comments

packages/shared/src/contracts/agents/signing-coordinator.ts
theref: :rocket:

packages/shared/test/signing-coordinator-cache.test.ts
theref: `getParticipants` calls `SessionStaticKey.fromBytes(fromHexString(...))` so i'd need to mock stuff - I'd prefer to just test on a real key

cygnusv: Fair enough!

packages/shared/src/contracts/agents/signing-coordinator.ts
```diff
@@ -14,15 +15,25 @@ export type SignerInfo = {
...
+  public static clearCache(): void {
+    this.cache.clear();
+  }
+
   public static async getParticipants(
     provider: ethers.providers.Provider,
     domain: Domain,
     cohortId: number,
   ): Promise<SignerInfo[]> {
+    const cacheKey = `participants:${domain}:${cohortId}`;
```
derekpierre: Something to consider: changing the order could make it easier to clear all cached data for a specific ID in the future (even if we don’t need that right now).

For example, if the cache key were structured as `<domain>:<cohortId>:<participants>`, you could more easily invalidate all entries for a given `<domain>`, or all entries for a specific `<cohortId>` within that domain, by targeting the key prefix. Using the identifier as a prefix generally makes this kind of invalidation simpler than placing it at the end.

To be clear, the current structure doesn’t prevent this and can still work. It’s just that organizing the key from broader to more specific values may simplify the logic.

Same comment for `threshold` caching below.

It might also make sense to put cache key construction a helper method to keep whatever specific structure we end up using (prefix or suffix) consistent.

packages/shared/src/cache.ts
```diff
@@ -0,0 +1,25 @@
+export class TtlCache {
```
derekpierre, line 1: One learning from the TTL cache on the `nucypher` side is that entries don't get periodically cleared from the cache if not specifically requested after expiry. That being said, just confirming that it is safe to assume that this is not a big problem and in the vast number of cases, client side usages by apps will likely just be one cohort anyway...?

KPrasch (approved):
Looking good, this is the single biggest taco-web side optimization on the signing hot path.  My tests show this change alone can remove up to a full second of network latency on a warm cache.  Agree with @cygnusv that the TTL can be long.  Even say, 600 seconds in my opinion. ✅

packages/shared/src/contracts/agents/signing-coordinator.ts
theref: changed :)

packages/shared/src/cache.ts
theref: yeah, i think managing multiple cohorts from one taco instance would actually be a pain

packages/shared/test/signing-coordinator-cache.test.ts
derekpierre: There is a `toHexString(...)` utility function from `@nucypher/shared` that you can use.

So this would simplify to something like `0x${toHexString(signerKey.publicKey().toBytes())}`

derekpierre (approved):
🎸  - minor comment about using `toHexString(...)` utility.

PR #772, opened 2026-04-09
feat(taco): Generate JSON Schema alongside markdown schema docs

theref:
## Summary

Adds a JSON Schema export to the existing `schema-docs-generation` script. The script already produces a human-readable markdown reference (`condition-schemas.md`); this PR makes it also emit `condition-schema.json` — a standard [JSON Schema](https://json-schema.org/) document describing the full `ConditionExpression` envelope (`{version, condition}`).

## Why

Conditions are JSON, and integrators increasingly author them with LLM assistance. A JSON Schema unlocks:

- **Editors:** any `conditions.json` with `"$schema": "https://.../condition-schema.json"` gets inline validation and autocomplete in VS Code, JetBrains, Cursor, etc.
- **LLMs:** structured-output APIs (OpenAI, Anthropic) accept JSON Schema directly and produce strictly valid conditions.
- **CI / non-TS users:** anyone can validate conditions with a standard JSON Schema validator (`ajv`, `jsonschema`, etc.) without depending on `@nucypher/taco`.

## How

- `zod-to-json-schema` added as a dev dependency.
- The lazy union root (`anyConditionSchema`) is wrapped in the `ConditionExpression` shape so the JSON Schema validates exactly the same envelope that `ConditionExpression.fromObj` parses at runtime.
- BigInt bounds (uint256 ranges on `BlockchainParamOrContextParam`) are stripped during serialization because standard JSON Schema requires numeric bounds. The `integer` type constraint still applies — only the explicit min/max disappears.

## Tested with

```bash
pnpm --filter @nucypher/taco generate-zod-docs
pnpm dlx ajv-cli validate -s packages/taco/schema-docs/condition-schema.json \
  -d /tmp/test-condition.json --strict=false
```

A valid `TimeCondition` envelope passes; a `TimeCondition` with `"comparator": "EQUALS"` is rejected with the expected `anyOf` mismatch.

## Notes

- The markdown output also picks up a small unrelated description-text refresh from regenerating against the current schemas (the `subIndices` description on `AbiParameterValidation`). Happy to split that into a separate commit if preferred.
- Output filename is `condition-schema.json` (singular) to mirror the singular envelope it describes, alongside the plural `condition-schemas.md` reference.

## Test plan

- [ ] CI passes
- [ ] `pnpm --filter @nucypher/taco generate-zod-docs` runs cleanly
- [ ] Generated `condition-schema.json` validates a known-good condition via `ajv`
- [ ] Generated `condition-schema.json` rejects a known-bad condition

### DIFF
--- a/packages/taco/package.json
+++ b/packages/taco/package.json
@@ -55,7 +55,8 @@
     "axios": "^1.8.4",
     "dotenv": "^16.5.0",
     "glob": "^11.0.1",
-    "modified-zod2md": "0.1.5-modified.4"
+    "modified-zod2md": "0.1.5-modified.4",
+    "zod-to-json-schema": "^3.24.5"
   },
   "engines": {
     "node": ">=18",
--- a/packages/taco/schema-docs/condition-schema.json
+++ b/packages/taco/schema-docs/condition-schema.json
@@ -1,0 +1,1087 @@
+{
+  "$ref": "#/definitions/ConditionExpression",
+  "definitions": {
+    "ConditionExpression": {
+      "type": "object",
+      "properties": {
+        "version": {
+          "type": "string",
+          "const": "1.0.0"
+        },
+        "condition": {
+          "anyOf": [
+            {
+              "type": "object",
+              "properties": {
+                "conditionType": {
+                  "type": "string",
+                  "const": "rpc",
+                  "default": "rpc"
+                },
+                "chain": {
+                  "type": "integer",
+                  "minimum": 0
+                },
+                "method": {
+                  "type": "string",
+                  "enum": [
+                    "eth_getBalance"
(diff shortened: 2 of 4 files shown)


derekpierre (approved):
🎸 - skimmed it and it looks good. It'll be interesting to have a fresh agent test it out.

cygnusv (approved):

derekpierre:
@theref you probably need to rebase over `signing-epic` which I just force-pushed 😊

theref:
looks like the force push has broken ci/cd?

derekpierre:
> looks like the force push has broken ci/cd?

I think this is related to #773 . We can review / merge that PR first and then merge this one.

